PR #2, opened 2018-01-12
Ying/dps

YingXue:
The unit test for enrollment and enrollment-group **update** fails, as I couldn't figure out the mocks.
I am sending this pre-PR in order to get some comments earlier.


azext_iot/_help.py
```diff
@@ -343,3 +343,137 @@
     type: command
     short-summary: Delete an IoT Hub Edge deployment.
 """
+
+helps['iot dps'] = """
+    type: group
+    short-summary: Manage Azure IoT provisioning services.
```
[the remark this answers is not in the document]
digimaun: Make sure you review the help for the other commands for this consistency.

azext_iot/_params.py
```diff
@@ -119,3 +120,43 @@ def load_arguments(self, _):
         c.argument('labels', options_list=['--labels', '-lab'],
                    help="""Map of labels to be applied to target configuration.
                            Use the following format:'{\"key0\":\"value0\", \"key1\":\"value1\"}'""")
+
+    with self.argument_context('iot dps') as c:
+        c.argument('dps_name', help='Name of the Azure IoT Hub device provisioning service')
+        c.argument('initial_twin_properties', options_list=['--initial-twin-properties', '-props'],
```
digimaun: Glad we are being consistent with parameter short names, like '-props'.

azext_iot/_params.py
```diff
@@ -119,3 +120,43 @@ def load_arguments(self, _):
...
+
+    with self.argument_context('iot dps enrollment') as c:
+        c.argument('enrollment_id', help='ID of device enrollment record')
+        c.argument('device_id', help='IoT Hub Device ID')
+        
+    with self.argument_context('iot dps enrollment create') as c:
+        c.argument('attestation_type', arg_type=get_enum_type(AttestationType), help='Attestation Mechanism')
+        c.argument('certificate_path', options_list=['--certificate-path', '-p'], 
+                   help='The path to the file containing the certificate. When choosing x509 as attestation type, certificate path is required.')
+        c.argument('endorsement_key', options_list=['--endorsement-key', '-k'],
```
digimaun: For consistency in style, for multi word parameter names like '--endorsement-key' , consider changing the short name to the first letter of each word like '-ek'.

azext_iot/commands.py
```diff
@@ -5,11 +5,13 @@
...
 
 from azure.cli.core.commands import CliCommandType
+from azext_iot._factory import iot_hub_service_factory 
+from azext_iot._factory import iot_service_provisioning_factory
 from azext_iot import iotext_custom
 
 
 def load_command_table(self, _):
-    with self.command_group('iot hub') as g:
+    with self.command_group('iot hub', client_factory=iot_hub_service_factory) as g:
```
digimaun: Because iot_hub_service_factory is the default client_factory, you don't need to set it here...DPS commands should override though.

azext_iot/commands.py
```diff
@@ -60,13 +62,32 @@ def load_command_table(self, _):
...
         g.custom_command('abandon', 'iot_c2d_message_abandon')
         g.custom_command('reject', 'iot_c2d_message_reject')
         g.custom_command('receive', 'iot_c2d_message_receive')
+        g.custom_command('receive-mqtt', 'iot_c2d_message_receive_mqtt')
+
+    with self.command_group('iot dps enrollment' , client_factory=iot_service_provisioning_factory) as g:
+        g.custom_command('create', 'iot_dps_device_enrollment_create')
+        g.custom_command('list', 'iot_dps_device_enrollment_list')
+        g.custom_command('show', 'iot_dps_device_enrollment_get')
+        g.custom_command('update', 'iot_dps_device_enrollment_update')
```
digimaun: Were you going to use generic_update_command style for consistency in entity updates?

azext_iot/common/certops.py
```diff
@@ -32,22 +32,23 @@ def create_self_signed_certificate(subject, valid_days, cert_output_dir):
...
-        open(join(cert_output_dir, key_file), "wt").write(key_dump)
+
+        if not cert_only:
+            open(join(cert_output_dir, key_file), "wt").write(key_dump)
 
     return {
         'certificate': cert_dump,
         'privateKey': key_dump,
-        'thumbprint': thumbprint
+        'thumbprint': thumbprint,
```
digimaun: Extra comma?

azext_iot/common/certops.py
```diff
@@ -32,22 +32,23 @@ def create_self_signed_certificate(subject, valid_days, cert_output_dir):
...
 
-
 def open_certificate(certificate_path):
     certificate = ""
     if certificate_path.endswith('.pem') or certificate_path.endswith('.cer'):
         with open(certificate_path, "rb") as cert_file:
             certificate = cert_file.read()
             try:
-                certificate = certificate.decode("utf-8")
+                certificate = certificate.decode("utf-8").rstrip("\r\n")
```
digimaun: For understanding, what is the benefit of this?

azext_iot/custom.py
```diff
@@ -800,3 +815,279 @@ def iot_device_upload_file(client, device_id, hub_name, file_path, content_type,
         custom_sdk.post_file_notification(device_id, container['correlationId'])
     except errors.ErrorDetailsException as e:
         raise CLIError(e)
+
+# DPS Enrollments
+
+API_VERSION = '2017-11-15'
```
digimaun: Consider storing API_VERSION in the SDK Client scope, so you don't have to provide it for every function. The other SDK's will do this.

Look at the custom SDK for a short example.

https://github.com/Azure/azure-iot-cli-extension/blob/master/azext_iot/custom_sdk/custom_api.py#L47

azext_iot/custom.py
```diff
@@ -800,3 +815,279 @@ def iot_device_upload_file(client, device_id, hub_name, file_path, content_type,
...
+def iot_dps_device_enrollment_list(client, dps_name, resource_group_name):
+    from azext_iot.dps_sdk.models.query_specification import QuerySpecification
+    target = get_iot_dps_connection_string(client, dps_name, resource_group_name)
+    try:
+        m_sdk, errors = _bind_sdk(target, SdkType.dps_sdk)
+        query_command = "SELECT *"
```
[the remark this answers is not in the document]
digimaun: Also for my own understanding 'SELECT *' is valid for DPS? Strange its not selecting FROM `<type>`.

azext_iot/custom.py
```diff
@@ -800,3 +815,279 @@ def iot_device_upload_file(client, device_id, hub_name, file_path, content_type,
...
+        return m_sdk.registration_status.delete_registration_state(registration_id, API_VERSION)
+    except errors.ErrorDetailsException as e:
+        raise CLIError(e)
+
+def _get_initial_twin(initial_twin_tags = None, initial_twin_properties = None):
+    import ast
+    if not initial_twin_tags and not initial_twin_properties:
+        return None
+    if initial_twin_tags:
+        initial_twin_tags = evaluate_literal(str(initial_twin_tags), dict)
```
digimaun: Please use the existing utility evaluate_literal, it will return None if the input is not parsable.

azext_iot/custom.py
```diff
@@ -800,3 +815,279 @@ def iot_device_upload_file(client, device_id, hub_name, file_path, content_type,
...
+                                     iot_hub_host_name = None,
+                                     initial_twin_tags = None,
+                                     initial_twin_properties = None,
+                                     provisioning_status = None):
+    target = get_iot_dps_connection_string(client, dps_name, resource_group_name)
+    try:
+        m_sdk, errors = _bind_sdk(target, SdkType.dps_sdk)
+           
+        enrollment_record = m_sdk.device_enrollment.get(enrollment_id, API_VERSION)
+        if not etag == enrollment_record['etag'].replace('"', ''):
```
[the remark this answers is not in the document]
YingXue: I sepecifically choose not to used the generic update, because the way this API is designed. Thare are quite a lot readonly properties in the GET response, and if I am the user, I would quite confused why I am not able to update a lot of the properties in them. That's why I have the update implemented very similar to create, to make it easier for the users.

azext_iot/common/certops.py
YingXue: This is a service bug, that they cannot handle certificate containing a \r\n as the ending. They are fixing it, but would take a while to deploy, so I am adding this as a work around. I will double check if they have already fixed it before we release the CLI

azext_iot/commands.py
YingXue: not planning to. See the comments below.

azext_iot/commands.py
YingXue: sure. will remove

azext_iot/commands.py
```diff
@@ -5,11 +5,13 @@
 # --------------------------------------------------------------------------------------------
 
 from azure.cli.core.commands import CliCommandType
+from azext_iot._factory import iot_hub_service_factory
```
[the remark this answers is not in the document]
YingXue: i see. will remove

azext_iot/_help.py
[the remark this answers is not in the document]
YingXue: Sure. I should have check the helps before the CR :)

azext_iot/custom.py
YingXue: Ah i thought i ready did, but i forgot to remove the import ast at the top

azext_iot/_factory.py
```diff
@@ -13,13 +13,19 @@ def iot_hub_service_factory(cli_ctx, *_):
...
+    return get_mgmt_service_client(cli_ctx, IotDpsClient)
+
 
 def _bind_sdk(target, sdk_type, device_id=None):
     from azext_iot.device_query_sdk.device_identities_api import DeviceIdentitiesAPI
     from azext_iot.device_twin_sdk.device_twin_api import DeviceTwinAPI
     from azext_iot.modules_sdk.iot_hub_client import IotHubClient
     from azext_iot.device_msg_sdk.iot_hub_device_client import IotHubDeviceClient
     from azext_iot.custom_sdk.custom_api import CustomClient
+    from azext_iot.dps_sdk.device_provisioning_service_service_runtime_client import DeviceProvisioningServiceServiceRuntimeClient
```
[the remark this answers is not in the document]
YingXue: I am not 100% sure about the approach here that we do too many cutomizations in the SDK. How are we planning to handle the swagger changes in the future? We could autorest new SDK, and see the diffs and modify accordingly, but i was thinking only if the SDK has big flaws or bugs that we have to workaround, I would prefer to use it as is in most cases. I am totally fine with changign the namesapce here as it is quite straighforward, but changing the class name may not to seem necessary. We can talk about this more later.

digimaun:
Looks good to me.

PR #10, opened 2018-02-06
Secondary cert

YingXue:
1. Added support for a secondary certificate for enrollments create and update operations.
2. Updated tests accordingly.

### DIFF
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -334,7 +334,7 @@ helps['iot edge deployment create'] = """
     - name: Create deployment with condition where a device is in 'building 9' and
             the environment is 'test'.
       text: >
-        az iot edge deployment create -c myconfig -n myhub --content ../mycontent.json
+        az iot edge deployment create -c [configuration] -n [IoTHub Name] --content ../mycontent.json
         -lab '{"key0":"value0", "key1":"value1"}'
         --target-condition "tags.building=9 and tags.environment='test'" --priority 3
 """
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -36,12 +36,12 @@ helps['iot hub device-identity create'] = """
     - name: Create an IoT device with self-signed certificate authorization,
             generate a cert valid for 10 days then use its thumbprint.
       text: >
-        az iot hub device-identity create -n [IoTHub Name] -d [Device ID] 
+        az iot hub device-identity create -n [IoTHub Name] -d [Device ID]
         -am x509_thumbprint --valid-days 10
     - name: Create an IoT device with self-signed certificate authorization,
             generate a cert of default expiration (365 days) and output to target directory.
       text: >
-        az iot hub device-identity create -n [IoTHub Name] -d [Device ID] -am x509_thumbprint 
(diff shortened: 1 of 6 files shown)


azext_iot/_help.py
```diff
@@ -331,10 +331,10 @@
     short-summary: Create an IoT Edge deployment in the target IoT Hub.
     long-summary: Configuration json must have root of 'content' like {"content":{...}}.
     examples:
-    - name: Create deployment with condition where a device is in 'building 9' and 
+    - name: Create deployment with condition where a device is in 'building 9' and
             the environment is 'test'.
       text: >
-        az iot edge deployment create -c myconfig -n myhub --content ../mycontent.json 
+        az iot edge deployment create -c myconfig -n myhub --content ../mycontent.json
```
digimaun: We might as well variable'ize the deployment config id and hub parameters for the edge deployment <> help for consistency. Something Like:

az iot edge deployment create -c [Config Id] -n [IoTHub Name] --content ../mycontent.json

YingXue: sure why not

azext_iot/common/certops.py
```diff
@@ -80,4 +81,4 @@ def open_certificate(certificate_path):
                 certificate = certificate.decode("utf-8")
             except UnicodeError:
                 certificate = base64.b64encode(certificate).decode("utf-8")
-    return certificate
+    return certificate.rstrip()  # Remove trailing white space from the certificate content
```
YingXue: This is added back in as the service seems to not able to deal with cert ended with an empty new line (again?). It does no harm to us and it more robust.

azext_iot/_params.py
```diff
@@ -164,25 +165,58 @@ def load_arguments(self, _):
...
                          help='TPM endorsement key for a TPM device. '
-                         ' When choosing tpm as attestation type, endorsement key is required.')
+                         'When choosing tpm as attestation type, endorsement key is required.')
 
     with self.argument_context('iot dps enrollment update') as context:
-        context.argument('certificate_path', options_list=['--certificate-path', '-cp'],
-                         help='The path to the file containing the certificate. '
-                         'When updating enrollment with x509 attestation mechanism, '
-                         'certificate path is required.')
+        context.argument('certificate_path',
```
digimaun: It looks like you can put '--certificate-path', '--secondary-certificate-path', '--remove-certificate' and '--remove-secondary-certificate' in the parent 'iot dps' argument context to reduce repetition.

YingXue: I will put '--remove-certificate' and '--remove-secondary-certificate' in the parent, but '--certificate-path', '--secondary-certificate-path' would have to stay because it differs between enrollment, enrollment-group as well as between create and update.

digimaun: Ok. Is that because of the difference in help text?

digimaun: Yea it should work like that

YingXue: Yes. Whether Enrollment cert path is required is based on its attestation type. Group enrollment cert path is always required. When update, none of them is required, and that is what I put into the parent since they are common.

PR #12, opened 2018-02-14
update enrollment update tests

YingXue:
Updated enrollment (individual and group) tests with more user updatable parameters

### DIFF
--- a/azext_iot/tests/test_iot_dps_unit.py
+++ b/azext_iot/tests/test_iot_dps_unit.py
@@ -7,14 +7,13 @@
 import pytest
 import json
 from azext_iot.operations import dps as subject
-from azext_iot.common.utility import evaluate_literal
 from knack.util import CLIError
 from azext_iot.common.sas_token_auth import SasTokenAuthentication
-from azure.cli.core.util import read_file_content
 
 enrollment_id = 'myenrollment'
 resource_group = 'myrg'
 registration_id = 'myregistration'
+etag = 'AAAA=='
 
 mock_target = {}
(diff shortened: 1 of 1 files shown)


azext_iot/tests/test_iot_dps_unit.py
```diff
@@ -195,32 +203,101 @@ def generate_enrollment_show(**kvp):
...
+                                        initial_twin_properties={'newKey': 'newValue'},
+                                        provisioning_status='enabled',
+                                        device_id='newId'))
     ])
-    def test_enrollment_group_update(self, serviceclient, result):
-        serviceclient.return_value.text = json.dumps(result)
-        subject.iot_dps_device_enrollment_update(None, enrollment_id, mock_target['entity'], resource_group, 'AAAA==', None,
-                                                 'newCertPath')
-
+    def test_enrollment_group_update(self, serviceclient, req):
```
digimaun: func name should be 'test_enrollment_update(...)'

azext_iot/tests/test_iot_dps_unit.py
```diff
@@ -195,32 +203,101 @@ def generate_enrollment_show(**kvp):
...
 
         body = args[0][2]
-'''
+
+        if not req['certificate_path']:
+            if req['remove_certificate_path']:
+                assert body['attestation']['x509']['clientCertificates']['primary'] is None
+            else:
+                assert body['attestation']['x509']['clientCertificates']['primary']['info'] is not None
+        if req['certificate_path']:
```
digimaun: You will get an index exception if ```req['thing']``` doesn't exist, better to do ```req.get('thing')``` which will return ```None``` if the index item doesn't exist.

PR #20, opened 2018-04-09
Ying/rebase

YingXue:
I recreated a new PR after the rebasing. 
Support CA certificate in DPS enrollment group -- users would able to provide CA certificate while creating enrollment group, certificate type can be changed during enrollment group update.
Updated unit tests and integration tests accordingly.

### DIFF
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -505,14 +505,20 @@ helps['iot dps enrollment-group create'] = """
     short-summary: Create an enrollment group in an Azure IoT Hub Device Provisioning Service.
     examples:
     - name: Create an enrollment group '[Enrollment ID]' in the Azure IoT provisioning service
-            '[DPS Name]' in the resource group '[Resource Group Name]'.
+            '[DPS Name]' in the resource group '[Resource Group Name] using an intermediate certificate as primary certificate'.
       text: >
         az iot dps enrollment-group create -g [Resource Group Name] --dps-name [DPS Name]
         --enrollment-id [Enrollment ID] --certificate-path /certificates/Certificate.pem
     - name: Create an enrollment group '[Enrollment ID]' in the Azure IoT provisioning service
+            '[DPS Name]' in the resource group '[Resource Group Name] using a CA certificate [Certificate Name]
+            as secondary certificate'.
+      text: >
+        az iot dps enrollment-group create -g [Resource Group Name] --dps-name [DPS Name]
+        --enrollment-id [Enrollment ID] --secondary-certificate-name [Certificate Name]
+    - name: Create an enrollment group '[Enrollment ID]' in the Azure IoT provisioning service
             'MyDps' in the resource group '[Resource Group Name]' with provisioning status
             'enabled', target IoT Hub '[IoTHub Name].azure-devices.net' and initial twin
-            tags '{"location":{"region":"US"}}'.
+            tags '{"location":{"region":"US"}} using an intermediate certificate as primary certificate'.
       text: >
         az iot dps enrollment-group create -g [Resource Group Name] --dps-name [DPS Name]
         --enrollment-id [Enrollment ID] --certificate-path /certificates/Certificate.pem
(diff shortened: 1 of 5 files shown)


azext_iot/_help.py
```diff
@@ -505,14 +505,20 @@
...
         --enrollment-id [Enrollment ID] --certificate-path /certificates/Certificate.pem
+    - name: Create an enrollment group '[Enrollment ID]' in the Azure IoT provisioning service
+            '[DPS Name]' in the resource group '[Resource Group Name] using a CA certificate [Certificate Name]
+            as secondary certificate'.
+      text: >
+        az iot dps enrollment-group create -g [Resource Group Name] --dps-name [DPS Name]
+        --enrollment-id [Enrollment ID] --secondary-certificate-name [Certificate Name]
     - name: Create an enrollment group '[Enrollment ID]' in the Azure IoT provisioning service
             'MyDps' in the resource group '[Resource Group Name]' with provisioning status
             'enabled', target IoT Hub '[IoTHub Name].azure-devices.net' and initial twin
```
digimaun: The '.azure-devices.net' part might be misleading as these commands should work across domains (for example the Chinese National Cloud).

azext_iot/_params.py
```diff
@@ -233,11 +234,27 @@ def load_arguments(self, _):
         context.argument('certificate_path',
                          options_list=['--certificate-path', '-cp'],
                          help='The path to the file containing the primary certificate. '
-                         'One of the certificate path is required.')
+                         'If choose to use intermediate certificate, one of the certificate path is required.')
```
digimaun: This description may be more clear 'If attestation with an intermediate certificate is desired then a certificate path is required.'

Similarly for CA:

'If attestation with a CA certificate is desired then a certificate name is required.'

azext_iot/operations/dps.py
```diff
@@ -223,6 +238,8 @@ def iot_dps_device_enrollment_group_update(client,
                                            etag,
```
digimaun: Consider removing the requirement for etag (like in the enrollment group update command). You don't need to follow the generic update process to achieve. In the current versions of the iot extension, the iot hub based commands don't require etag as it is derived wherever necessary. This is one less parameter users have to deal with (especially when some commands have tons of overloads).

azext_iot/operations/dps.py
```diff
@@ -193,6 +194,8 @@ def iot_dps_device_enrollment_group_create(client,
                                            resource_group_name,
                                            certificate_path=None,
                                            secondary_certificate_path=None,
+                                           certificate_name=None,
+                                           secondary_certificate_name=None,
                                            iot_hub_host_name=None,
                                            initial_twin_tags=None,
```
digimaun: For processing initial_twin_tags and initial_twin_properties (like in _get_initial_twin)...use azext_iot.common.utility.shell_safe_json_parse this allows you to support json _or_ python dictionary notation.

azext_iot/_params.py
digimaun: or even 'If attestation with an intermediate certificate is desired then a certificate path must be provided.' ........

azext_iot/operations/dps.py
YingXue: Do we consider the possibility that multiple users running the command at the same time?

azext_iot/_params.py
```diff
@@ -233,11 +234,27 @@ def load_arguments(self, _):
...
+                         help='The name of the primary CA certificate. '
+                         'If choose to use CA certificate, one of the certificate name is required.')
+        context.argument('secondary_certificate_name',
+                         options_list=['--secondary-certificate-name', '-scn'],
+                         help='The name of the secondary CA certificate. '
+                         'If choose to use CA certificate, one of the certificate name is required.')
+
+    with self.argument_context('iot dps enrollment-group update') as context:
+        context.argument('certificate_name',
+                         options_list=['--certificate-name', '-cn'],
```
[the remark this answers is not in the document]
YingXue: yes, it is. sure.

azext_iot/operations/dps.py
digimaun: @YingXue and I discussed this topic and determined the best compromise was to have etag be an optional parameter.

PR #28, opened 2018-05-19
Changeset for 0.4.5

digimaun:
* Introduces experimental C2D message send.
* Concurrently support 0.4.0 + 0.5.0 IoT mgmt SDK
* Improved top parameter for list ops
* Generalize uamqp dependency check


azext_iot/_params.py
```diff
@@ -60,7 +61,8 @@ def load_arguments(self, _):
         context.argument('duration', options_list=['--duration', '-du'], type=int,
                          help='Valid token duration in seconds.')
         context.argument('etag', options_list=['--etag', '-e'], help='Entity tag value.')
-        context.argument('top', type=int, help='Maximum number of elements to return.')
+        context.argument('top', type=int, options_list=['--top', '-top'],
+                         help='Maximum number of elements to return. Use -1 for unlimited.')
```
YingXue: Looking at the code, even if users don't provide value -1, it is still unlimited, we can inform the user too?

digimaun: There are 2 types of top parameters to consider...one is for query, which is optional and defaults to unlimited return size. The other is for the explicit list operations (configs, devices, modules)...these have a default top value which can't be set to None, idea being user doesn't get overwhelmed with data and explicitly sets the top to desired amount. For most of the explicit lists -1 is how you achieve 'unlimited' for top. The exception here is with deployments/configurations because the service limit is only 10 (which is also the default top) so -1 shouldn't work. On the other hand query assumes unlimited (because query is more oriented towards power users) and -1 is a synonym in that case.

All this being said the top param text needs to change anyway for configs since -1 should not be valid for it!!

digimaun: Oh, definitely ready to discuss more if needed.

YingXue: I see. Makes total sense.

PR #34, opened 2018-06-26
Major updates to support IoT Configurations GA

digimaun:
* Integrated new Data-plane SDK
* New configuration command group
* New edge set-modules (deprecates apply-configuration)
* New show-metric functionality for deploys/configs
* uAMQP increment to v0.1.0rc1 exact version (vs compatible)
* Test updates
* Misc tweaks and improvements


azext_iot/_constants.py
```diff
@@ -6,16 +6,17 @@
 
 import os
 
-VERSION = '0.4.5'
+VERSION = '0.5.0'
 EXTENSION_NAME = 'azure-cli-iot-ext'
 EXTENSION_ROOT = os.path.dirname(os.path.abspath(__file__))
 EXTENSION_CONFIG_ROOT_KEY = 'iotext'
+BASE_API_VERSION = '2018-06-30'            # '2018-06-01-preview'  # '2018-05-02'
```
digimaun: Whoops the comment needs to be removed

azext_iot/_help.py
```diff
@@ -400,58 +481,90 @@
 
 helps['iot edge'] = """
     type: group
-    short-summary: Deploy and manage IoT solutions on the Edge.
+    short-summary: Manage IoT solutions on the Edge.
```
YingXue: this summary feels a bit plain to me...Something copied from this page and maybe with a link is more useful: https://docs.microsoft.com/en-us/azure/iot-edge/how-iot-edge-works

azext_iot/_params.py
```diff
@@ -72,8 +73,10 @@ def load_arguments(self, _):
         context.argument('auth_method', options_list=['--auth-method', '-am'],
                          arg_type=get_enum_type(DeviceAuthType),
                          help='The authorization type an entity is to be created with.')
-        context.argument('content', options_list=['--content', '-k'],
-                         help='IoT Edge configuration content json. Provide file path or raw json.')
+        context.argument('metric_type', options_list=['--metric-type', '-mt'], arg_type=get_enum_type(MetricType),
```
YingXue: do they still support -mt, or if we want to move away from this convension in the future?

digimaun: that breaking change was rolled back, but its a good idea to stay away from the convention for the future

azext_iot/_help.py
digimaun: good idea

digimaun:
Squash + merge based on offline discussion with @YingXue .

PR #57, opened 2018-11-17
add output param to monitor-events

jongio:


azext_iot/operations/events3/_events.py
```diff
@@ -127,7 +127,15 @@ def _output_msg_kpi(msg):
         event_source = {'event': {}}
 
         event_source['event']['origin'] = origin
-        event_source['event']['payload'] = str(next(msg.get_data()), 'utf8')
+
+        payload = ''
```
jongio: `str(next(msg.get_data()))` was crashing the vs code terminal when --data was null.  I'm checking for data before calling next.

azext_iot/operations/events3/_events.py
```diff
@@ -142,7 +150,18 @@ def _output_msg_kpi(msg):
             if app_prop:
                 event_source['event']['properties']['application'] = unicode_binary_map(app_prop)
 
-        six.print_(yaml.dump(event_source, default_flow_style=False), flush=True)
+        if output and output == 'json':
+            try:
+                if event_source['event']['payload']:
+                    event_source['event']['payload'] = json.loads(event_source['event']['payload'].replace('\\n', '').replace('\\r', ''))
```
jongio: Do you think we should RegEx this?

azext_iot/operations/events3/_events.py
```diff
@@ -142,7 +150,18 @@ def _output_msg_kpi(msg):
...
+            except Exception:  # pylint: disable=broad-except
+                # We don't want to crash the monitor if JSON parsing fails.
+                six.print_('Message payload is not valid JSON.')
```
jongio: I'm catching all JSON errors so we don't crash the terminal.

azext_iot/operations/hub.py
```diff
@@ -1125,6 +1126,7 @@ def iot_hub_monitor_events(cmd, hub_name=None, device_id=None, consumer_group='$
     timeout = (timeout * 1000)
 
     config = cmd.cli_ctx.config
+    output = cmd.cli_ctx.invocation.data.get("output", None) #get_output_format(cmd.cli_ctx)
```
jongio: Ignore this for now.  I'm working with the CLI team to figure out the best way to get a global parameter from within an extension.

jongio:
1. This is not ready to be merged. Starting a PR to get init review
2. We are waiting to hear from CLI team on best way to access global param
3. Still need to write tests for this feature.

azext_iot/operations/events3/_events.py
digimaun: Yes I do...a compiled pattern would fit the bill...something like:

```python
pattern = re.compile(r'\\r\\n|\\r|\\n')
...
...
...
if event_source['event']['payload']:
    search_result = pattern.search(event_source['event']['payload'])
    if search_result:
        event_source['event']['payload'] = event_source['event']['payload'].replace('\\n', '').replace('\\r', '')
```

azext_iot/operations/events3/_events.py
digimaun: The try/except in Python is extremely efficient but not if the expectation is wrong most of the time (exception gets triggered more often than not). In the current state, if the output setting is json and you are receiving telemetry that isn't compatible with that, you are taking a notable performance hit since every message triggers the exception.

This six.print_ msg should indicate to switch to yaml output for raw data and it should be clearly noted in params and help text to not use output JSON if you don't expect JSON.

Ideally, maybe in a future state the incoming messages content-type system property should be set with say application/json, then on the receiving end the message payload goes through the respective conversion process and if that fails, fallback to default (raw text) and output the invalid json notice (if content-type is application/json the expectation is for valid json).

jongio: The default output is json. So as of now this code will try to output as json. 

Maybe we need a content-type property or a flag that indicates the input type and only parse the payload as json if that is set to json. 

The current code assumes that if you want json output that means you have json input which isn’t always the case. 

Let me revise this and see if we can inspect and determine payload type without having to loads.

digimaun: Yes this what I'm trying to describe in the last paragraph. There is a system property called content-type. You can test setting it via send-d2c-message by including the property like --props '$.ct=application/json'

jongio: This on send-d2c-message:

`--props $.ct=application/json`

Is automatically translated to this:
```
"properties": {
            "system": {
                "content_type": "application/json"
            }
        }
```

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -40,6 +43,8 @@
 class TestIoTHub(LiveScenarioTest):
     def setUp(self):
         self._entity_names = None
+        self.cmd('az iot hub consumer-group create --hub-name {} --resource-group {} --name {}'.format(LIVE_HUB, LIVE_RG, LIVE_CONSUMER_GROUP1), checks=[self.check('name', LIVE_CONSUMER_GROUP1)])
```
digimaun: Does an exception get raised if the 'test1', 'test2' consumer-groups already exist?

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -1165,76 +1170,31 @@ def test_hub_monitor_event_all(self):
...
                      checks=[self.check('deviceId', device_ids[i])])
 
-        for i in range(device_count):
-            execute_onthread(method=iot_simulate_device,
-                             args=[client, device_ids[i], LIVE_HUB, 'complete', 'Ping from event monitor test 1, part a', 5, 1],
-                             max_runs=1)
-
-        self.cmd('iot hub monitor-events -n {} -g {} -t 10 -y -p sys anno app'.format(
-            LIVE_HUB, LIVE_RG), checks=None)
+        enqueued_time = datetime.datetime.now().microsecond
```
digimaun: Please see comment in the specific commit

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -1165,76 +1170,31 @@ def test_hub_monitor_event_all(self):
...
 
-        # With connection string
         for i in range(device_count):
-            execute_onthread(method=iot_simulate_device,
```
digimaun: The advantage with iot_simulate_device is you can see the event monitor stability over an elapsed period of time. iot_device_send_message does have a lot more power/flexibility per message though.

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -38,8 +40,27 @@
...
 class TestIoTHub(LiveScenarioTest):
+    def __init__(self, _):
+        from iotext_test_tools import DummyCliOutputProducer
+        super(TestIoTHub, self).__init__(_)
+        self.cli_ctx = DummyCliOutputProducer()
+
     def setUp(self):
         self._entity_names = None
+        for cg in LIVE_CONSUMER_GROUPS:
+            self.cmd('az iot hub consumer-group create --hub-name {} --resource-group {} --name {}'.format(LIVE_HUB, LIVE_RG, cg), checks=[self.check('name', cg)])
```
digimaun: The setUp at this level runs per test method. Should be moved to the test needing custom cg's.

azext_iot/operations/events3/_events.py
```diff
@@ -127,13 +129,34 @@ def _output_msg_kpi(msg):
...
+
+        payload = ''
+
+        data = msg.get_data()
+        if data:
+            payload = str(next(data), 'utf8')
+
+        system_props = unicode_binary_map(parse_entity(msg.properties, True))
+
+        ct = content_type
```
digimaun: Ensure char casing is uniform, I didn't see content_type.lower() prior in execution

azext_iot/tests/test_iot_ext_int.py
jongio: Cool. I thought that only ran once per test run.  I'll update

azext_iot/operations/events3/_events.py
```diff
@@ -127,13 +129,34 @@ def _output_msg_kpi(msg):
...
+        if not ct:
+            ct = system_props['content_type'].lower() if 'content_type' in system_props else ''
```
digimaun: You could skip the `system_props['content_type'].lower()` here, then in the next line do `ct = ct.lower()` to lowercase once, content type either from explicit parameter or from msg payload.

jongio: That's exactly what I just did :)

PR #61, opened 2018-12-04
Add device filtering capabilities (wildcard and regex) to monitor-events

c-ryan-k:
The purpose of this PR is to add device filtering capabilities to the monitor-events command.

This allows users to filter devices with wildcards in the Device ID (like: "Device?Name*" ) as well as with a separate regex argument such as "Device[0-9]*"

---

### General Guidelines

### DIFF
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -119,8 +119,6 @@ def load_arguments(self, _):
         context.argument('properties', options_list=['--properties', '--props', '-p'], arg_type=event_msg_prop_type)
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
-        context.argument('device_regex', options_list=['--device-regex', '--regex'],
-                         help='Specify a regular expression to filter devices')
 
     with self.argument_context('iot hub monitor-feedback') as context:
         context.argument('wait_on_id', options_list=['--wait-on-msg', '-w'],
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -119,6 +119,8 @@ def load_arguments(self, _):
         context.argument('properties', options_list=['--properties', '--props', '-p'], arg_type=event_msg_prop_type)
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
+        context.argument('device_regex', options_list=['--device_regex', '--regex'],
+                         help='Specify a regular expression to filter devices')
 
     with self.argument_context('iot hub monitor-feedback') as context:
         context.argument('wait_on_id', options_list=['--wait-on-msg', '-w'],
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -119,7 +119,7 @@ def load_arguments(self, _):
         context.argument('properties', options_list=['--properties', '--props', '-p'], arg_type=event_msg_prop_type)
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
-        context.argument('device_regex', options_list=['--device_regex', '--regex'],
+        context.argument('device_regex', options_list=['--device-regex', '--regex'],
                          help='Specify a regular expression to filter devices')
 
     with self.argument_context('iot hub monitor-feedback') as context:
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -119,6 +119,7 @@ def load_arguments(self, _):
         context.argument('properties', options_list=['--properties', '--props', '-p'], arg_type=event_msg_prop_type)
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
+        context.argument('device_query', options_list=['--device-query', '-q'], help='Specify a custom query to filter devices.')
 
     with self.argument_context('iot hub monitor-feedback') as context:
         context.argument('wait_on_id', options_list=['--wait-on-msg', '-w'],
--- a/azext_iot/operations/events3/_events.py
+++ b/azext_iot/operations/events3/_events.py
@@ -26,10 +26,11 @@ logger = get_logger(__name__)
 DEBUG = True
 
 
-def executor(target, consumer_group, enqueued_time, device_id=None, properties=None, timeout=0, output=None, content_type=None):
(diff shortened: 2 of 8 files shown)


azext_iot/_params.py
```diff
@@ -119,6 +119,8 @@ def load_arguments(self, _):
         context.argument('properties', options_list=['--properties', '--props', '-p'], arg_type=event_msg_prop_type)
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
+        context.argument('device_regex', options_list=['--device-regex', '--regex'],
```
jongio: I would prefer not to use the query technology in the name - "regex".

How about something like --device-query and support queries like we do in the "query" operation as well...
https://docs.microsoft.com/en-us/cli/azure/ext/azure-cli-iot-ext/iot/hub?view=azure-cli-latest#ext-azure-cli-iot-ext-az-iot-hub-query

azext_iot/operations/events3/_events.py
```diff
@@ -125,7 +127,12 @@ def _get_conn_props():
     def _output_msg_kpi(msg):
         # TODO: Determine if amqp filters can support boolean operators for multiple conditions
         origin = str(msg.annotations.get(b'iothub-connection-device-id'), 'utf8')
-        if device_id and origin != device_id:
+        if device_id:
```
jongio: Do we have to use regex on every request?  I would rather us not take the regex overhead (even if small) if we don't have to.

azext_iot/_params.py
jongio: It would be great if we could support full queries as well - to allow users to query for devices or filter by tags on the devices.  I would rather implement that then this one off just for regex on device id.

@iluican

c-ryan-k: That's a great suggestion - I'll admit in hindsight regex for device IDs is a little overkill considering wildcards pretty much handle most uses of device ID filtering. I'll edit this PR to keep it to just device ID wildcards and work on the --device-query feature in another PR

jongio: Do you need both?  If you have --device-query, then you don't really need device wildcard.  Less surface area, less code, less bugs.

digimaun: @c-ryan-k Can you please validate if 'like' or similar wildcard matching is supported in the IoT Hub Query API. At the time of the implementation of the query command that was not supported yet. This data point will help guide the path forward.

jongio: Correct. It doesn't not support LIKE.  

Is it possible to ask the user to use tags instead of name?  If so, then supporting the full query like in the link above would work.

If they can't, then allowing user to put a wildcard in the deviceId property and detect supported wildcards would be idea...versus a new parameter.

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -1238,6 +1218,22 @@ def test_hub_monitor_events(self):
                                     .format(LIVE_HUB, LIVE_RG, DEVICE_PREFIX + '*', enqueued_time),
                                     device_ids)
 
+        # Monitor events for specific devices using query language
+        device_subset_include = device_ids[:device_count//2]
+        device_include_string = ', '.join(['\'' + deviceId + '\'' for deviceId in device_subset_include])
```
jongio: I feel like we should be using a format string here.

azext_iot/operations/events3/_events.py
```diff
@@ -127,12 +127,14 @@ def _get_conn_props():
     def _output_msg_kpi(msg):
         # TODO: Determine if amqp filters can support boolean operators for multiple conditions
         origin = str(msg.annotations.get(b'iothub-connection-device-id'), 'utf8')
-        if device_id:
-            regex = re.escape(device_id).replace("\\*", ".*").replace('\\?', ".") + "$"
-            if not re.match(regex, origin):
+        if device_id and device_id != origin:
+            if '*' in device_id or '?' in device_id:
```
jongio: Consider putting * and ? in an array and then do lookup.

azext_iot/operations/events3/_events.py
```diff
@@ -125,7 +127,14 @@ def _get_conn_props():
...
         origin = str(msg.annotations.get(b'iothub-connection-device-id'), 'utf8')
-        if device_id and origin != device_id:
+        if device_id and device_id != origin:
+            if '*' in device_id or '?' in device_id:
+                regex = re.escape(device_id).replace("\\*", ".*").replace('\\?', ".") + "$"
+                if not re.match(regex, origin):
+                    return
+            else:
+                return
+        if devices and origin not in devices:
```
digimaun: Use dict.get() instead of iteration through keys.

digimaun:
Please make these remaining tweaks and we will get this in.

azext_iot/operations/events3/_events.py
c-ryan-k: Can you elaborate a bit? We just want to check if that value is in the "devices" dictionary (x in y), we don't need the value (even though they're only set to True when we create the dictionary).

In my testing, the `key in dict` lookup was more performant - though I understand `dict.get()` is safer if we're using the resultant value since we can specify a default / no-exist value.

azext_iot/operations/events3/_events.py
c-ryan-k: In my testing, two string lookups with `in` performed much better than, say: 
`any(x in <string> for x in ['*', '?']`

`timeit.timeit("'*' in 'random_string' or '?' in 'random_string'")
0.07972277299995767
 timeit.timeit("any(x in 'random_string' for x in ['*', '?'])")
0.4291381340000271`

We could loop through a 'wildcard_chars' array and check for each wildcard character, but for only two characters is it necessary?

jongio: excellent.  Thanks for doing the research on that.  It might warrant a blog post and further research to see at what point someone should consider an array lookup.

azext_iot/operations/events3/_events.py
digimaun: The thought was, iterating through a list of keys (particularly when sufficient large) is slower than a safe (due to default value) direct look up.

digimaun: Just wrote a small test script for this, it does looks like 'in' is faster like you mentioned. This is probably due to bytecode optimization and not needing to lookup a member attribute get, and executing method (expensive).

Kudos.

digimaun: Side note/clarification, dict[...] is the expected performant lookup.

c-ryan-k: I see...from my understanding, as long as the right-hand operand is a dictionary, python does a key lookup with the "in" keyword, as opposed to iterating through the keys.

PR #63, opened 2019-01-29
Update generate-sas-token command for modules

anusapan:
---

### General Guidelines

### DIFF
--- a/azext_iot/common/_azure.py
+++ b/azext_iot/common/_azure.py
@@ -30,6 +30,11 @@ def parse_iot_device_connection_string(cs):
     return _parse_connection_string(cs, validate, 'Device')
 
 
+def parse_iot_device_module_connection_string(cs):
+    validate = ['HostName', 'DeviceId', 'ModuleId', 'SharedAccessKey']
+    return _parse_connection_string(cs, validate, 'Module')
+
+
 CONN_STR_TEMPLATE = 'HostName={};SharedAccessKeyName={};SharedAccessKey={}'
 
 
--- a/azext_iot/common/_azure.py
+++ b/azext_iot/common/_azure.py
@@ -16,7 +16,7 @@ def _parse_connection_string(cs, validate=None, cstring_type='entity'):
     decomposed_lower = dict((k.lower(), v) for k, v in decomposed.items())
     if validate:
         for k in validate:
-            if not decomposed.get(k) and not decomposed_lower.get(k.lower()):
+            if not any([decomposed.get(k), decomposed_lower.get(k.lower())]):
                 raise ValueError('{} connection string has missing property: {}'.format(cstring_type, k))
     return decomposed
 
--- a/azext_iot/common/_azure.py
+++ b/azext_iot/common/_azure.py
@@ -13,9 +13,10 @@ from azext_iot._factory import iot_hub_service_factory
 
 def _parse_connection_string(cs, validate=None, cstring_type='entity'):
     decomposed = validate_key_value_pairs(cs)
+    decomposed_lower = dict((k.lower(), v) for k, v in decomposed.items())
     if validate:
         for k in validate:
-            if not decomposed.get(k):
+            if not decomposed.get(k) and not decomposed_lower.get(k.lower()):
                 raise ValueError('{} connection string has missing property: {}'.format(cstring_type, k))
     return decomposed
 
@@ -73,14 +74,15 @@ def get_iot_hub_connection_string(
     if login:
         try:
             decomposed = parse_iot_hub_connection_string(login)
+            decomposed_lower = dict((k.lower(), v) for k, v in decomposed.items())
         except ValueError as e:
             raise CLIError(e)
 
         result = {}
         result['cs'] = login
-        result['policy'] = decomposed['SharedAccessKeyName']
-        result['primarykey'] = decomposed['SharedAccessKey']
-        result['entity'] = decomposed['HostName']
+        result['policy'] = decomposed_lower['sharedaccesskeyname']
+        result['primarykey'] = decomposed_lower['sharedaccesskey']
+        result['entity'] = decomposed_lower['hostname']
         return result
 
     client = None
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -935,21 +935,24 @@ def _iot_c2d_message_receive(target, device_id, lock_timeout=60):
         request_headers['IotHub-MessageLockTimeout'] = str(lock_timeout)
 
     try:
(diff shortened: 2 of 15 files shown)


azext_iot/operations/hub.py
```diff
@@ -735,7 +739,7 @@ def iot_device_module_method(cmd, device_id, module_id, method_name, hub_name=No
 
 # Utility
 
-def iot_get_sas_token(cmd, hub_name=None, device_id=None, policy_name='iothubowner',
+def iot_get_sas_token(cmd, hub_name=None, device_id=None, policy_name='iothubowner', module_id=None,
```
jongio: I would like to see consistent order of params.  For example, below you have module_id right after device_id, and here it is after policy_name.

azext_iot/common/_azure.py
```diff
@@ -30,6 +30,11 @@ def parse_iot_device_connection_string(cs):
     return _parse_connection_string(cs, validate, 'Device')
 
 
+def parse_iot_device_module_connection_string(cs):
+    validate = ['HostName', 'DeviceId', 'ModuleId', 'SharedAccessKey']
```
jongio: Please ensure that this code is not case sensitive.  For example, if someone uses hostname instead of HostName, your code should still work.

anusapan: I think it would work as "parse_iot_hub_connection_string" & "parse_iot_device_connection_string" other methods have same format

jongio: I suggest writing a test to make sure

azext_iot/operations/hub.py
anusapan: fixed

azext_iot/common/_azure.py
anusapan: The user can't submit the connectionstring as it is generated in code with pre-defined template "HostName={};DeviceId={};ModuleId={};{}" mentioned [here](https://github.com/Azure/azure-iot-cli-extension/blob/11d6e4f9a2cb3bd8e69741d5fe9548a60943fb30/azext_iot/operations/hub.py#L801), So it won't be in lowercase.

anusapan:
@jongio Could you let me know the exact reason behind build failure and how could I able to fix this?

jongio:
@c-ryan-k Should be able to help.  Looks like CI dependencies aren't configured correctly.

azext_iot/common/_azure.py
iluican: It's irrelevant that the connection string is generated as of right now, as that can change in the future. As Jon suggested, let's just write unit tests to make sure that we accept any combination of lower/upper case inputs.

digimaun: The intent of this PR is to add support of module SAS token generation to the existing `az iot hub generate-sas-token` command. Some test examples for devices are [here](https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/tests/test_iot_ext_int.py#L344) you can follow them for the module sas token tests and expand on them. Remember to include updated documentation and examples.

For supporting case insensitive connection string input, I recommend submitting a new PR with the concise implementation of that. Currently connection string case sensitivity is strictly enforced.  That being said its a relatively smaller item to support effort wise.

Connection string generation for the user should be strict with respect to specs and documentation as in the casing should remain the same.

digimaun:
> @jongio Could you let me know the exact reason behind build failure and how could I able to fix this?

@anusapan, to start you off, CLICoreValidationTests step is failing because the Azure CLI Core linter has updated its rules to flag the variable pattern we are using in the IoT extension help yaml as non compliant. I.e. the [IoTHub Name] in `az iot edge set-modules --hub-name [IoTHub Name]`...we use that variable pattern everywhere for examples.

Because we use the latest CLI core linter in our CI, these surprises occasionally happen. This linter is also run when submitting to the extension index - better to find out now than later.

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -506,6 +506,15 @@ def test_hub_modules(self):
                          self.exists('authentication.symmetricKey.primaryKey'),
                          self.exists('authentication.symmetricKey.secondaryKey')])
 
+        # Error can't get a sas token for module without device
+        self.cmd('az iot hub generate-sas-token -n {} -g {} -m {}'
+                 .format(LIVE_HUB, LIVE_RG, module_ids[1]), expect_failure=True)
+
+        # sas token for module
+        self.cmd('iot hub generate-sas-token -n {} -g {} -d {} -m {}'
```
digimaun: Remember to add a test to generate-sas-token with connection string like [here](https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/tests/test_iot_ext_int.py#L358).

anusapan: Added

c-ryan-k:
@anusapan - I was tinkering a bit with travis-ci trying to figure out what's happening to pytest-cov…

Can you try editing your dev_requirements file to specify `pytest-cov==2.6.0`?

I haven't made changes to my code to include the helpfile updates, but mine is at least passing the unit tests.

The only other change I've made is to remove the `-q` switch from `pip install -r dev_requirements` so I can see that pytest-cov actually gets installed.

azext_iot/common/_azure.py
anusapan: Fixed case-sensitive issue

digimaun:
Can you please confirm **all** integration tests have passed when you are done making changes.

azext_iot/common/_azure.py
```diff
@@ -13,9 +13,10 @@
 
 def _parse_connection_string(cs, validate=None, cstring_type='entity'):
     decomposed = validate_key_value_pairs(cs)
+    decomposed_lower = dict((k.lower(), v) for k, v in decomposed.items())
     if validate:
         for k in validate:
-            if not decomposed.get(k):
+            if not decomposed.get(k) and not decomposed_lower.get(k.lower()):
```
digimaun: Consider this:  `if not any([decomposed.get(k), decomposed_lower.get(k.lower())]):`

anusapan: fixed

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -506,6 +506,20 @@ def test_hub_modules(self):
...
```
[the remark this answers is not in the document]
anusapan: added a test with using mixed-case CS

anusapan:
> Can you please confirm **all** integration tests have passed when you are done making changes.

Successfully passing all unit and integration tests

PR #82, opened 2019-07-24
IoT Central commands

Jackbk:
---

### General Guidelines

- [-] Have **all** unit **and** integration tests passed locally? i.e. `pytest -s <project root>/azext_iot/tests`

I have had trouble getting all of the integration tests to run properly, but have run the relevant ones to what I have changed (monitor events).


azext_iot/operations/events3/_events.py
```diff
@@ -76,33 +77,57 @@ def _get_conn_props():
...
 
-    if not target.get('events'):
-        endpoint = _build_iothub_amqp_endpoint_from_target(target)
-        _, update = await evaluate_redirect(endpoint)
-        target['events'] = update['events']
+    if not target:
+        logger.debug('No Event Hub target found provided.')
+        return
+
+    print(target)
```
digimaun: Assuming left over debug statement

digimaun:
Please add a synopsis of your changes to HISTORY.rst.

azext_iot/common/basic_sas_token_auth.py
```diff
@@ -0,0 +1,36 @@
...
+# --------------------------------------------------------------------------------------------
+
+"""
+basic_sas_token_auth: Module containing Basic Shared Access Signature token class.
+
+"""
+from msrest.authentication import Authentication
+
+
+class BasicSasTokenAuthentication(Authentication):
```
digimaun: Rather than make a new module for this variant of auth, please put and reference this class in the existing `sas_token_auth.py` module.

azext_iot/operations/central.py
```diff
@@ -0,0 +1,65 @@
...
+    target = {'entity': endpoint}
+    auth = BasicSasTokenAuthentication(sas_token=sasToken)
+    service_sdk, errors = _bind_sdk(target, SdkType.service_sdk, auth=auth)
+    try:
+        return service_sdk.get_twin(device_id)
+    except errors.CloudError as e:
+        raise CLIError(unpack_msrest_error(e))
+
+
+def iot_central_monitor_events(cmd, app_id, device_id=None, consumer_group='$Default', timeout=300, enqueued_time=None,
```
digimaun: This command uses a lot of the same "setup" elements from iot_hub_monitor_events. Consider combining these elements in a reusable way (i.e. something like `def _init_monitoring(...):`

azext_iot/common/_azure.py
```diff
@@ -9,6 +9,25 @@
 from azext_iot.common.utility import validate_key_value_pairs
 from azext_iot.common.utility import trim_from_start
 from azext_iot._factory import iot_hub_service_factory
+from azure.cli.core._profile import Profile
+
+
+def get_aad_token(cmd, subscription=None, resource=None):
```
digimaun: This function is used by another function in the module `get_iot_central_tokens` and not from external (other modules in the package). It makes sense to hint as private i.e. prefix with `_`.

digimaun: When should subscription have a value? The only usage omits the subscription argument. Consider removing the parameter if its not used.

azext_iot/common/_azure.py
```diff
@@ -9,6 +9,25 @@
 from azext_iot.common.utility import validate_key_value_pairs
 from azext_iot.common.utility import trim_from_start
 from azext_iot._factory import iot_hub_service_factory
+from azure.cli.core._profile import Profile
+
```
digimaun: nit: If you are so inclined, please rename azext_iot/common/_azure.py to azext_iot/common/azure.py (no underscore). The modules usage evolved overtime doesn't need a underscore now. If not, we'll do it  in a subsequent PR.

azext_iot/operations/central.py
```diff
@@ -0,0 +1,65 @@
...
+    config = cmd.cli_ctx.config
+    output = cmd.cli_ctx.invocation.data.get("output", None)
+    if not output:
+        output = 'json'
+    ensure_uamqp(config, repair)
+
+    events3 = importlib.import_module('azext_iot.operations.events3._events')
+
+    if not properties:
+        properties = []
```
digimaun: In iothub monitor events `properties = set((key.lower() for key in properties))`  is used to deal with dupes in the input property collection.

azext_iot/operations/events3/_events.py
```diff
@@ -76,33 +77,56 @@ def _get_conn_props():
         properties["platform"] = sys.platform
         return properties
 
-    if not target.get('events'):
-        endpoint = _build_iothub_amqp_endpoint_from_target(target)
-        _, update = await evaluate_redirect(endpoint)
-        target['events'] = update['events']
+    if not target:
+        logger.debug('No Event Hub target found provided.')
```
digimaun: English/grammar needs fix.

azext_iot/operations/events3/_events.py
```diff
@@ -76,33 +77,56 @@ def _get_conn_props():
...
+        return
+
+    if not target.get('central'):
```
digimaun: If checking existence of key, its more efficient to do `if 'central' not in target:`

azext_iot/tests/test_iot_central_int.py
```diff
@@ -0,0 +1,51 @@
...
+        # Verify incorrect device-id throws error
+        self.cmd('az iotcentral device show --app-id "{}"  --device-id incorrect-device  --aad-token {}'.
+                 format(APP_ID, AAD_TOKEN), expect_failure=True)
+        # Verify that no errors are thrown when device shown
+        # We cannot verify that the result is correct, as the Azure CLI for IoT Central does not support adding devices
+        self.cmd('az iotcentral device show --app-id "{}"  --device-id "{}"  --aad-token "{}"'.
+                 format(APP_ID, DEVICE_ID, AAD_TOKEN), expect_failure=False)
+
+    def test_central_monitor_events(self):
+        # Test with invalid aad token
```
digimaun: Thanks for commenting the purpose of each test.

azext_iot/operations/events3/_events.py
```diff
@@ -76,33 +77,56 @@ def _get_conn_props():
...
+            for i in range(int(partition_count)):
+                partition_ids.append(str(i))
+            target['events']['partition_ids'] = partition_ids
+        else:
+            endpoint = target['events']['endpoint']
+            path = target['events']['path']
+        partitions = target['events']['partition_ids']
         auth = _build_auth_container(target)
-        meta_data = await query_meta_data(target['events']['address'], target['events']['path'], auth)
+    else:
```
digimaun: Could you spend some time optimizing this pattern. It is getting unwieldy + repeated code. Consider building a class/transformation component that intakes a `target` understands KPI's with respect to central or vanilla iothub and outputs an object whose attributes can easily be fed into the monitor_events co-routine?

azext_iot/operations/central.py
```diff
@@ -0,0 +1,65 @@
...
+
+
+def iot_central_device_show(cmd, device_id, app_id, aad_token=None):
+    sasToken = get_iot_hub_token_from_central_app_id(cmd, app_id, aad_token)
+    endpoint = find_between(sasToken, 'SharedAccessSignature sr=', '&sig=')
+    target = {'entity': endpoint}
+    auth = BasicSasTokenAuthentication(sas_token=sasToken)
+    service_sdk, errors = _bind_sdk(target, SdkType.service_sdk, auth=auth)
+    try:
+        return service_sdk.get_twin(device_id)
```
digimaun: The command is to show an iot central device, but the description and actual API method call is to get the twin. This is inconsistent with vanilla iothub where there are separate commands for device identity and device twin.

Why not `az iotcentral device-twin show`? 

Is this inconsistency intended? There is value in ensuring iothub and iotcentral command groups work similarly.

Jackbk: I have updated it to be az iotcentral device-twin show :)

azext_iot/common/_azure.py
Jackbk: I tried to do this, but I ran into issues with it conflicting with the Azure module, it may be best to do this separately :)

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -124,8 +124,9 @@ def _remove_entities(self, names=None):
                              .format(i, LIVE_HUB, LIVE_RG), checks=self.is_empty())
 
     def tearDown(self):
-        if self._entity_names:
-            self._remove_entities()
+        # if self._entity_names:
```
digimaun: This shouldn't be commented out. The teardown cleans up after the test.

azext_iot/operations/central.py
```diff
@@ -0,0 +1,53 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+# pylint: disable=wrong-import-order,too-many-lines
+
+import asyncio
```
digimaun: A top level import of asyncio is not going to work. We use asyncio functionality for Python 3.5+ ....by adding import here the module is no longer compatible with Python 2.7.

Jackbk: Ah, thanks for letting me know.  How is it used in the events library? Is it restricted to commands with the python 3.5+ check?

azext_iot/operations/events3/_events.py
```diff
@@ -66,44 +68,88 @@ def stop_and_suppress_eloop():
...
-    if not partitions:
+    if not target['partitions']:
         logger.debug('No Event Hub partitions found to listen on.')
         return
 
     coroutines = []
-
-    auth = _build_auth_container(target)
-    async with uamqp.ConnectionAsync(target['events']['endpoint'], sasl=auth,
+    print('big boiiiiii')
```
digimaun: debug statement

azext_iot/operations/central.py
digimaun: Yea thats right. The pattern is to check that we have a compatible runtime (3.5+) then import the constrained elements such as `asyncio` and `uamqp`.

Jackbk: > Yea thats right. The pattern is to check that we have a compatible runtime (3.5+) then import the constrained elements such as `asyncio` and `uamqp`.

Hmm interesting, I'll move stuff around so it works as expected then.  Thanks~!

PR #83, opened 2019-07-29
Merged PnP stuff from pnp-cli-publicpreview

anusapan:
---

### General Guidelines


azure-pipelines.yml
```diff
@@ -0,0 +1,38 @@
+# Starter pipeline
```
c-ryan-k: Are we using Azure Pipelines as well as Travis for builds, or was this just leftover from a test build?

anusapan: I don't think that we are using azure-pipelines, as this file was the part of PnP-Cli-publicpreview repo.

digimaun: Yes, we were using Azure Pipelines for private preview.

digimaun: Should be deleted if we are not making changes to the CI engine for this release.

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
+        --interface {plug_and_play_interface} --device-id {device_id}
+        --command-name {command_name} --command-payload {payload}
+    - name: Execute a command on device within current session.
+      text: >
+        az iot dt invoke-command --hub-name {iothub_name}
+        --interface {plug_and_play_interface} --device-id {device_id}
+        --command-name {command_name} --command-payload {payload}
+"""
+
+helps['iot dt list-interfaces'] = """
```
digimaun: Why is digitaltwin in respective commands shortened to 'dt' but capability-model (like `iot pnp capability-model create`) spelled out? Also in this [PnP CLI article](https://github.com/MicrosoftDocs/azure-docs-pr/pull/81679/files), the digitaltwin example commands are spelled out, like `az iot digitaltwin list-devices`. Let's be consistent with our command set in the extension and also make sure docs are accurate.

anusapan: According to the latest discussion with Ilie, we have to use "dt"(digitaltwin) in commands which are interacting with digitaltwin instances and use "pnp" for model repo commands. As far as concern with "capability-model", this was asked me to use this instead of "model".

> **Previous**: az iot pnp model create

> **Now**: az iot pnp capability-model create

I don't have the access on "PnP CLI article"

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,227 @@
...
+        raise CLIError('Target interface is not implemented by the device!')
+
+    found_commands = _pnp_interface_elements(cmd, target_interface[0]['urn_id'], INTERFACE_COMMAND,
+                                             repo_endpoint, repo_id, repo_login)
+
+    if command_name not in [c.get('name') for c in found_commands if c.get('name') == command_name]:
+        raise CLIError('Command \'{}\' is not defined in the target interface.'.format(command_name))
+
+    target_command = [c for c in found_commands if c.get('name') == command_name]
+    if schema:
```
YingXue: Do we still need to keep this --schema parameter? What is this trying to do?

digimaun: The purpose of --schema (at least in private preview) was a shortcut to see the definition of a particular command or property.

anusapan: it will result the schema of command

azext_iot/_help.py
digimaun: Thanks for your explanation @anusapan. If that is the decision then the document needs to be changed to reflect this update. Also in general we should be mindful how long our commands get.

digimaun: +@iluican

digimaun: Ok the [latest preview docs](https://github.com/MicrosoftDocs/azure-docs-pr/blob/release-preview-adb-pnp/articles/iot-pnp/howto-install-pnp-cli.md) have 'dt' as you have here, but `capability-model` is still `model`.

azext_iot/tests/test_device_digitaltwin_invoke_command.json
```diff
@@ -0,0 +1,5 @@
+{
+    "blinkRequest": {
```
YingXue: is this the payload for this command? 
      {
        "@type": "Command",
        "commandType": "synchronous",
        "description": "This command will begin blinking the LED for given time interval.",
        "name": "blink",
        "request": {
          "name": "blinkRequest",
          "schema": "long"
        },
        "response": {
          "name": "blinkResponse",
          "schema": "string"
        }
      }
If so, the payload should simply be 1. We fixed the bug in the UI, where we should not wrap the name "blinkRequest" in command payload. So you would need to remove "blinkRequest". But I am not sure where did this "interval" key come from.

azext_iot/operations/events3/_events.py
```diff
@@ -26,11 +26,16 @@
 DEBUG = True
 
 
-def executor(target, consumer_group, enqueued_time, device_id=None, properties=None, timeout=0, output=None, content_type=None,
-             devices=None):
+# TODO: Shortcuts for PnP
```
YingXue: comment needs to be cleaned up here

azext_iot/operations/digitaltwin.py
YingXue: I am not sure if we should keep this to be honest.
For public preview, there are so many sources of model definition. I am assuming what we want to do here is, if user don't specify repo, get schema from public repo, and if so, get it from private repo.
But there are other sources as well, of which "get model definition from device itself" is what we we are going to enable on the UI. I am not sure if what works before still serves it well today, so I was trying to eliminating this complexity from CLI, or we will need to support a more sophiscated configuration of sources like our UI does. 
What do you say @iluican

YingXue: And in my personal opinion, we should simply remove it, since we are not supporting this --schema parameter for property, why only commands then.

azext_iot/operations/events3/_events.py
```diff
@@ -137,15 +147,21 @@ def _output_msg_kpi(msg):
...
+                    return
+
         event_source = {'event': {}}
 
         event_source['event']['origin'] = origin
 
-        payload = ''
-
-        data = msg.get_data()
-        if data:
```
digimaun: What's deleted here is the newer pattern. `payload = str(next(msg.get_data()), 'utf8')` was the older one. There is a chance iterating msg.get_data() could return None.

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
+      text: >
+        az iot dt monitor-events --login {iothub_cs} --device-id {device_id}
+        --interface {plug_and_play_interface} --consumer-group {consumer_group_name}
+        --properties all
+"""
+
+helps['iot dt update-property'] = """
+    type: command
+    short-summary: Update an IoT Plug and Play device interfaces read-write property.
+    examples:
```
digimaun: nit: long-summary missing from some of the new commands.

azext_iot/_params.py
```diff
@@ -378,3 +385,70 @@ def load_arguments(self, _):
...
+    with self.argument_context('iot pnp interface') as context:
+        context.argument('interface_definition', options_list=['--definition', '--def'],
+                         help='IoT Plug and Play interface definition written in PPDL (JSON-LD). '
+                         'Can be directly input or a file path where the content is extracted.')
+
+    with self.argument_context('iot pnp interface list') as context:
+        context.argument('search_string', options_list=['--search', '--ss'],
+                         help='Searches IoT Plug and Play interfaces for given string in the'
+                              ' \"Description, DisplayName, comment and Id\".')
+        context.argument('page_size', options_list=['--page', '--ps'],
```
digimaun: Does it make sense that the end user needs to think about `page_size` on a list operation? Unless this is intended in behavior to be like `top` in our other list operations (then it should be called `top`)? If its really an API page size, as in chunks of elements to pull at a time to collect the whole collection, then it would be better to hide from the user.

digimaun: Point is it looks like an implementation detail.

azext_iot/_params.py
```diff
@@ -378,3 +385,70 @@ def load_arguments(self, _):
...
+                         help='Page Size.')
+        context.argument('token', options_list=['--token', '-t'],
```
digimaun: Same with --token. This should be abstracted from users.

azext_iot/_params.py
```diff
@@ -378,3 +385,70 @@ def load_arguments(self, _):
...
+    with self.argument_context('iot pnp capability-model show') as context:
+        context.argument('expand', options_list=['--expand'],
+                         help='Indicates whether to expand the device capability model\'s'
+                              ' interface definitions or not.')
+
+    with self.argument_context('iot pnp capability-model list') as context:
+        context.argument('search_string', options_list=['--search', '--ss'],
+                         help='Searches IoT Plug and Play models for given string in the'
+                              ' \"Description, DisplayName, comment and Id\".')
+        context.argument('page_size', options_list=['--page', '--ps'],
```
digimaun: Same comment as above regarding `page_size`. Take a look at how the iot hub query functionality [is implemented](https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/operations/generic.py). Token and page size is abstracted from the end user. A 'top' argument is leveraged instead.

anusapan: replace "top" in place of page_size and token

azext_iot/_params.py
anusapan: replace "top" in place of page_size and token

azext_iot/_params.py
anusapan: replace "top" in place of page_size and token

azext_iot/operations/events3/_events.py
anusapan: reverted

azext_iot/operations/events3/_events.py
anusapan: removed temp comments

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,227 @@
...
+from azext_iot.common.shared import SdkType
+from azext_iot.common._azure import get_iot_hub_connection_string
+from azext_iot.common.utility import (shell_safe_json_parse,
+                                      unpack_msrest_error)
+from azext_iot.operations.pnp import (iot_pnp_interface_show,
+                                      iot_pnp_interface_list)
+from azext_iot.operations.hub import iot_hub_monitor_events
+
+
+DEFAULT_INTERFACE_KEY_NAME = 'urn_azureiot_ModelDiscovery_DigitalTwin'
```
digimaun: nit: If the interface key isn't configurable just call it `INTERFACE_KEY_NAME`.

azext_iot/common/utility.py
```diff
@@ -280,3 +294,14 @@ def calculate_millisec_since_unix_epoch_utc():
     now = datetime.utcnow()
     epoch = datetime.utcfromtimestamp(0)
     return int(1000 * (now - epoch).total_seconds())
+
+
+def get_sas_token(target):
```
digimaun: This function is so specific to Digital Twin auth. It makes sense to add a class method to `DigitalTwinSasTokenAuthentication` so you can do something like `DigitalTwinSasTokenAuthentication.from_target(target)` where `from_target` is a class method.

digimaun: Also...don't need to do it in this PR, but we should think about shortening the names of the auth elements...they are quite verbose especially when the module name h as a lot of the same info as the class name.

azext_iot/operations/hub.py
```diff
@@ -1284,8 +1284,10 @@ def iot_device_upload_file(cmd, device_id, file_path, content_type, hub_name=Non
         raise CLIError(e)
 
 
+# TODO: Shortcut for PnP
 # pylint: disable=too-many-locals
-def iot_hub_monitor_events(cmd, hub_name=None, device_id=None, consumer_group='$Default', timeout=300,
+def iot_hub_monitor_events(cmd, interface=None, pnp_context=None,
```
digimaun: The pnp_context argument at this layer was part of a private preview shortcut pattern. pnp_context shouldn't show up at the top level (by top level I mean the function associated with a command). Ideally a user running `az iot hub monitor-events` has no knowledge of pnp or a pnp_context, it's a less optimal user experience having it here.

digimaun: Consider a lower level "private" function that intakes pnp_context (among the other elements). BTW some improvements  to event monitoring setup is in work in the iot central PR.

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
+helps['iot dt monitor-events'] = """
+    type: command
+    short-summary: Monitor Plug & Play device interface events.
+    long-summary: You can leverage az login and provide --hub-name instead of --login for every command.
+    examples:
+    - name: Monitor an IoT Plug and Play device events.
+      text: >
+        az iot dt monitor-events --login {iothub_cs} --device-id {device_id}
+        --interface {plug_and_play_interface} --consumer-group {consumer_group_name}
+    - name: Monitor an IoT Plug and Play device events within current session.
```
digimaun: The monitoring descriptions + examples should probably replace `IoT Plug and Play` to  `Digital Twin` events. Especially when `dt` is in the command.

azext_iot/operations/hub.py
digimaun: Same idea for `interface`.

azext_iot/_params.py
```diff
@@ -121,6 +123,11 @@ def load_arguments(self, _):
         context.argument('content_type', options_list=['--content-type', '--ct'],
                          help='Specify the Content-Type of the message payload to automatically format the output to that type.')
         context.argument('device_query', options_list=['--device-query', '-q'], help='Specify a custom query to filter devices.')
+        context.argument('interface', options_list=['--interface', '-i'],
```
digimaun: Please see my below comments regarding the elements.

azext_iot/_params.py
```diff
@@ -378,3 +385,66 @@ def load_arguments(self, _):
...
+        context.argument('schema', options_list=['--schema', '-s'],
+                         arg_type=get_three_state_flag(),
+                         help='Switch to skip the operation and retrieve the PnP repository '
+                         'interface entity schema. Default: false')
+
+    with self.argument_context('iot dt monitor-events') as context:
+        context.argument('repair', options_list=['--repair'],
+                         arg_type=get_three_state_flag(),
+                         help='Reinstall uamqp dependency compatible with extension version.')
+        context.argument('consumer_group', options_list=['--consumer-group', '--cg'],
```
digimaun: `interface` should be moved here along with other relevant monitoring arguments. Also consider moving common monitoring arguments (if there are any) with single intention to the base `iot` command group.

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,227 @@
...
+
+def iot_digitaltwin_monitor_events(cmd, device_id, interface, repo_endpoint=PNP_ENDPOINT, repo_id=None,
+                                   consumer_group='$Default', timeout=300, hub_name=None, resource_group_name=None,
+                                   yes=False, properties=None, repair=False, login=None, repo_login=None):
+    target_interfaces = []
+    pnp_context = {'enabled': True, 'interface': {}}
+    device_interfaces = _iot_digitaltwin_interface_list(cmd, device_id, hub_name, resource_group_name, login)
+    interface_list = _get_device_default_interface_dict(device_interfaces)
+    interface_count = len([k for k in interface_list if k['name'] == interface])
+    if interface and interface_count == 0:
```
digimaun: Regarding `interface_count` the list is already being iterated below extending `target_interfaces`. So move the statement on line# 168 up and then this check can be `if interface and not target_interfaces`.  Also is it a valid use case to have multiple interfaces with the same name? If so how is that supposed to work here? You monitor the same interface but different versions?

anusapan: fixed

azext_iot/_params.py
anusapan: fixed as mentioned in the comment

azext_iot/_params.py
anusapan: fixed as mentioned in comments

azext_iot/_help.py
anusapan: changed to digitaltwin

azext_iot/operations/hub.py
anusapan: remove pnp_context and interface, create new private method

azext_iot/common/utility.py
anusapan: correct, I'll issue a new PR against it.

azext_iot/operations/digitaltwin.py
anusapan: Done

azext_iot/operations/digitaltwin.py
anusapan: Implemented as per the mail conversation

azure-pipelines.yml
anusapan: deleted

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
 """
+
+helps['iot dt'] = """
+    type: group
+    short-summary: Manage digital twin of an IoT Plug and Play device.
+"""
+
+helps['iot dt invoke-command'] = """
+    type: command
+    short-summary: Executes a command on an IoT Plug and Play device.
```
digimaun: This command and other `iot dt` based commands don't mentioned "Digital Twin" in their summaries. Do we want to update summaries (and example names) to include Digital Twin language?

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
+"""
+
+helps['iot dt list-properties'] = """
+    type: command
+    short-summary: List properties of a target IoT Plug and Play device interface(s).
+    long-summary: You can leverage az login and provide --hub-name instead of --login for every command.
+    examples:
+    - name: List all properties of all interfaces on an IoT Plug and Play device.
+      text: >
+        az iot dt list-properties --login {iothub_cs} --source device
```
digimaun: For new users it helps describe every core element of a commands usage example, particularly new concepts such as `source`. I suggest explaining in the example what `-- source device` means. As well as `private`, `public` etc.

azext_iot/_params.py
```diff
@@ -378,3 +381,70 @@ def load_arguments(self, _):
...
+        context.argument('command_name', options_list=['--command-name', '--cn'],
+                         help='IoT Plug and Play interface command name.')
+        context.argument('command_payload', options_list=['--command-payload', '--cp', '--cv'],
+                         help='IoT Plug and Play interface command payload. '
+                         'Content can be directly input or extracted from a file path.')
+        context.argument('interface_payload', options_list=['--interface-payload', '--ip', '--iv'],
+                         help='IoT Plug and Play interface payload. '
+                         'Content can be directly input or extracted from a file path.')
+        context.argument('source_model', options_list=['--source', '-s'],
+                         help='Source to get model definition. ',
```
digimaun: Is it well known/well documented how to use or configure `source_model`? Err on the side of more details if it makes sense.

azext_iot/_help.py
```diff
@@ -912,3 +912,253 @@
...
+    - name: List all public interfaces.
+      text: >
+        az iot pnp interface list
+"""
+
+helps['iot pnp interface show'] = """
+    type: command
+    short-summary: Get the details of an interface.
+    examples:
+    - name: Get the details of a private repository interface.
```
digimaun: For list/show of interface or cm, there is specific reference to public or private repos. It shows different usage of the command for public vs private. Contrasting with create/update/publish there is just repo, without consideration of public/private. Is this intended?

anusapan: Yes, create/update/publish is applicable for private repo only

azext_iot/_help.py
anusapan: updated help file

azext_iot/_params.py
anusapan: updated the text

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,276 @@
...
+    for k, v in interface['properties']['modelInformation']['reported']['value']['interfaces'].items():
+        result.append({'name': k, "urn_id": v})
+    return result
+
+
+def _pnp_interface_elements(cmd, interface, target_type, repo_endpoint, repo_id, login):
+    interface_elements = []
+    results = iot_pnp_interface_list(cmd, repo_endpoint, repo_id, interface, login=login)
+    results_count = len(results)
+    if results_count > 0:
```
digimaun: can be simplified to `if results:`

azext_iot/operations/pnp.py
```diff
@@ -0,0 +1,243 @@
...
+
+logger = get_logger(__name__)
+
+
+def iot_pnp_interface_publish(cmd, interface, repo_endpoint=PNP_ENDPOINT, repo_id=None, login=None):
+    _validate_repository(repo_id, login)
+    model_list = _iot_pnp_model_list(cmd, repo_endpoint, repo_id, interface, PnPModelType.interface,
+                                     -1, login=login)
+    model_list_count = len(model_list)
+    if model_list_count > 0 and model_list[0].urn_id == interface:
```
digimaun: `if model_list and model_list[0].urn_id == interface`, then remove `model_list_count`

anusapan: Done

azext_iot/operations/digitaltwin.py
anusapan: Done

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,275 @@
...
+
+
+def iot_digitaltwin_properties_list(cmd, device_id, source_model, interface=None, schema=False,
+                                    repo_endpoint=PNP_ENDPOINT, repo_id=None, repo_login=None,
+                                    hub_name=None, resource_group_name=None, login=None):
+    result = []
+    target_interfaces = []
+    device_interfaces = _iot_digitaltwin_interface_list(cmd, device_id, hub_name, resource_group_name, login)
+    interface_list = _get_device_default_interface_dict(device_interfaces)
+    interface_count = len([k for k in interface_list if k['name'] == interface])
```
digimaun: In both `iot_digitaltwin_properties_list` and  `iot_digitaltwin_command_list` looks like there are opportunities for refactoring (similar that was done to `iot_digitaltwin_monitor_events`). Idea being to iterate over `interface_list` once. I am still not sure if multiple interfaces with the same name is a valid use case (differentiating between version?). So based on that, you can adjust accordingly.

Example, which is assuming single interface of a particular name.

```python
interface_list = _get_device_default_interface_dict(device_interfaces)
target_interface = next((item for item in interface_list if item['name'] == interface), None)
if interface and not target_interface:
    raise CLIError('Target interface is not implemented by the device!')

target_interfaces.append(target_interface)
...
```

Example, assuming there can be multiple interfaces on a digital twin with the same name (differentiated by version):

```python
interface_list = _get_device_default_interface_dict(device_interfaces)
target_interfaces = [k for k in interface_list if k['name'] == interface]
if interface and not target_interfaces:
  raise CLIError('Target interface is not implemented by the device!')

# You already have interface_list so no need to extend
if not target_interfaces:
    target_interfaces = interface_list

...
```

azext_iot/tests/test_iot_pnp_int.py
```diff
@@ -0,0 +1,183 @@
...
+
+
+def change_dir():
+    from inspect import getsourcefile
+    os.chdir(os.path.dirname(os.path.abspath(getsourcefile(lambda: 0))))
+
+
+class TestPnPModel(LiveScenarioTest):
+
+    LiveScenarioTest.rand_val = random.randint(1, 10001)
```
digimaun: You don't need to set this on the base class. You should be able to have just `rand_val = random.randint(1, 10001)` and be able to reference `TestPnPModel.rand_val`.

azext_iot/tests/test_iot_pnp_int.py
```diff
@@ -0,0 +1,183 @@
...
+            self.kwargs.update({'model_id': _model_id})
+            model_newContent = model.replace(json.loads(model)['@id'], self.kwargs['model_id'])
+            model_newContent = model_newContent.replace('\n', '')
+
+            fo = open(self.kwargs['model'], "w+", encoding='utf-8')
+            fo.write(model_newContent)
+            fo.close()
+
+    # TODO: @anusapan - Maybe put a helper like this in the shared lib, when you create it?
+    def command_execute_assert(self, command, asserts):
```
digimaun: It looks like you are not using this particular helper. Safe to delete?

azext_iot/tests/test_iot_pnp_int.py
```diff
@@ -0,0 +1,183 @@
...
+            os.remove(self.kwargs['interface'])
+        if exists(self.kwargs['model']):
+            os.remove(self.kwargs['model'])
+
+    def test_interface_life_cycle(self):
+
+        # Error: missing repo-id or login
+        self.cmd('iot pnp interface create -e {endpoint} --def {interface}', expect_failure=True)
+
+        # Error: Invalid Interface definition file
```
digimaun: Thanks for commenting the expectation.

azext_iot/tests/test_iot_pnp_int.py
```diff
@@ -0,0 +1,183 @@
...
+        interface = str(read_file_content(self.kwargs['interface']))
+        display_name = json.loads(interface)['displayName']
+        interface_newContent = interface.replace(display_name, '{}-Updated'.format(display_name))
+        interface_newContent = interface_newContent.replace('\n', '')
+        fo = open(self.kwargs['interface-updated'], "w+", encoding='utf-8')
+        fo.write(interface_newContent)
+        fo.close()
+        self.cmd('iot pnp interface update -e {endpoint} -r {repo} --def {interface-updated}', checks=self.is_empty())
+
+        # Todo: Publish Interface
```
digimaun: Please add a test for Interface Publish.

azext_iot/tests/test_device_digitaltwin_invoke_command.json
digimaun: @anusapan , please address this -- my only concern being we have accurate test assets (payloads, interfaces, models etc)

azext_iot/tests/test_iot_pnp_int.py
```diff
@@ -0,0 +1,183 @@
...
+        model = str(read_file_content(self.kwargs['model']))
+        display_name = json.loads(model)['displayName']
+        model_newContent = model.replace(display_name, '{}-Updated'.format(display_name))
+        model_newContent = model_newContent.replace('\n', '')
+        fo = open(self.kwargs['model-updated'], "w+", encoding='utf-8')
+        fo.write(model_newContent)
+        fo.close()
+        self.cmd('iot pnp capability-model update -e {endpoint} -r {repo} --def {model-updated}', checks=self.is_empty())
+
+        # Todo: Publish Capability-Model
```
digimaun: Please add a test for capability model Publish.

azext_iot/operations/digitaltwin.py
```diff
@@ -0,0 +1,275 @@
...
+                                                               hub_name, resource_group_name, login)
+    result = _get_device_default_interface_dict(device_default_interface)
+    return {'interfaces': result}
+
+
+def iot_digitaltwin_command_list(cmd, device_id, source_model, interface=None, schema=False,
+                                 repo_endpoint=PNP_ENDPOINT, repo_id=None, repo_login=None,
+                                 hub_name=None, resource_group_name=None, login=None):
+    result = []
+    target_interfaces = []
```
digimaun: Consider adding `source_model = source_model.lower()` here, so you only `lower()` once on it.

azext_iot/operations/digitaltwin.py
anusapan: Expecting only one interface, not version based interfaces as of now. So changed accordingly

azext_iot/tests/test_iot_pnp_int.py
anusapan: Done

azext_iot/tests/test_iot_pnp_int.py
anusapan: Removed

azext_iot/tests/test_iot_pnp_int.py
anusapan: Added the test

azext_iot/tests/test_iot_pnp_int.py
anusapan: Added

azext_iot/operations/digitaltwin.py
anusapan: Done

PR #105, opened 2019-10-10
Initial DevOps pipeline configuration.

digimaun:
This is part 1 of replacing legacy Travis CI scripts with DevOps Pipelines. Currently this will run unit tests (in all supported environments), style checks and CLI core linter. This already does more than our legacy CI. Running integration tests is possibly coming in a follow-up PR after some recorded test issues are resolved.

image

image


.azure-devops/templates/build-publish-azure-cli-test-sdk.yml
```diff
@@ -0,0 +1,21 @@
+steps:
+  - script: 'rm -rf azure-cli'
+    displayName: 'delete azure cli directory'
+
+  - script: 'git clone https://github.com/Azure/azure-cli.git'
```
YingXue: with my limited knowledge, does pipeline always to clean clone, in the sense of it would not try to pull and then clone if needed?
It seems not needed at this point, but maybe you want a clone resursive just in case for the future.

digimaun: I'm leaning towards lets add --recursive when we need it rather than preemptively.

PR #121, opened 2019-11-21
Support layered edge deployment + ADM enhancement

digimaun:
* Rework of *all* related tests, unit and integration
* New tests and components
* Rework of command logic
  * Still need another *future* phase of this which is extraction of the configurations logic from the hub module and put into a dedicated class (manager/handler)
* Better validation handling and error output format
* Updated help + param contents

Here is what the new validation error format looks like for edge deployments:

image

---

### General Guidelines


HISTORY.rst
```diff
@@ -3,12 +3,19 @@
...
++++++++++++++++
+- Support IoT Edge layered deployments
+- Support ADM module twin definitions
+- Improved json schema validation error handling
+- Update top maximum for config list to 200
+
 0.8.6
 +++++++++++++++
-- For IoTHub commands - improves json handling for arguments that require json.
+- For IotHub commands - improves json handling for arguments that require json.
```
montgomp: Product name should be "IoT Hub"

azext_iot/operations/hub.py
```diff
@@ -557,95 +561,97 @@ def _iot_hub_configuration_create(cmd, config_id, content, hub_name=None, target
...
+    from knack.util import to_snake_case
+
+    # Supports scenario where configuration payload is contained in 'content' key
+    if "content" in content:
+        content = content["content"]
+
+    # Create new config dict to remove superflous properties
+    processed_content = {}
+    if config_type == ConfigType.adm:
+        valid_adm_keys = ["deviceContent", "moduleContent"]
```
montgomp: If a user submits both "deviceContent" and "moduleContent"we should error out rather than assuming they wanted deviceContent.

HISTORY.rst
```diff
@@ -3,12 +3,19 @@
...
+- Update top maximum for config list to 200
+
 0.8.6
 +++++++++++++++
-- For IoTHub commands - improves json handling for arguments that require json.
+- For IoT Hub commands - improves json handling for arguments that require json.
 - Edge deployments support metric definitions at creation time (like device configurations)
 - Fixes issue with `az iot hub invoke-device-method` preventing primitive value payloads.
-- The `az iot device simulate` command will send default values for content-type and content-encoding. These values can be overriden.
+- The `az iot device simulate` command will send default values for content-type and content-encoding. These values can be overridden.
```
YingXue: Have a english wording question. Is it more appropriate to say overwritten or overridden here?

azext_iot/operations/hub.py
```diff
@@ -734,10 +741,10 @@ def iot_hub_configuration_delete(cmd, config_id, hub_name=None, resource_group_n
         raise CLIError(err)
 
 
-def iot_edge_deployment_metric_show(cmd, config_id, metric_id,
+def iot_edge_deployment_metric_show(cmd, config_id, metric_id, metric_type='user',
```
YingXue: Is this a requirement change

HISTORY.rst
digimaun: Changing to "overwritten"

azext_iot/operations/hub.py
digimaun: We talked about this - so we can ensure both the configuration and edge show-metric work exactly the same. We'd have to make a behavior change either way since the old version of edge show-metric was specifically for system metrics.

PR #122, opened 2019-11-22
Jabark/iot central environments

Jackbk:
Updated iot central commands to allow the api for token collection to be overridden.  This will allow us to use this extension in dev and ppe environments.

---

### General Guidelines


azext_iot/_params.py
```diff
@@ -412,6 +412,9 @@ def load_arguments(self, _):
         context.argument('repair', options_list=['--repair', '-r'],
                          arg_type=get_three_state_flag(),
                          help='Reinstall uamqp dependency compatible with extension version. Default: false')
+        context.argument('central_api_uri', options_list=['--central-api-uri'],
+                         help='IoT Central API override.  For use with environments other than production. '
+                         'Default: api.azureiotcentral.com.')
```
digimaun: Are you sure you need to add the 'Default: api.azureiotcentral.com' part of the string? The knack arg parser should automatically add the default value if its assigned to an arg in the command bound function. Please double check.

digimaun: The other explicit 'Default: blah' are for some limited cases *usually* for the 'three state flag' args i.e. when you require switches/flags.

azext_iot/_params.py
```diff
@@ -420,9 +423,15 @@ def load_arguments(self, _):
     with self.argument_context('iotcentral device-twin show') as context:
```
digimaun: Just curious is there a removal of the deprecated 'iotcentral' in favor of 'iot central' in the works?

azext_iot/_params.py
Jackbk: Yep you are right, I have removed and tested it.  Thanks!

azext_iot/_params.py
Jackbk: I asked about this, it's still planned but no work has started on it yet.  The IoT Central team is focusing on stabilizing the ignite features first

PR #130, opened 2020-01-08
Jobs v2 command set

digimaun:
The PR has 2 main components

- Adds the Jobs v2 command set
- Refactors `SasTokenAuthentication` behavior to be a bit easier to use and to support session injection.
- Also, CI runs unit tests on Python 3.8.

---

### General Guidelines

### DIFF
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -162,14 +162,16 @@ def load_arguments(self, _):
                          'If no start time is provided, the job is queued for asap execution.')
         context.argument('ttl', options_list=['--ttl'], type=int,
                          help='Max execution time in seconds, before job is terminated.')
-        context.argument('twin_patch', options_list=['--twin-patch'],
+        context.argument('twin_patch', options_list=['--twin-patch', '--patch'],
                          help='The desired twin patch. Provide file path or raw json.')
         context.argument('wait', options_list=['--wait', '-w'],
                          arg_type=get_three_state_flag(),
                          help='Block until the created job is in a completed, failed or cancelled state. '
                          'Will regularly poll on interval specified by --poll-interval.')
-        context.argument('poll_interval', options_list=['--poll-interval'], type=int,
+        context.argument('poll_interval', options_list=['--poll-interval', '--interval'], type=int,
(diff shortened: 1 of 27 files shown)


azext_iot/_params.py
```diff
@@ -138,6 +144,33 @@ def load_arguments(self, _):
...
+        context.argument('job_status', options_list=['--job-status', '--js'],
+                         help='The status of a scheduled job.',
+                         arg_type=get_enum_type(JobStatusType))
+        context.argument('job_type', options_list=['--job-type', '--jt'],
+                         help='The type of scheduled job.',
+                         arg_type=get_enum_type(JobType))
+        context.argument('query_condition', options_list=['--query-condition', '-q'],
+                         help='Condition for device query to get devices to execute the job on. '
+                         'Required if job type is scheduleDeviceMethod or scheduleUpdateTwin. '
+                         'Note: The service will prefix "SELECT * FROM devices WHERE " to the input')
```
montgomp: Isn't this going to SELECT * from jobs?

azext_iot/iothub/job.py
```diff
@@ -0,0 +1,178 @@
...
+    target = get_iot_hub_connection_string(
+        cmd, hub_name, resource_group_name, login=login
+    )
+    service_sdk, errors = _bind_sdk(target, SdkType.service_sdk)
+
+    try:
+        job_result = service_sdk.create_job1(id=job_id, job_request=job_request)
+        if wait:
+            logger.info("Waiting for job completion state...")
+            while True:
```
montgomp: It seems there should be more control flow here. Thinks that come to mind are if `job_result` immediately returned in line 105 with an acceptable status, we shouldn't need to sleep and re-query. Also, it seems like we could end up in an indefinite sleep/poll.

digimaun: We talked about this - if the job request is accepted by the API it will always returned a queued status object. If no `--wait` is specified the status object is returned to the caller. To mitigate the infinite loop issue, we will introduce a `--poll-max-duration <minutes>` argument.

digimaun: Opted for `--poll-duration <seconds>` to keep the same units between interval and duration.

PR #140, opened 2020-02-21
Add ADM module config target condition validation

digimaun:
* Add ADM module config target condition validation.
* Second attempt at removing "--config-id" from edge deployments.
  * This should be safe since we renamed the extension.

---

### General Guidelines

### DIFF
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -343,9 +343,8 @@ def load_arguments(self, _):
                          help='Maximum number of configurations to return.')
 
     with self.argument_context('iot edge') as context:
-        context.argument('config_id', options_list=['--deployment-id', '-d', '--config-id', '-c'],
-                         help="Target deployment name. Option '--config-id' has been deprecated and will be "
-                         "removed in a future release. Use '--deployment-id' instead.")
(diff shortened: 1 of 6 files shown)


azext_iot/_params.py
```diff
@@ -343,9 +343,8 @@ def load_arguments(self, _):
                          help='Maximum number of configurations to return.')
 
     with self.argument_context('iot edge') as context:
-        context.argument('config_id', options_list=['--deployment-id', '-d', '--config-id', '-c'],
-                         help="Target deployment name. Option '--config-id' has been deprecated and will be "
-                         "removed in a future release. Use '--deployment-id' instead.")
+        context.argument('config_id', options_list=['--deployment-id', '-d'],
```
YingXue: since --config-id is the one being removed, should we change context.argument('config_id', options_list=['--deployment-id', '-d'] to context.argument('deployment_id', options_list=['--deployment-id', '-d']?

digimaun: Both edge deployments and ADM ultimately share the same API's, so API wise its a "config Id" ...but no issues wanting to update to the identifier here. We wanted to separate --deployment-id/--config-id for the different edge deployment/hub configuration command spaces (from the perspective of the user).

PR #155, opened 2020-04-07
Add support for az iot central app validate-messages command

prbans:
---

### General Guidelines


azext_iot/_params.py
```diff
@@ -30,531 +30,1035 @@
...
+        context.argument("app_id", options_list=["--app-id"], help="Target App.")
+        context.argument(
+            "central_api_uri",
+            options_list=["--central-api-uri"],
+            help="IoT Central API override.  For use with environments other than production.",
+        )
+
+    with self.argument_context("iot central app validate-messages") as context:
+        context.argument(
+            "device_id", options_list=["--device-id", "-d"], help="Target Device."
```
digimaun, line 899: You don't need to redefine param/argument definitions that were defined in a parent command group. In this case `device_id` has been defined in root `iot`.

azext_iot/_params.py
```diff
@@ -30,531 +30,1035 @@
...
+            "enqueued_time",
+            options_list=["--enqueued-time", "--et", "-e"],
+            type=int,
+            help="Indicates the time that should be used as a starting point to read messages from the partitions. "
+            "Units are milliseconds since unix epoch. "
+            'If no time is indicated "now" is used.',
+        )
+        context.argument("properties", arg_type=event_msg_prop_type)
+        context.argument(
+            "content_type",
```
digimaun, line 929: Please go through the rest of these and remove any redundancies. content_type is another example.

azext_iot/_params.py
```diff
@@ -30,531 +30,1035 @@
...
+        )
+        context.argument(
+            "yes",
+            options_list=["--yes", "-y"],
+            arg_type=get_three_state_flag(),
+            help="Skip user prompts. Indicates acceptance of dependency installation (if required). "
+            "Used primarily for automation scenarios. Default: false",
+        )
+
+    with self.argument_context("iotcentral device-twin show") as context:
```
digimaun, line 878: When is `iotcentral` going away? It is unfortunate to still see deprecated remnants.

azext_iot/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.9.1"
+VERSION = "255.9.4"
```
digimaun, line 10: Why change the version convention/pattern?

azext_iot/models/parsers.py
```diff
@@ -0,0 +1,252 @@
...
+class Event3Parser(object):
+    _info = []
+    _warnings = []
+    _errors = []
+    _logger = get_logger(__name__)
+
+    def __init__(self, logger=None):
+        if logger:
+            self._logger = logger
+        pass
```
digimaun, line 24: Do you really need this `pass`?

dev_requirements
```diff
@@ -3,6 +3,7 @@ pytest-mock
 pytest-cov
 pytest-env
 pylint
+uamqp
```
digimaun, line 6: This should be `uamqp~=1.2`.

azext_iot/models/parsers.py
```diff
@@ -0,0 +1,252 @@
...
+from uamqp.message import Message
+from azext_iot.common.utility import parse_entity, unicode_binary_map
+
+SUPPORTED_ENCODINGS = ["utf-8"]
+DEVICE_ID_IDENTIFIER = b"iothub-connection-device-id"
+INTERFACE_NAME_IDENTIFIER = b"iothub-interface-name"
+random.seed(0)
+
+
+class Event3Parser(object):
```
digimaun, line 15: Good idea to encapsulate the message stream processing in class like `Event3Parser`.

azext_iot/models/parsers.py
```diff
@@ -0,0 +1,252 @@
+import random
```
digimaun, line 1: Don't forget the encoding and license headers.

azext_iot/_params.py
prbans: Chatted offline

azext_iot/models/parsers.py
```diff
@@ -0,0 +1,248 @@
...
+    ) -> str:
+        message_interface_name = ""
+
+        try:
+            message_interface_name = str(
+                message.annotations.get(INTERFACE_NAME_IDENTIFIER), "utf8"
+            )
+        except Exception:
+            self._errors.append(
+                f"Unable to parse interface_name given a pnp_device. {origin_device_id}. "
```
digimaun, line 135: Can you please verify if f strings work in python 3.5? Last I recall it was min python 3.6. We have to support at least python 3.5.

azext_iot/models/parsers.py
```diff
@@ -0,0 +1,248 @@
...
+            self._warnings.append(
+                "Content type not found in system_properties. "
+                f"System_properties: {system_properties}"
+            )
+
+        if "application/json" not in content_type.lower():
+            self._warnings.append(
+                "Content type not supported. "
+                f"Content type found: {content_type}. "
+                "Content type expected: application/json. "
```
digimaun, line 207: Theoretically regular IoT Hub telemetry does not require content type. Also, general IoT Hub wise application/json is not necessarily required, but you do need application/json if you want to route messages.

azext_iot/operations/events3/_events.py
```diff
@@ -401,3 +355,51 @@ def handle_msg(msg):
...
+    devices,
+    pnp_context,
+    interface_name,
+    content_type,
+    properties,
+    output,
+    validate_messages,
+    simulate_errors,
+):
+    parser = Event3Parser()
```
digimaun, line 372: Can we refactor the parser initialization so we don't instantiate a new Event3Parser for every message?

azext_iot/operations/events3/_events.py
```diff
@@ -401,3 +355,51 @@ def handle_msg(msg):
...
+        dump = yaml.safe_dump(parsed_msg, default_flow_style=False)
+
+    if validate_messages:
+        parser.write_logs()
+
+    if not validate_messages:
+        six.print_(dump, flush=True)
+
+
+def _should_process_device(origin_device_id, device_id, devices):
```
digimaun, line 394: Good stuff breaking this out.

azext_iot/_help.py
```diff
@@ -1031,22 +1209,58 @@
         az iot central app monitor-events --app-id {app_id} --output json
   """
 
-helps['iot central device-twin'] = """
+
+helps[
+    "iot central app validate-messages"
```
digimaun, line 1214: In reviewing, I am not sure if it makes sense for `validate-messages` to be a separate command. It may be more apt and concise to have a `--validate-messages` / `--validate` / `--vm` switch against `iot central app monitor-events` as its practically the same flow. Why does this need to be a separate command?

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -48,103 +57,232 @@ def get_twin(self, device_id):
...
     def test_monitor_events_invalid_args(self, timeout, exception, fixture_cmd):
         with pytest.raises(exception):
             subject.iot_central_monitor_events(fixture_cmd, app_id, timeout=timeout)
+
+
+@pytest.mark.skipif(
+    not validate_min_python_version(3, 5, exit_on_fail=False),
+    reason="minimum python version not satisfied",
+)
+class TestEvents3Parser:
```
digimaun, line 181: Good parser test pattern and its clear Event3Parser simplifies Message testing.

azext_iot/models/parsers.py
prbans: Ah good catch, they are only available in 3.6 and up. I might update the contribution docs to suggest spinning up the env using python 3.5 or below for backwards compat

azext_iot/operations/events3/_events.py
prbans: How strongly do you feel about this? I think this will be tricky due to thread safety issues since these messages are being parsed async.

The issue is that the error/warning arrays are encapsulated within the class so that the caller doesn't have to worry about implementation details. Since we need to clear these each time a new message comes in, we risk deleting the errors/warnings from another message.

I also believe the extra memory used by this shouldn't be a concern. And since we don't really hold on to the EventParser object, it should get garbage collected once the function returns

digimaun: nevermind, on second look validation is done per device event, so this is OK.

prbans: #wontfix

azext_iot/_help.py
prbans: long story short: because this is just the first iteration on this command. The longer term goal of this command warrants it to have its own namespace

prbans:
TODO: screenshot of messaging_int tests passing

azext_iot/operations/hub.py
```diff
@@ -9,25 +9,27 @@
...
+    shell_safe_json_parse,
+    read_file_content,
+    validate_key_value_pairs,
+    url_encode_dict,
+    unpack_msrest_error,
+    init_monitoring,
+    process_json_arg,
+)
 from azext_iot._factory import SdkResolver, CloudError
+from azext_iot.operations import events3
```
digimaun, line 32: There are way too many formatting changes in this module to difference between actual logic changes.

Can you save me time and itemize the core changes along with this one?

Also, while you are here, you could clean-up usage of `importlib` that's a legacy remnant from supporting Python 2.7 and preventing conflicts from static checkers. Should be replaced with natural import syntax since we are 3.5+ now.

prbans: Sounds like we're on the same page, the only change I was making here was removing `importlib`.

Went ahead and did it for the rest of the file too

prbans:
> 
> 
> TODO: screenshot of messaging_int tests passing

image

azext_iot/models/parsers.py
prbans: I've got ideas on how to make this better - i.e. allow callers to specify severity of various errors. Lets leave that for a follow up PR. For now I'll chalk this up to the needs of validate-messages and monitor-events being different

prbans: #wontfix

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -48,103 +57,235 @@ def get_twin(self, device_id):
...
+        )
+        parser = _parser.Event3Parser()
+
+        # act
+        parsed_msg = parser.parse_message(message, None, None, None, None, False)
+
+        # verify
+        assert parsed_msg["event"]["payload"] == self.payload
+        assert parsed_msg["event"]["origin"] == device_id
+
```
digimaun, line 208: Can we assert message properties as well

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -48,103 +57,235 @@ def get_twin(self, device_id):
...
+        message = Message(
+            body=json.dumps(self.payload).encode(),
+            properties=properties,
+            annotations={_parser.DEVICE_ID_IDENTIFIER: device_id.encode()},
+        )
+        parser = _parser.Event3Parser()
+
+        # act
+        parsed_msg = parser.parse_message(message, None, None, None, None, False)
+
```
digimaun, line 204: Let's also add a case for parsing interface name from a PnP event.

digimaun:
Please add to (and/or improve) central integration tests with the new command. At the least executing the command should not throw. You don't have to test cases already covered by the IoTHub messaging tests.

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -48,103 +57,235 @@ def get_twin(self, device_id):
...
     def test_monitor_events_invalid_args(self, timeout, exception, fixture_cmd):
         with pytest.raises(exception):
             subject.iot_central_monitor_events(fixture_cmd, app_id, timeout=timeout)
+
+
+@pytest.mark.skipif(
+    not validate_min_python_version(3, 5, exit_on_fail=False),
+    reason="minimum python version not satisfied",
+)
+class TestEvents3Parser:
```
digimaun, line 181: I'm not thrilled with the parser tests living in the central unit test module since it's a generic IoT Hub message parser. But I suppose we can revisit this later.

prbans: I'd be more than happy to move it for you, I figured it was a good idea to add it to either this or the IoTHub unit tests since they are more likely to get run than something just called event parser unit tests

digimaun: It would be great if you could move it to `test_iot_utility_unit.py`.

azext_iot/tests/test_iot_central_int.py
```diff
@@ -11,41 +11,62 @@
...
         # We cannot verify that the result is correct, as the Azure CLI for IoT Central does not support adding devices
-        self.cmd('az iotcentral device-twin show --app-id "{}" --device-id "{}"'.
-                 format(APP_ID, DEVICE_ID), expect_failure=False)
-        self.cmd('az iot central device-twin show --app-id "{}" --device-id "{}"'.
-                 format(APP_ID, DEVICE_ID), expect_failure=False)
+        self.cmd(
+            'az iotcentral device-twin show --app-id "{}" --device-id "{}"'.format(
+                APP_ID, DEVICE_ID
+            ),
+            expect_failure=False,
```
digimaun, line 45: *nit, expect_failure is default `False` (I know this existed before)

prbans:
Utility tests passing:

image

IOTC unit tests passing:

image

prbans:
iot hub integration tests passing

image

PR #161, opened 2020-04-14
Validate field names match DCM

prbans:
This PR has 2 purposes:

1) Add ability to retrieve DCM from central, and validate that messages sent by device match this DCM

2) Propose a new way to organize code in repo:

```
azext_iot/
|--dps
|--central
  |-- providers
  |-- services
|--hub
...
|-- common
|-- utility
```
---

### General Guidelines


prbans:
Tests passing

image

azext_iot/tests/test_iot_utility_unit.py
```diff
@@ -1,36 +1,51 @@
...
 
 
 class TestMinPython(object):
     @pytest.mark.parametrize("pymajor, pyminor", [(3, 6), (3, 4), (2, 7)])
-    def test_min_python(self, mocker, pymajor, pyminor):
-        version_mock = mocker.patch("azext_iot.common.utility.sys.version_info")
-        version_mock.major = pymajor
-        version_mock.minor = pyminor
+    def test_min_python(self, pymajor, pyminor):
+        with mock.patch("azext_iot.common.utility.sys.version_info") as version_mock:
```
prbans, line 32: @digimaun - this is the fix for the VSCode debugger I was mentioning

azext_iot/central/services/device_template.py
```diff
@@ -0,0 +1,50 @@
...
+        device: dict
+    """
+
+    if not token:
+        token = get_token(token, cmd)
+
+    url = "https://{}.{}/api/preview/deviceTemplates/{}".format(
+        app_id, central_dns_suffix, device_template_urn
+    )
+    headers = {"Authorization": token}
```
digimaun, line 41: Would be nice to add the IoT extension user agent to any `requests` based calls.

azext_iot/operations/central.py
```diff
@@ -30,6 +34,15 @@ def iot_central_device_show(
         raise CLIError(unpack_msrest_error(e))
 
 
+def iot_central_device_capability_model_get(
+    cmd, device_id, app_id, central_api_uri="api.azureiotcentral.com"
+):
+    provider = CentralDeviceProvider(cmd, app_id)
+    device_template = provider.get_device_template(device_id)
+    output = json.dumps(device_template, indent=4)
+    six.print_(output)
```
digimaun, line 43: It's for special cases where we print the result. You should return the Python object (which should be something that is compatible with json.dumps()). There are downstream ramifications if you don't. For example, being unable to use the CLI core --query capability.

azext_iot/central/providers/device_provider.py
```diff
@@ -0,0 +1,79 @@
...
+        self._devices = {}
+
+    def get_device_template(
+        self, device_id, central_dns_suffix="azureiotcentral.com",
+    ):
+        device = self.get_device(device_id, central_dns_suffix)
+        device_template_urn = device["instanceOf"]
+
+        if not device_template_urn:
+            raise ValueError(
```
digimaun, line 36: It's good and pythonic that you raise exceptions. However from a CLI devEx perspective we don't want to show a stack trace to the user - which is what will happen with an uncaught exception. Using `CLIError` will hide the stack trace. You could catch the true underlying exception at a higher level and wrap it in a CLIError.

digimaun: There are various raised exceptions through-out this PR. Please review and apply the feedback.

azext_iot/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.9.3"
+VERSION = "0.9.4"
```
digimaun, line 10: The next release would actually be `0.9.3`. We recently released `0.9.2`.

azext_iot/operations/central.py
```diff
@@ -43,19 +56,21 @@ def iot_central_validate_messages(
...
-        consumer_group,
-        timeout,
-        enqueued_time,
-        repair,
-        properties,
-        yes,
-        central_api_uri,
+        cmd=cmd,
+        app_id=app_id,
+        device_id=device_id,
```
digimaun, line 63: thanks, using keyword args is good practice when the argument list grows unwieldly.

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -169,3 +173,56 @@ class TestMonitorEvents:
     def test_monitor_events_invalid_args(self, timeout, exception, fixture_cmd):
         with pytest.raises(exception):
             subject.iot_central_monitor_events(fixture_cmd, app_id, timeout=timeout)
+
+
+@pytest.mark.skipif(
+    not validate_min_python_version(3, 5, exit_on_fail=False),
+    reason="minimum python version not satisfied",
```
digimaun, line 180: We can remove the validate_min_python_version checks now since min 3.5 is assumed.

azext_iot/central/services/device_template.py
```diff
@@ -0,0 +1,50 @@
...
+from ._utility import get_token
+
+
+def get_device_template(
+    cmd,
+    device_template_urn: str,
+    app_id: str,
+    token: str,
+    central_dns_suffix="azureiotcentral.com",
+) -> str:
```
digimaun, line 20: type hint: `dict`?

azext_iot/central/providers/device_provider.py
```diff
@@ -0,0 +1,79 @@
...
+                device_template_urn
+            ] = central_services.device_template.get_device_template(
+                self._cmd,
+                device_template_urn,
+                self._app_id,
+                self._token,
+                central_dns_suffix,
+            )
+
+        device_template = self._device_templates.get(device_template_urn)
```
digimaun, line 56: `.get()` is slower than `self._device_templates[device_template_urn]`.  You know the key exists in the dictionary from `get_device_template` right?

azext_iot/commands.py
```diff
@@ -188,6 +188,13 @@ def load_command_table(self, _):
             "validate-messages", "iot_central_validate_messages", is_preview=True
         )
 
+    with self.command_group(
+        "iot central app capability-model", command_type=iotcentral_ops
+    ) as cmd_group:
+        cmd_group.command(
+            "get", "iot_central_device_capability_model_get", is_preview=True
```
digimaun, line 195: It seems like the whole `az iot central app` command group should be set to is_preview.

prbans: Monitor-events is not `is_preview`

azext_iot/operations/events3/_parser.py
```diff
@@ -240,20 +270,50 @@ def _parse_payload(
...
+                    "List of allowed telemetry values for this type of device: {}. "
+                    "NOTE: telemetry names are CASE-SENSITIVE".format(
+                        telemetry_name, origin_device_id, all_names
+                    )
                 )
-            )
+
+    def _extract_schema_from_template(self, template):
+        all_schema = []
+        dcm = template["capabilityModel"]
```
digimaun, line 313: Assuming there's no chance the template format is off. We need to be careful about crashing the monitor from malformed data.

azext_iot/commands.py
digimaun: Double checking on this - that is intended right? central app monitor-events is not supposed to be in preview? Please resolve the comment if true.

azext_iot/commands.py
```diff
@@ -188,6 +188,13 @@ def load_command_table(self, _):
             "validate-messages", "iot_central_validate_messages", is_preview=True
         )
 
+    with self.command_group(
+        "iot central app capability-model", command_type=iotcentral_ops
```
digimaun, line 192: Please remember to add the necessary help content for this command. Use the shared one for now.

azext_iot/commands.py
prbans: Confirmed with Jack, it is NOT in preview

azext_iot/central/providers/device_provider.py
prbans: oh good catch, I think if the code reaches here and its still missing from catch then we should throw a KeyError and let the handler handle it

PR #171, opened 2020-04-25
Add provisioning filters

valluriraj:
---

### General Guidelines


azext_iot/central/commands_device.py
```diff
@@ -7,6 +7,7 @@
 
 from knack.util import CLIError
 from azext_iot.central.providers import CentralDeviceProvider
+from azext_iot.common.shared import DeviceStatus
```
prbans, line 10: remove - unused

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
 
         return info
 
-    def get_all_registration_info(self, central_dns_suffix="azureiotcentral.com"):
+    def filter_dps_info(self, dps_info, device_status):
```
prbans, line 204: what is the expected behavior if device_status is None?

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+            "registered": "Device it not yet provisioned.",
+            "blocked": "Device is blocked by admin",
+            "unassociated": "Device does not have a valid template associated with it",
+        }
+
+        filtered_dps_info = {
+            "status": dps_info.get("status"),
+        }
+
+        filtered_dps_info.update({"error": error.get(device_status)})
```
prbans, line 216: nit: do it this way:

```
filtered_dps_info = {
  "status": dps_info.get("status"),
  "error": error.get(device_status)
}
```

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+                    updated_device = self.update_device_info_filter_status(
+                        device, DeviceStatus.registered.value
+                    )
+                else:
+                    updated_device = self.update_device_info_filter_status(
+                        device, DeviceStatus.provisioned.value
+                    )
+        return updated_device
+
+    def update_all_device_status(self, devices):
```
prbans, line 252: try to make this function name a bit more specific, since we might need to do different kinds of updates in the future

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
             device for device in devices.values() if not device["simulated"]
         ]
-        if len(devices) != len(real_devices):
+
+        filtered_devices = self.filter_device_list(real_devices, device_status)
+
+        if len(devices) != len(filtered_devices):
             logger.warning(
                 "Getting registration info for following devices. "
                 "All other devices are simulated. "
```
prbans, line 284: nit: update this comment

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+        self, device_status, central_dns_suffix="azureiotcentral.com"
+    ):
+
         logger.warning("This command may take a long time to complete execution.")
         devices = self.list_devices(central_dns_suffix=central_dns_suffix)
         real_devices = [
             device for device in devices.values() if not device["simulated"]
         ]
-        if len(devices) != len(real_devices):
+
```
prbans, line 278: I think the flow here should be something like:

```
real_devices_with_status = [self.add_device_status(device) for device in real_devices]

filtered_devices = real_devices_with_status
if device_status:
  filtered_devices = [
      device 
      for device in devices
      # remove devices without central device provisioning information
      # as that means the device does not have DPS information
      if device.get("deviceStatus") and device.get("deviceStatus") == device_status
  ]
```

azext_iot/central/providers/device_provider.py
prbans: `"All other devices are simulated. "`

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+            "id": device["id"],
+            "displayName": device.get("displayName"),
+            "instanceOf": device.get("instanceOf"),
+            "simulated": device.get("simulated"),
+            "deviceStatus": value,
+        }
+        return updated_device_data
+
+    def update_device_status(self, device):
+        if device["approved"] is False:
```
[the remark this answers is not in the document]
prbans: infact, I think this code could be moved into the `get_device` code:

```
device = central_services.device.get_device(
                cmd=self._cmd,
                app_id=self._app_id,
                device_id=device_id,
                token=self._token,
                central_dns_suffix=central_dns_suffix,
            )
device_status = self.determine_device_status(device)
if device_status:
  device["deviceStatus"] = device_status
```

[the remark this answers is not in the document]
prbans: or perhaps even at the service level

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+
+        filtered_dps_info.update({"error": error.get(device_status)})
+        return filtered_dps_info
+
+    def update_device_info_filter_status(self, device, value):
+        if value is None:
+            return self.update_device_status(device)
+        updated_device_data = {
+            "id": device["id"],
+            "displayName": device.get("displayName"),
```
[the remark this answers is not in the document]
valluriraj: yes in some cases displayName is not present and so using get works better

azext_iot/central/providers/device_provider.py
valluriraj: device_status should have a value of DeviceStatus and in which case all the values for enum are handled. Passing in None will result in get returning NULL which is acceptable.

azext_iot/central/providers/device_provider.py
```diff
@@ -194,20 +201,92 @@ def get_device_registration_info(
...
+        }
+
+        filtered_dps_info = {
+            "status": dps_info.get("status"),
+        }
+
+        filtered_dps_info.update({"error": error.get(device_status)})
+        return filtered_dps_info
+
+    def update_device_info_filter_status(self, device, value):
```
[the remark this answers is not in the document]
valluriraj: this step does include adding only the selected fields so current flow will be the best solution here.

azext_iot/central/providers/device_provider.py
valluriraj: Restricted the call to devices that are provisioned.

azext_iot/central/providers/device_provider.py
```diff
@@ -170,44 +171,91 @@ def get_device_credentials(
...
+
+        real_devices_with_status = [
+            central_services.device.update_device_status(device)
+            for device in real_devices
+        ]
+
+        filtered_devices = real_devices_with_status
+
+        if device_status:
+            filtered_devices = [
```
prbans, line 245: Thanks for doing this!

azext_iot/central/providers/device_provider.py
```diff
@@ -170,44 +171,91 @@ def get_device_credentials(
...
+            ):
+                credentials = self.get_device_credentials(
+                    device_id=device_id, central_dns_suffix=central_dns_suffix
+                )
+                id_scope = credentials["idScope"]
+                key = credentials["symmetricKey"]["primaryKey"]
+                dps_state = dps_global_service.get_registration_state(
+                    id_scope=id_scope, key=key, device_id=device_id
+                )
+            dps_state = self.dps_populate_essential_info(
```
prbans, line 199: I like these function names a lot more thanks

azext_iot/central/params.py
```diff
@@ -60,3 +61,9 @@ def load_central_arguments(self, _):
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
+        context.argument(
+            "device_status",
+            options_list=["--devicestatus", "-s"],
```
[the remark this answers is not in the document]
digimaun: Agreed on this nit. Thanks.

azext_iot/central/providers/device_provider.py
```diff
@@ -170,44 +171,91 @@ def get_device_credentials(
         return credentials
 
     def get_device_registration_info(
-        self, device_id, central_dns_suffix="azureiotcentral.com",
+        self,
+        device_id,
+        device_status: DeviceStatus,
+        central_dns_suffix="azureiotcentral.com",
     ):
+        dps_state = {}
```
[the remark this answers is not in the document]
valluriraj: state is only populated for provisioned only - all other cases use the default initialized value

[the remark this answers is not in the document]
prbans: ok, might be good to move inside the `if not info` check in that case - still a nit though can leave as wont fix

valluriraj:
image

azext_iot/central/providers/device_provider.py
```diff
@@ -170,44 +171,91 @@ def get_device_credentials(
...
 
         self._device_registration_info[device_id] = info
 
         return info
 
-    def get_all_registration_info(self, central_dns_suffix="azureiotcentral.com"):
+    def dps_populate_essential_info(self, dps_info, device_status):
+        error = {
+            "provisioned": "None",
+            "registered": "Device it not yet provisioned.",
```
ppathan, line 215: typo: `Device is not yet provisioned.`

azext_iot/central/commands_device.py
```diff
@@ -68,10 +68,16 @@ def registration_info(
     device_id=None,
     token=None,
     central_dns_suffix="azureiotcentral.com",
+    device_status=None,
 ):
-    provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token)
+    provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token,)
```
ppathan, line 73: Why is there a `,` at the end?

azext_iot/central/commands_device.py
```diff
@@ -68,10 +68,16 @@ def registration_info(
...
     if not device_id:
-        return provider.get_all_registration_info(central_dns_suffix=central_dns_suffix)
+        return provider.get_all_registration_info(
+            central_dns_suffix=central_dns_suffix, device_status=device_status
```
ppathan, line 76: What pattern are we following, new line for parameter or on the same line?

azext_iot/central/commands_device.py
prbans: We use an auto-formatter called "Black", that is adding this additional comma at the end

azext_iot/central/commands_device.py
prbans: We use an auto-formatter called "Black", that tool is deciding the spacing for us

azext_iot/central/commands_device.py
prbans: Though in this particular case it should probably be removed

digimaun: I would not worry about this one. This is standard `python black` behavior, in this case adding a comma after the last kwarg element. If you remove it now (which is fine), the next change to this module and a re-run of the formatter would re-introduce it.

`python black` is an opinionated formatter with minimal configuration. Idea for using `Black` is to save time and mental energy for more important matters. By using it, you agree to cede control over minutiae of hand-formatting. Also `Black` will check that the reformatted code still produces a valid AST that is equivalent to the original. It's a _relatively_ newer Python tool that has [immense popularity](https://github.com/psf/black/network/dependents?package_id=UGFja2FnZS01MDMzMzg4OQ%3D%3D) in the open source world.

azext_iot/central/providers/device_provider.py
digimaun, line 215: nit: The `registered` value string has a period at the end `.`  It would be good to standardize on ending with a period or not.

azext_iot/tests/test_iot_central_int.py
```diff
@@ -204,5 +204,41 @@ def test_central_device_registration_info(self):
...
+            checks=[
+                self.check("approved", True),
+                self.check("displayName", device_name),
+                self.check("id", device_id),
+                self.check("simulated", False),
+            ],
+        )
+
+        result = self.cmd(
+            "iot central app device registration-info --app-id {} --ds {}".format(
```
digimaun, line 228: Do we have the necessary foundation in place to exercise assertions against devices with other possible device status values?

prbans: Unfortunately we don't, iot central creating a device doesn't actually provision it so its somewhat difficult to get the device into the other states programatically during an IT.

I'll look into this as a follow up and see if theres anything we can do

azext_iot/central/providers/device_provider.py
prbans: I'd suggest that cleaning up error messages is a P2 since I have a task in our backlog to clean up all error messages once we are ready to ship.

Let me know if this is something PR blocking and I can work with you to resolve

azext_iot/tests/test_iot_central_int.py
valluriraj: One option is to create an app with existing devices in different states and  then adding validations. IT will take dependency on external environment. Will  explore other options as well.

PR #179, opened 2020-05-07
Central validate messages ux cleanup

prbans:
Monitor code updates -

* Moved core central validation logic to its own parser and handler
* Made validation error framework cleaner
* Updated error messages

Misc - 

* Move all central commands and params to new pattern
* Collected all "deprecation" entry points into their own functions for easier deletion
---

### General Guidelines


azext_iot/monitor/parsers/issue.py
```diff
@@ -0,0 +1,234 @@
...
+
+
+SUPPORTED_ENCODING = ["utf-8"]
+SUPPORTED_FIELD_NAME_CHARS = ["a-z", "A-Z", "0-9", "_"]
+SUPPORTED_CONTENT_TYPE = ["application/json"]
+SUPPORTED_MESSAGE_HEADERS = []
+
+
+class IssueMessageBuilder:
+    @staticmethod
```
digimaun, line 128: We do not want to use staticmethod. Preference is for module level functions instead. Static methods are rare and often from legacy libraries. The father of Python has mentioned in the past that [static methods were an accident](https://mail.python.org/pipermail/python-ideas/2012-May/014969.html). Not using staticmethods is also inline with Azure Python SDK guidelines.

azext_iot/monitor/handlers/common_handler.py
```diff
@@ -0,0 +1,75 @@
...
+        device_id       (str)   only process messages sent by this device
+        devices         (list)  only process messages sent by these devices
+        pnp_context     (bool)  interpret the device as being a pnp device
+        interface_name  (str)   expected interface name of pnp device
+        content_type    (str)   assumed content type (utf-8, ascii, etc)
+        properties      (list)  list of properties to extract from message headers
+        output          (str)   output format (json, yaml, etc)
+    """
+
+    def __init__(self, **kwargs):
```
digimaun, line 30: I'm not crazy about this kwargs style. Normally kwargs are used for flags, options and less commonly used arguments.

azext_iot/central/params.py
```diff
@@ -57,13 +77,69 @@ def load_central_arguments(self, _):
...
             "central_dns_suffix",
-            options_list=["--central-dns-suffix"],
+            options_list=["--central-dns-suffix", "--central-api-uri"],
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
         context.argument(
             "device_status",
-            options_list=["--devicestatus", "--ds"],
+            options_list=["--device-status", "--ds"],
```
digimaun, line 86: Thanks.

azext_iot/monitor/handlers/central_handler.py
```diff
@@ -0,0 +1,120 @@
...
+        self.progress_interval = kwargs.get("progress_interval") or 5
+        self.minimum_severity = kwargs.get("minimum_severity") or Severity.warning
+
+        self.messages = []
+        self.issues: List[Issue] = []
+
+        if self.timeout:
+            loop = get_loop()
+            # the monitor takes a few seconds to start
+            loop.call_later(self.timeout + 5, self._quit_on_timeout)
```
digimaun, line 45: This timeout is different from the currently released timeout. The original intent for timeout is to indicate how long to maintain a connection without receiving message and its handled at the amqp client level. Versus, here where its a static timeout and close connection. Is that intended?

azext_iot/central/providers/monitor_provider.py
```diff
@@ -0,0 +1,113 @@
...
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from knack.util import CLIError
+
+from azext_iot.central.providers import CentralDeviceProvider
+from azext_iot.common.utility import init_monitoring
+
+
+class MonitorProvider:
```
digimaun, line 13: Are you intending for this PR to also have vanilla IoT Hub monitoring able to use MonitorProvider directly? Just wondering since there are some central specific things here.

azext_iot/monitor/base_classes.py
```diff
@@ -0,0 +1,35 @@
...
+class AbstractBaseParser(ABC):
+    def __init__(self, logger=None):
+        super().__init__()
+
+    @abstractmethod
+    def write_logs(self, severity=Severity.info) -> None:
+        raise NotImplementedError()
+
+    @abstractmethod
+    def parse_message(self, message, **kwargs) -> dict:
```
prbans, line 21: remove this - parse_message should only take message, no kwargs

azext_iot/central/providers/monitor_provider.py
prbans: No, I plan to have this monitor provider just be used by central. It should be relatively easy to replicate this pattern in IoT hub once iothub code is moved out of `operations.py`

azext_iot/monitor/handlers/central_handler.py
prbans: Yes, our monitor has slightly different needs. But after chatting with the team earlier today we decided to move it to another variable so that the experience of using --timeout is consistent with the other monitors

prbans:
image

azext_iot/central/params.py
```diff
@@ -57,13 +69,92 @@ def load_central_arguments(self, _):
...
+
+    with self.argument_context("iot central app monitor-events") as context:
+        context.argument("timeout", arg_type=event_timeout_type)
+        context.argument("properties", arg_type=event_msg_prop_type)
+
+    with self.argument_context("iot central app validate-messages") as context:
+        context.argument("timeout", arg_type=event_timeout_type)
+        context.argument("properties", arg_type=event_msg_prop_type)
+        context.argument(
+            "time_range",
```
digimaun, line 91: Why not just name this `duration` or similar?  `--time-range` sounds a bit off to me. Like I'm looking at time series data. `--duration` also seems more inline with the use of `--duration` in other commands.

It would also be good to include the unit in the description.

prbans:
image

prbans:
image

azext_iot/monitor/models/arguments.py
```diff
@@ -0,0 +1,69 @@
...
+        )
+        self.output = output
+        self.timeout = timeout_ms
+        self.properties = unique_properties
+        self.enqueued_time = enqueued_time
+
+
+class CommonParserArguments:
+    def __init__(
+        self, properties=[], interface_name="", content_type="",
```
digimaun, line 33: Let's not use any mutable objects as default values for function parameters in order to avoid issues later on. properties can be `None` here.

azext_iot/monitor/models/arguments.py
```diff
@@ -0,0 +1,69 @@
...
+        self.interface_name = interface_name or ""
+        self.content_type = content_type or ""
+
+
+class CommonHandlerArguments:
+    def __init__(
+        self,
+        output: str,
+        common_parser_args: CommonParserArguments,
+        devices=[],
```
digimaun, line 45: Another place to change the default mutable parameter. Please review the code to make sure its refactored out.

azext_iot/monitor/models/arguments.py
```diff
@@ -0,0 +1,69 @@
...
+        self,
+        cmd: AzCliCommand,
+        timeout: int,
+        properties: list,
+        enqueued_time: int,
+        repair: bool,
+        yes: bool,
+    ):
+        (enqueued_time, unique_properties, timeout_ms, output) = init_monitoring(
+            cmd, timeout, properties, enqueued_time, repair, yes
```
digimaun, line 23: nit: can we use keyword args when calling init_monitoring

prbans: I'm not sure if I can see the benefit of moving this to kwargs, how strongly do you feel about this?

I also feel like the `init_monitoring` call might need some refactoring, its slightly odd that it does stuff like set defaults and return values with the same name as the input params. I was even thinking about putting it somewhere in the `monitor/` path

digimaun: The benefit when used is order of args won't matter which is increased in value when the number of args for a function increase. One less thing to worry about.

This one is just a nit - I am perfectly fine if no change is made.

Yes, I agree on refactoring but not in this PR please.

digimaun: Just so we make sure we are talking about the same thing

I meant 

init_monitoring(cmd=cmd, timeout=timeout, properties=properties, enqueued_time=enqueued_time, repair=repair, yes=yes)

**Not** usage of **kwargs expansion

prbans: OHHH ok, I'm in favor of that, I'll make that fix

prbans:
image

PR #183, opened 2020-05-14
Change error severities, remove dynamic validation logic

prbans:
---

### General Guidelines

### DIFF
--- a/azext_iot/monitor/parsers/common_parser.py
+++ b/azext_iot/monitor/parsers/common_parser.py
@@ -44,10 +44,9 @@ class CommonParser(AbstractBaseParser):
         self._parse_content_encoding(message, system_properties)
 
         content_type = self._parse_content_type(content_type, system_properties)
+        message_interface_name = self._parse_interface_name(message)
 
         if interface_name:
-            message_interface_name = self._parse_interface_name(message, interface_name)
-
             event["interface"] = message_interface_name
 
         if properties:
@@ -88,7 +87,7 @@ class CommonParser(AbstractBaseParser):
             self._add_issue(severity=Severity.error, details=details)
             return ""
 
-    def _parse_interface_name(self, message: Message, expected_interface_name) -> str:
+    def _parse_interface_name(self, message: Message) -> str:
         actual_interface_name = ""
 
         try:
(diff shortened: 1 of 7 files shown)


azext_iot/monitor/parsers/common_parser.py
```diff
@@ -44,10 +44,9 @@ def parse_message(self) -> dict:
         self._parse_content_encoding(message, system_properties)
 
         content_type = self._parse_content_type(content_type, system_properties)
+        message_interface_name = self._parse_interface_name(message)
 
         if interface_name:
-            message_interface_name = self._parse_interface_name(message, interface_name)
```
digimaun, line 49: For the non-central IoT Hub monitoring, If a message has an interface associated with it should be shown. If a target interface argument is provided, then all messages not coming from the interface should be filtered out/not shown.

prbans: Huh - interesting. The old behavior implementation wise says that the interface_name is only parsed out if pnp_context is passed in.

I'd be in favor of locking down the expected behavior here. If you want the behavior to change per your suggestion I can implement that instead

https://github.com/Azure/azure-iot-cli-extension/blob/24428b9faa49daa11e24d002ee89892f12be4c71/azext_iot/operations/events3/_events.py#L198

prbans: since we removed `pnp_context` last iteration, might have introduced this bug as a side-effect

digimaun: That's because there is/was a DT monitoring command (in the `az iot dt` space) and a non DT monitoring command. The DT monitoring command is the one that dealt with PnP context and had these rules.

This play would remove the need for an additional superfluous monitoring command.

prbans: Could you please let me know if my below interpretation of your proposal is correct?

1) Always parse interface_name and add it to return value
2) If "filter_interface_name" is specified, then only return value if actual interface_name matches filter_interface_name
3) Remove `az iot dt monitor-events`

If this is correct, I can address (1) and (2) with this PR. We can file (3) as a backlog item

digimaun: Yes this interpretation is correct.

prbans: ok cool I'll address (1) and (2) with this PR in that case, thanks for the advice!

prbans: Ok I think this one is resolved now, let me know what you think

digimaun: yea looks good

prbans:
image

PR #184, opened 2020-05-14
Provisioning filters updates

valluriraj:
---

### General Guidelines

### DIFF
--- a/azext_iot/central/commands_device.py
+++ b/azext_iot/central/commands_device.py
@@ -69,12 +69,28 @@ def registration_info(
     token=None,
     central_dns_suffix="azureiotcentral.com",
     device_status=None,
+    max_devices=None,
+    summarize_registration=False,
 ):
     provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token,)
     if not device_id:
-        return provider.get_all_registration_info(
+        if summarize_registration:
+            summary = provider.get_registration_summary(
+                central_dns_suffix=central_dns_suffix
+            )
+            return summary
+
+        device_collection = provider.get_all_registration_info(
             central_dns_suffix=central_dns_suffix, device_status=device_status
         )
+
+        if max_devices:
+            provider.print_limited_devices(
+                device_collection=device_collection,
+                devices_to_display=int(max_devices),
+            )
+            return
+        return device_collection
 
     return provider.get_device_registration_info(
         device_id=device_id,
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -86,6 +86,17 @@ def load_central_arguments(self, _):
             arg_type=get_enum_type(DeviceStatus),
             help="Indicates filter option for device status",
         )
+        context.argument(
+            "summarize_registration",
+            options_list=["--registrationsummary", "--rs"],
+            arg_type=get_three_state_flag(),
+            help="Add this flag to get a summary of registration information",
+        )
+        context.argument(
+            "max_devices",
+            options_list=["--maxdevices", "--md"],
+            help="This value sets the maximum numner of devices to be displayed in a single instance",
+        )
 
     with self.argument_context("iot central app monitor-events") as context:
         context.argument("timeout", arg_type=event_timeout_type)
--- a/azext_iot/central/providers/device_provider.py
+++ b/azext_iot/central/providers/device_provider.py
@@ -11,6 +11,7 @@ from azext_iot.central import services as central_services
 from azext_iot.central.models.enum import DeviceStatus
 from azext_iot.central.models.device import Device
 from azext_iot.dps.services import global_service as dps_global_service
+import json
 
 logger = get_logger(__name__)
 
@@ -35,6 +36,14 @@ class CentralDeviceProvider:
         self._device_templates = {}
         self._device_credentials = {}
         self._device_registration_info = {}
+
+    def get_device_essential_info(
+        self, device_id, central_dns_suffix="azureiotcentral.com",
+    ) -> Device:
+        device = self.get_device(device_id, central_dns_suffix)
+        registration_info = device.get_device_registration_info()
+
+        return registration_info
 
     def get_device(
         self, device_id, central_dns_suffix="azureiotcentral.com",
@@ -181,8 +190,12 @@ class CentralDeviceProvider:
         if info:
             return info
 
-        device = self.get_device(device_id)
-        if device.device_status == DeviceStatus.provisioned:
+        device_essential_info = self.get_device_essential_info(device_id)
+
+        if (
+            DeviceStatus(device_essential_info.get("device_status"))
+            == DeviceStatus.provisioned
+        ):
             credentials = self.get_device_credentials(
                 device_id=device_id, central_dns_suffix=central_dns_suffix
             )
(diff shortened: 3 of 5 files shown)


azext_iot/central/commands_device.py
```diff
@@ -69,13 +69,29 @@ def registration_info(
     token=None,
     central_dns_suffix="azureiotcentral.com",
     device_status=None,
+    max_devices=None,
```
prbans, line 72: I'd suggest setting this to 0 or something

azext_iot/central/providers/device_provider.py
```diff
@@ -246,3 +262,28 @@ def get_all_registration_info(
...
+        for device in devices.values():
+            if not device.simulated:
+                registration_summary[device.device_status.name] = (
+                    registration_summary.get(device.device_status.name, 0) + 1
+                )
+        return registration_summary
+
+    def print_limited_devices(self, devices_to_display, device_collection):
+
+        if devices_to_display != 0:
```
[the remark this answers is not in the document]
prbans: the general convention in this code base is that `0` maps to `all` or `infinite`. So this logic may also need some updates

azext_iot/central/providers/device_provider.py
```diff
@@ -246,3 +262,28 @@ def get_all_registration_info(
...
+
+        devices = self.list_devices(central_dns_suffix=central_dns_suffix)
+        for device in devices.values():
+            if not device.simulated:
+                registration_summary[device.device_status.name] = (
+                    registration_summary.get(device.device_status.name, 0) + 1
+                )
+        return registration_summary
+
+    def print_limited_devices(self, devices_to_display, device_collection):
```
prbans, line 277: this also looks like a private method

prbans: Actually I don't think this method should be in the provider, it should be in the `commands_*` layer since its not really doing work, its just breaking up the results

azext_iot/tests/test_iot_central_int.py
```diff
@@ -190,7 +190,7 @@ def test_central_device_registration_info_filter_unassociated(self):
             device_info_results.append(device.get("device_info"))
 
         for device in device_info_results:
-            assert device.get("deviceStatus") == device_status_expected
+            assert device.get("device_status") == device_status_expected
```
prbans, line 193: I'd suggest adding coverage for restriction of what you should expect at the console. I.e. write a test such that a dev can't add unexpected properties to the payload without the test failing.

azext_iot/central/commands_device.py
```diff
@@ -69,13 +69,29 @@ def registration_info(
...
+    summarize_registration=False,
 ):
     provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token,)
     if not device_id:
```
prbans, line 76: looks like the code under `if not device` is now longer than the other case. Let's reverse the logic in that case:

```
if device:
  return provider.get_device_registration_info()

# do multi-device stuff here
```
```

File: azext_iot/central/commands_device.py [hunk 6]
```diff
@@ -69,13 +69,29 @@ def registration_info(
...
+            return summary
+
+        device_collection = provider.get_all_registration_info(
             central_dns_suffix=central_dns_suffix, device_status=device_status
         )
 
+        if max_devices:
+            provider.print_limited_devices(
+                device_collection=device_collection,
+                devices_to_display=int(max_devices),
```
prbans, line 90: we can dodge this conversion to `int` if we never let the value take on a non-integer value

azext_iot/central/commands_device.py
```diff
@@ -69,13 +69,29 @@ def registration_info(
...
+            )
+            return summary
+
+        device_collection = provider.get_all_registration_info(
             central_dns_suffix=central_dns_suffix, device_status=device_status
         )
 
+        if max_devices:
+            provider.print_limited_devices(
+                device_collection=device_collection,
```
prbans, line 89: nit - change `device_collection` to `devices` to be consistent with the rest of the code

azext_iot/central/commands_device.py
```diff
@@ -69,13 +69,29 @@ def registration_info(
     token=None,
     central_dns_suffix="azureiotcentral.com",
     device_status=None,
+    max_devices=None,
+    summarize_registration=False,
 ):
     provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token,)
     if not device_id:
-        return provider.get_all_registration_info(
+        if summarize_registration:
```
prbans, line 77: the fact that this code path is so specific leads me to think it deserves its own command

azext_iot/central/providers/device_provider.py
```diff
@@ -246,3 +262,28 @@ def get_all_registration_info(
         ]
 
         return result
+
+    def get_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
+        registration_summary = {}
+
+        devices = self.list_devices(central_dns_suffix=central_dns_suffix)
```
prbans, line 269: lets move this to an object. Create something like

```
class RegistrationSummary():
  def __init__(self):
    self.counter = {name: 0 for name in DeviceStatus}
  
  def process_device(self, device: Device):
    self.counter[device.device_status.name] += 1
```

Then when calling, this code can be simplified to:
```
summary = RegistrationSummary()

devices = ...
[summary.process_device(device) for device in devices if not devices.simulated]
return summary.counter
```

prbans: I'd highly suggest familiarizing yourself with list and dictionary comprehensions. They're really really important in python, and are faster than for-loops in general. More than happy to help if you have any questions about it

azext_iot/central/providers/device_provider.py
```diff
@@ -231,7 +247,7 @@ def get_all_registration_info(
             filtered_devices = [
                 device
                 for device in real_devices
-                if device.device_status == device_status
+                if device.device_status == DeviceStatus(device_status)
             ]
 
         if len(devices) != len(filtered_devices):
```
prbans, line 253: we need to update the logger warning on line 239

azext_iot/central/providers/device_provider.py
prbans, line 269: FYI - this feature is currently going to be broken for apps with more than 25 devices since we don't yet support device paging. I think it might be a good exercise for you to try implement device paging as part of this PR.

azext_iot/central/params.py
```diff
@@ -86,6 +86,17 @@ def load_central_arguments(self, _):
...
         )
+        context.argument(
+            "summarize_registration",
+            options_list=["--registrationsummary", "--rs"],
+            arg_type=get_three_state_flag(),
+            help="Add this flag to get a summary of registration information",
+        )
+        context.argument(
+            "max_devices",
+            options_list=["--maxdevices", "--md"],
```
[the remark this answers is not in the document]
prbans: lets also add `type=int` here so we can force users to only enter numbers

azext_iot/central/params.py
```diff
@@ -86,6 +86,17 @@ def load_central_arguments(self, _):
             arg_type=get_enum_type(DeviceStatus),
             help="Indicates filter option for device status",
         )
+        context.argument(
+            "summarize_registration",
+            options_list=["--registrationsummary", "--rs"],
```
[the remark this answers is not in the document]
prbans: infact - looking at the rest of the PR, i think that the summary should be its own command, I think we're overloading registration-info a bit too much

PR #189, opened 2020-05-20
Adds authentication-type parameters to device-identity import and export commands

c-ryan-k:
---

### General Guidelines


azext_iot/_params.py
```diff
@@ -384,21 +386,37 @@ def load_arguments(self, _):
             help="If set, keys are exported normally. Otherwise, keys are "
             "set to null in export output.",
         )
+        context.argument(
+            "storage_authentication_type",
+            options_list=["--auth-type", "--storage-authentication-type"],
+            arg_type=get_enum_type(AuthenticationType),
+            help="Authentication type for communicating with the storage container.  "
+            "Allowed values: identity, key",
```
digimaun, line 394: I would double check if you need to provide the text of allowed values for enum types. The arg parse framework should take care of it.

c-ryan-k: This actually may have been left behind before I added the `arg_type` and removed the `Based` suffix... You're right, this is likely unnecessary

azext_iot/common/utility.py
```diff
@@ -462,3 +462,8 @@ def is_iso8601_time(self, to_validate: str) -> bool:
             return bool(isodate.parse_time(to_validate))
         except Exception:
             return False
+
+
+def ensure_min_version(cur_ver, min_ver):
+    from packaging import version
```
digimaun, line 468: Is the `packaging` package ;) pulled in with setuptools or is this a Core CLI dependency import?

azext_iot/operations/hub.py
```diff
@@ -2041,14 +2042,45 @@ def _handle_c2d_msg(target, device_id, receive_settle):
...
+    storage_authentication_type=None,
+    resource_group_name=None,
 ):
     from azext_iot._factory import iot_hub_service_factory
+    from azure.mgmt.iothub import __version__ as iot_sdk_version
 
     client = iot_hub_service_factory(cmd.cli_ctx)
     target = get_iot_hub_connection_string(client, hub_name, resource_group_name)
+
+    if ensure_min_version(iot_sdk_version, "0.12.0"):
```
digimaun, line 2058: Yea this looks cleaner than the introspection path

azext_iot/common/utility.py
c-ryan-k: Ah, I see - it looks like this `packaging` was a dependency of pytest. I'll get the setuptools version imported, I believe it's 
```
from pkg_resources import packaging
packaging.version.parse(...)
```

c-ryan-k:
image
FYI:
Tests are passing...few transient issues but pretty much green.

azext_iot/operations/hub.py
```diff
@@ -2057,17 +2089,43 @@ def iot_device_import(
...
+            input_blob_container_uri=input_blob_container_uri,
+            output_blob_container_uri=output_blob_container_uri,
+            input_blob_name=None,
+            output_blob_name=None,
+            authentication_type=storage_authentication_type,
+        )
+        return client.import_devices(
+            target["resourcegroup"], hub_name, import_devices_parameters=import_request,
+        )
+    if storage_authentication_type is not None:
```
digimaun, line 2120: nit: this could be `if storage_authentication_type:`

azext_iot/operations/hub.py
```diff
@@ -2041,14 +2042,45 @@ def _handle_c2d_msg(target, device_id, receive_settle):
...
+        from azext_iot.common.shared import AuthenticationType
+
+        storage_authentication_type = (
+            AuthenticationType(storage_authentication_type).name
+            if storage_authentication_type
+            else None
+        )
+        export_request = ExportDevicesRequest(
+            export_blob_container_uri=blob_container_uri,
+            exclude_keys=not include_keys,
```
prbans, line 2069: nit - this `not include_keys` is a little wierd, and I see that both usages of it in this function are using `exclude_keys`. Would it be worth while changing the function definition to be `exclude_keys` instead?

c-ryan-k:
image
third time's the charm apparently

azext_iot/operations/hub.py
c-ryan-k: Yeah, I agree it's a bit wonky. This command interface was pre-extension (way back when) and it's likely been kept the way it is to avoid a breaking change.

We can eventually deprecate `include_keys` as a parameter in place of `exclude_keys` to match the SDK model and parameters, like you mentioned, but I'd rather address that as a separate PR / issue.

PR #192, opened 2020-05-22
Add support for registration  summary

valluriraj:
---

### General Guidelines

### DIFF
--- a/azext_iot/central/_help.py
+++ b/azext_iot/central/_help.py
@@ -129,7 +129,7 @@ def _load_central_devices_help():
             short-summary: Provides a registration summary of all the devices in an app.
             long-summary: |
                 Note: This command can take a significant amount of time to return
-                if no device id is specified and your app contains a lot of devices
+                if your app contains a lot of devices
             examples:
             - name: Registration summary
               text: >
--- a/azext_iot/central/_help.py
+++ b/azext_iot/central/_help.py
@@ -111,16 +111,36 @@ def _load_central_devices_help():
         "iot central app device registration-info"
     ] = """
         type: command
-        short-summary: Get registration info for a given device in IoT Central
+        short-summary: Get registration info on device(s) from IoT Central
         long-summary: |
             Note: This command can take a significant amount of time to return
             if no device id is specified and your app contains a lot of devices
 
         examples:
+        - name: Get registration info on all devices. This command may take a long time to complete execution.
+          text: >
+            az iot central app device registration-info
+            --app-id {appid}
+
         - name: Get registration info on specified device
           text: >
             az iot central app device registration-info
-            --app-id {appid} --device-id {deviceid}
+            --app-id {appid}
+            --device-id {deviceid}
+    """
+
+
+helps[
+    "iot central app device registration-summary"
+] = """
+        type: command
+        short-summary: Provides a registration summary of all the devices in an app. 
+
+        examples:
+        - name: Registration summary
+          text: >
+            az iot central app device registration-summary
+            --app-id {appid}            
     """
 
 
--- a/azext_iot/central/_help.py
+++ b/azext_iot/central/_help.py
@@ -117,31 +117,24 @@ def _load_central_devices_help():
             if no device id is specified and your app contains a lot of devices
 
         examples:
-        - name: Get registration info on all devices. This command may take a long time to complete execution.
-          text: >
-            az iot central app device registration-info
-            --app-id {appid}
-
         - name: Get registration info on specified device
           text: >
-            az iot central app device registration-info
-            --app-id {appid}
-            --device-id {deviceid}
+            az iot central app device registration-info --app-id {appid} --device-id {deviceid}
     """
 
-
-helps[
-    "iot central app device registration-summary"
-] = """
-        type: command
-        short-summary: Provides a registration summary of all the devices in an app. 
-
-        examples:
-        - name: Registration summary
-          text: >
-            az iot central app device registration-summary
-            --app-id {appid}            
-    """
+    helps[
+        "iot central app device registration-summary"
+    ] = """
+            type: command
+            short-summary: Provides a registration summary of all the devices in an app.
+            long-summary: |
+                Note: This command can take a significant amount of time to return
+                if no device id is specified and your app contains a lot of devices
+            examples:
+            - name: Registration summary
+              text: >
+                az iot central app device registration-summary --app-id {appid}
+        """
 
 
 def _load_central_device_templates_help():
--- a/azext_iot/central/providers/device_provider.py
+++ b/azext_iot/central/providers/device_provider.py
@@ -207,6 +207,14 @@ class CentralDeviceProvider:
 
         return info
 
+    def get_device_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
+        return central_services.device.get_device_registration_summary(
+            cmd=self._cmd,
+            app_id=self._app_id,
+            token=self._token,
+            central_dns_suffix=central_dns_suffix,
+        )
+        
     def _dps_populate_essential_info(self, dps_info, device_status: DeviceStatus):
         error = {
             DeviceStatus.provisioned: "None.",
@@ -222,10 +230,4 @@ class CentralDeviceProvider:
         }
         return filtered_dps_info
 
-    def get_device_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
-        return central_services.device.get_device_registration_summary(
-            cmd=self._cmd,
-            app_id=self._app_id,
-            token=self._token,
-            central_dns_suffix=central_dns_suffix,
-        )
+
--- a/azext_iot/central/providers/device_provider.py
+++ b/azext_iot/central/providers/device_provider.py
@@ -221,3 +221,11 @@ class CentralDeviceProvider:
             "error": error.get(device_status),
         }
         return filtered_dps_info
+
+    def get_device_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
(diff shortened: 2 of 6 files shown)


azext_iot/central/providers/device_provider.py
```diff
@@ -221,3 +221,11 @@ def _dps_populate_essential_info(self, dps_info, device_status: DeviceStatus):
             "error": error.get(device_status),
         }
         return filtered_dps_info
+
+    def get_device_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
```
prbans, line 225: pass the token down too

prbans, line 225: >get_device_registration_summary [](start = 8, length = 31)

minor nit: publics generally go before privates

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
...
+        DeviceStatus.provisioned.value: 0,
+        DeviceStatus.registered.value: 0,
+        DeviceStatus.unassociated.value: 0,
+        DeviceStatus.blocked.value: 0,
+    }
+
+    url = "https://{}.{}/{}".format(app_id, central_dns_suffix, BASE_PATH)
+    headers = _utility.get_headers(token, cmd)
+
+    while url is not None:
```
prbans, line 115: `while url`

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
...
+        response = requests.get(url, headers=headers)
+        result = _utility.try_extract_result(response)
+
+        if "value" not in result:
+            raise CLIError("Value is not present in body: {}".format(result))
+
+        for device in result["value"]:
```
prbans, line 122: >device [](start = 12, length = 6)

yeah I think this is the correct way to do it. If we really really wanted to we could do something like:

[registration_summary[Device(device).device_status.value] += 1 for device in result["value"]]

but I personally think that's a bit too chaotic to read.

List comprehensions are generally faster than creating a list, but not always the best answer. I appreciate you raising the concern with me offline though

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
     return devices
 
 
+def get_device_registration_summary(
```
prbans, line 88: >get_device_registration_summary [](start = 4, length = 31)

A lot of the code in this function is copy-pasted from `list_devices`, can we functionalize the common code and reuse it? That way if we discover a bug with one function, we can fix it everywhere at the same time

azext_iot/central/providers/device_provider.py
prbans, line 225: >get_device_registration_summary [](start = 8, length = 31)

as we discussed offline maybe drop a logger.warn("This command may take a long time to complete if your app contains a lot of devices") or similar

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
...
+
+    url = "https://{}.{}/{}".format(app_id, central_dns_suffix, BASE_PATH)
+    headers = _utility.get_headers(token, cmd)
+
+    while url is not None:
+        response = requests.get(url, headers=headers)
+        result = _utility.try_extract_result(response)
+
+        if "value" not in result:
+            raise CLIError("Value is not present in body: {}".format(result))
```
prbans, line 120: So I know this is from my earlier implementation, but do we want to throw away the whole result if there's an issue?

Or do we want to surface a warning and return the data we did manage to process?

I'd also raise this concern for using the `_utility.try_extract_result(response)` here

prbans: We can also file a followup item on this

---
In reply to: [428981926](https://github.com/Azure/azure-iot-cli-extension/pull/192#discussion_r428981926) [](ancestors = 428981926)

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
...
+        DeviceStatus.registered.value: 0,
+        DeviceStatus.unassociated.value: 0,
+        DeviceStatus.blocked.value: 0,
+    }
+
+    url = "https://{}.{}/{}".format(app_id, central_dns_suffix, BASE_PATH)
+    headers = _utility.get_headers(token, cmd)
+
+    while url is not None:
+        response = requests.get(url, headers=headers)
```
prbans, line 116: reminder: as we discussed offline maybe worth a `logger.info("Processed {} devices...".format(sum(registration.summary.keys()))`

azext_iot/central/_help.py
```diff
@@ -111,16 +111,36 @@ def _load_central_devices_help():
...
     ] = """
         type: command
-        short-summary: Get registration info for a given device in IoT Central
+        short-summary: Get registration info on device(s) from IoT Central
         long-summary: |
             Note: This command can take a significant amount of time to return
             if no device id is specified and your app contains a lot of devices
 
         examples:
+        - name: Get registration info on all devices. This command may take a long time to complete execution.
```
prbans, line 120: did you plan to add this functionality back in?

azext_iot/central/_help.py
```diff
@@ -111,16 +111,36 @@ def _load_central_devices_help():
...
         - name: Get registration info on specified device
           text: >
             az iot central app device registration-info
-            --app-id {appid} --device-id {deviceid}
+            --app-id {appid}
+            --device-id {deviceid}
+    """
+
+
+helps[
```
prbans, line 133: careful - this helps is outside of the above function. Python uses tabs to mark start and end of a function. Please add this help text back into the above function

azext_iot/central/services/device.py
```diff
@@ -84,6 +85,47 @@ def list_devices(
...
+        app_id: name of app (used for forming request URL)
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        registration summary
+    """
+
+    registration_summary = {
```
prbans, line 105: for safety, I'd recommend moving this over to a dictionary comprehension. Reason being - if we add a new enum value tomorrow we won't have to update this code due to pythonic magic.

`registration_summary = {status.value: 0 for status in DeviceStatus}`

azext_iot/tests/test_iot_central_int.py
```diff
@@ -229,6 +229,19 @@ def test_central_device_registration_info_unassociated(self):
...
+        result = self.cmd(
+            "iot central app device registration-summary --app-id {}".format(APP_ID,)
+        )
+
+        json_result = result.get_output_in_json()
+        assert json_result[DeviceStatus.provisioned.value] is not None
+        assert json_result[DeviceStatus.registered.value] is not None
+        assert json_result[DeviceStatus.unassociated.value] is not None
+        assert json_result[DeviceStatus.blocked.value] is not None
+        assert len(json_result) == 4
```
prbans, line 243: Nice!!

If we want to be flexible to DeviceStatus updates we can do this:

```
assert len(json_result) == len(DeviceStatus)
for val in json_result.values():
    assert val >= 0
```

This change means that if we add or remove fields from DeviceStatus that this test will still work.

If we want to enforce that only these 4 props should ever be present then keep the test as is

azext_iot/central/services/device.py
prbans: So the reason this works is because an Enum in python is iterable. Its pretty nifty

---
In reply to: [428986126](https://github.com/Azure/azure-iot-cli-extension/pull/192#discussion_r428986126) [](ancestors = 428986126)

azext_iot/tests/test_iot_central_int.py
prbans: It will give us the additional benefit of checking that the values in the dictionary are `int`s

---
In reply to: [428987160](https://github.com/Azure/azure-iot-cli-extension/pull/192#discussion_r428987160) [](ancestors = 428987160)

valluriraj: #wontfix - I would prefer to keep it this way so that the test fails for any change.

azext_iot/central/services/device.py
valluriraj: #wontfix  - partial data will not get consistent result and is not useful.

azext_iot/central/_help.py
```diff
@@ -111,18 +111,31 @@ def _load_central_devices_help():
...
     """
 
+    helps[
+        "iot central app device registration-summary"
+    ] = """
+            type: command
+            short-summary: Provides a registration summary of all the devices in an app.
+            long-summary: |
+                Note: This command can take a significant amount of time to return
+                if no device id is specified and your app contains a lot of devices
```
prbans, line 132: device-id can't be specified in registration-summary

valluriraj:
========================================== 894 passed, 2 skipped in 2619.60 seconds ===========================================

valluriraj:
image

azext_iot/central/providers/device_provider.py
```diff
@@ -207,6 +207,14 @@ def get_device_registration_info(
 
         return info
 
+    def get_device_registration_summary(self, central_dns_suffix="azureiotcentral.com"):
+        return central_services.device.get_device_registration_summary(
```
prbans, line 211: did we want to add the logger.warning in

azext_iot/central/services/device.py
prbans: spoke offline, #wontfix

PR #193, opened 2020-05-26
Add module filtering

valluriraj:
---

### General Guidelines

### DIFF
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -161,9 +161,13 @@ def load_deprecated_iotcentral_params(self, _):
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
-        context.argument(
-            "module_id", options_list=["--module-id", "-m"], help="Iot Edge Module ID",
-        )
+
+        with self.argument_context("iotcentral app") as context:
+            context.argument(
+                "module_id",
+                options_list=["--module-id", "-m"],
+                help="Iot Edge Module ID",
+            )
 
     with self.argument_context("iot central device-twin") as context:
         context.argument("app_id", options_list=["--app-id"], help="Target App.")
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -161,6 +161,10 @@ def load_deprecated_iotcentral_params(self, _):
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
+        context.argument(
+            "module_id", options_list=["--module-id", "-m"], help="Iot Edge Module ID",
+        )
+
 
     with self.argument_context("iot central device-twin") as context:
         context.argument("app_id", options_list=["--app-id"], help="Target App.")
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -165,7 +165,6 @@ def load_deprecated_iotcentral_params(self, _):
             "module_id", options_list=["--module-id", "-m"], help="Iot Edge Module ID",
         )
 
-
     with self.argument_context("iot central device-twin") as context:
         context.argument("app_id", options_list=["--app-id"], help="Target App.")
         context.argument(
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -82,10 +82,16 @@ def load_central_arguments(self, _):
     with self.argument_context("iot central app monitor-events") as context:
         context.argument("timeout", arg_type=event_timeout_type)
         context.argument("properties", arg_type=event_msg_prop_type)
+        context.argument(
+            "module_id", options_list=["--module-id", "-m"], help="Module Id"
+        )
 
     with self.argument_context("iot central app validate-messages") as context:
         context.argument("timeout", arg_type=event_timeout_type)
         context.argument("properties", arg_type=event_msg_prop_type)
+        context.argument(
+            "module_id", options_list=["--module-id", "-m"], help="Module Id"
+        )
         context.argument("style", arg_type=style_type)
         context.argument(
             "duration",
--- a/azext_iot/central/params.py
+++ b/azext_iot/central/params.py
@@ -78,20 +78,17 @@ def load_central_arguments(self, _):
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
+        context.argument(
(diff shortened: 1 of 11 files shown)


azext_iot/central/params.py
```diff
@@ -82,10 +82,16 @@ def load_central_arguments(self, _):
     with self.argument_context("iot central app monitor-events") as context:
         context.argument("timeout", arg_type=event_timeout_type)
         context.argument("properties", arg_type=event_msg_prop_type)
+        context.argument(
```
prbans, line 85: I'd suggest removing `module_id` from the `validate-messages` and `monitor-events` parents, and adding it to the `iot central app` parent instead

azext_iot/monitor/handlers/common_handler.py
```diff
@@ -70,3 +73,24 @@ def _should_process_interface(self, interface_name):
...
+                regex = (
+                    re.escape(expected_module_id)
+                    .replace("\\*", ".*")
+                    .replace("\\?", ".")
+                    + "$"
+                )
+                if not re.match(regex, module_id):
+                    return False
+            else:
+                return False
```
prbans, line 95: Lets pull 85 - 96 into its own function because it looks copy-pasted from `should_process_device`

azext_iot/monitor/parsers/common_parser.py
```diff
@@ -84,6 +86,14 @@ def _parse_device_id(self, message: Message) -> str:
             self._add_issue(severity=Severity.error, details=details)
             return ""
 
+    def _parse_module_id(self, message: Message) -> str:
+        try:
+            return str(message.annotations.get(MODULE_ID_IDENTIFIER), "utf8")
+        except Exception:
+            details = strings.unknown_module_id()
```
prbans, line 93: I'd suggest not logging the module name as an error, because non-edge devices should NOT be sending a module name.

I'd suggest just "passing" here, similar to how parsing interface is currently done

azext_iot/monitor/handlers/common_handler.py
prbans: infact, we could extract everything below 81 out into a function

azext_iot/monitor/parsers/common_parser.py
```diff
@@ -27,6 +28,7 @@ def __init__(self, message: Message, common_parser_args: CommonParserArguments):
         self._message = message
         self.device_id = self._parse_device_id(message)
         self.interface_name = self._parse_interface_name(message)
+        self.module_id = self._parse_module_id(message)
```
prbans, line 31: Would it be possible to add a quick UT for parsing `model_id`?

valluriraj: #wontfix

prbans: This is something I'd like to insist on, any parsing issues related to `model_id` will crash our event monitor, and the crash will likely be silent due to the way we are suppressing output

azext_iot/monitor/handlers/common_handler.py
```diff
@@ -43,22 +46,25 @@ def _should_process_device(self, device_id):
...
-            if "*" in expected_device_id or "?" in expected_device_id:
+        process_device = self.perform_id_match(expected_device_id, device_id)
+
+        if expected_devices:
+            if device_id not in expected_devices:
+                return False
+
+        return process_device
+
+    def perform_id_match(self, expected_id, actual_id):
```
prbans, line 57: nit - add leading `_` so that we know its private

prbans, line 57: Awesome!!! thanks for doing this, makes maintenance a lot easier

azext_iot/monitor/handlers/common_handler.py
```diff
@@ -43,22 +46,25 @@ def _should_process_device(self, device_id):
         expected_device_id = self._common_handler_args.device_id
         expected_devices = self._common_handler_args.devices
 
-        if expected_device_id and expected_device_id != device_id:
-            if "*" in expected_device_id or "?" in expected_device_id:
+        process_device = self.perform_id_match(expected_device_id, device_id)
+
+        if expected_devices:
```
prbans, line 51: was there a reason we changed this from 
```
if expected_device_id and expected_device_id != device_id:
  return False
```

azext_iot/tests/test_monitor_parsers_unit.py
```diff
@@ -100,6 +114,7 @@ def test_parse_message_should_succeed(
         payload,
         properties,
         app_properties,
+        module_id,
     ):
         # setup
         properties = MessageProperties(
```
prbans, line 120: lets also add an assert on the module name being what we passed in

azext_iot/central/params.py
```diff
@@ -78,6 +78,9 @@ def load_central_arguments(self, _):
             help="Central dns suffix. "
             "This enables running cli commands against non public/prod environments",
         )
+        context.argument(
+            "module_id", options_list=["--module-id", "-m"], help="Module Id"
```
prbans, line 82: nit - for the help maybe include some more info like `IoT Edge Module ID`

azext_iot/tests/test_monitor_parsers_unit.py
prbans: Should be able to add a similar snippet to whats on 143:
image

```
if module_id:
    assert parsed_msg["event"]["annotations"][module_identifier] == module_id
else:
    assert not parsed_msg["event"]["annotations"].get(module_identifier)
```

azext_iot/monitor/handlers/common_handler.py
valluriraj: readability - changed it back.

azext_iot/central/params.py
```diff
@@ -159,6 +162,13 @@ def load_deprecated_iotcentral_params(self, _):
             "This enables running cli commands against non public/prod environments",
         )
 
+        with self.argument_context("iotcentral app monitor-events") as context:
```
prbans, line 165: just add module-id to the `iotcentral` subcommand group as a whole (unless thats still causing issues). See how `timeout` is done

prbans: i.e. unparent this from `iotcentral app monitor-events`

PR #197, opened 2020-06-08
Central - run command-response on device

prbans:
```
az iot central app device run-command `
-n <APP-NAME> `
-d simulated-c2d `
-i c2dTestingTemplate_356 `
--cn sync_command `
-k "{'request':{'Object1':{'Object1_Double':1,'Object1_String':'awd'},'Object2':{'Object2_Vector':{'x':1,'y':3,'z':-3},'Object2_Date':'2020-07-02'}}}"
```
---

### General Guidelines


prbans:
image

prbans:
image

prbans:
image

prbans:
image

azext_iot/central/_help.py
```diff
@@ -136,6 +136,51 @@ def _load_central_devices_help():
                 az iot central app device registration-summary --app-id {appid}
         """
 
+    helps[
+        "iot central app device run-command"
```
digimaun, line 140: Is "run command" IoT Central nomenclature? On the IoT hub side its usually invoke or execute command.

azext_iot/central/_help.py
```diff
@@ -136,6 +136,51 @@ def _load_central_devices_help():
...
+    ] = """
+            type: command
+            short-summary: Run a command on a device and view associated response. Does NOT monitor property updates that the command may perform.
```
digimaun, line 143: Isn't the second sentence assumed? This isn't an active monitoring command.

azext_iot/central/_help.py
```diff
@@ -136,6 +136,51 @@ def _load_central_devices_help():
...
+            long-summary: |
+                Note: payload should be nested under "request".
+                i.e. if your device expects the payload in a shape {"key": "value"}
+                payload should be {"request": {"key": "value"}}.
+                --content can be pointed at a filepath too (.../path/to/payload.json)
+            examples:
+            - name: Run command response
+              text: >
+                az iot central app device run-command
```
digimaun, line 152: Are there no connectivity or response timeout arguments to deal with on the central side?

azext_iot/central/_help.py
```diff
@@ -136,6 +136,51 @@ def _load_central_devices_help():
...
 
+    helps[
+        "iot central app device run-command"
+    ] = """
+            type: command
+            short-summary: Run a command on a device and view associated response. Does NOT monitor property updates that the command may perform.
+            long-summary: |
+                Note: payload should be nested under "request".
+                i.e. if your device expects the payload in a shape {"key": "value"}
+                payload should be {"request": {"key": "value"}}.
```
digimaun, line 147: nit: consider adding flexibility to support either style. There is precedence...for example in [edge deployments](https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/_help.py#L784)

azext_iot/central/_help.py
```diff
@@ -136,6 +136,51 @@ def _load_central_devices_help():
...
+                -i {interfaceid}
+                --cn {commandname}
+                -k {payload}
+        """
+
+    helps[
+        "iot central app device show-command-history"
+    ] = """
+            type: command
+            short-summary: Get most recent command-response request and response payload.
```
digimaun, line 173: There is currently an API limitation which is limiting the result of this operation to one item (the most recent) right?

prbans: Correct

azext_iot/central/services/device.py
```diff
@@ -220,3 +220,79 @@ def get_device_credentials(
...
+        command_name: name of command to execute
+        payload: params for command
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        result (currently a 201)
+    """
+    url = "https://{}.{}/{}/{}/components/{}/commands/{}".format(
```
digimaun, line 252: Note to discuss offline.

azext_iot/central/_help.py
digimaun: I foresee this needing a top/last X # of command history

azext_iot/tests/test_iot_central_int.py
```diff
@@ -190,6 +196,45 @@ def test_central_device_registration_info_registered(self):
         assert device_registration_info.get("status") is None
         assert dps_state.get("error") == "Device is not yet provisioned."
 
+    def test_central_run_command(self):
```
digimaun, line 199: I like how this test has come together combined with the various helpers.

azext_iot/central/_help.py
prbans: So I'd prefer not to do sniffing if possible.

2 main reasons that we can drill into below.
1) If a developer plans to use CLI as testing for how API's work, I want CLI to behave as closely to API as possible
2) No restrictions on param naming, so if a customer sets `request` as a param to a function, sniffing will lead to anomalous behavior

Drill down: 
1) A customer might just be playing around with CLI to expose concepts/test payloads. If they give the CLI a payload that doesn't work with the API, then they will be confused why the "api isn't working" when we have done sniffing for them.

2) The sniffing can also introduce ambiguity, which I'm not a personal fan of.

For example - there is no restriction for parameter names. So a user could have "request" as a perfectly valid name, which leads to "odd" behavior.

I.e.

```
{
  request: {
    request: ....
    a: ....
    ....
  }
}
```

But if the customer does something like:

```
{
  request: ....
  a: ....
  ....
}
```

Does request get passed to the device? are the other props ignored? etc

azext_iot/central/_help.py
prbans: Doesn't look like it as of now, its not a well documented concept. Known issue that the team is working on

https://docs.microsoft.com/en-us/rest/api/iotcentral/devices/executecomponentcommand

image
image

azext_iot/central/_help.py
prbans: We have a concept of async command that we should chat about offline

azext_iot/central/_help.py
prbans: I'll check with my team about "invoke". My team was trying to avoid "execute"

azext_iot/central/_help.py
prbans: #wontfix

azext_iot/central/_help.py
prbans: I think its a matter of exposing it via public API, I'm seeing we have this functionality in Web UI at the moment. Might be a v-NEXT thing

image

prbans: for now #wontfix

azext_iot/central/_help.py
digimaun: Fair points particularly #2 resonates. There are more restrictions/structured format on edge payloads which doesn't have this ambiguity. For vanilla IoT Hub only the command payload is expected i.e. not {payload: {a: ..., b: ...}} but {a: ..., b: ...}.

I would caution the second part of #1 as a primary design goal as the CLI's purpose is not to mirror service API's...its closer to enabling users and scenarios to achieve Azure IoT goals through automation as well as supporting inner dev loop tooling. DevEx and enablement are among core goals. Users can always use debug facilities along with nicely layed out logging to understand more of how a command works.

prbans: Ok that makes sense, yeah I'll keep your feedback on #1 in mind. If we get customer feedback that they want sniffing then I'll add more significant sniffing than just `request` sniffing. I'll likely do some `component_name` sniffing etc too

prbans: for now #wontfix

azext_iot/central/_help.py
prbans: So I checked with the team, they want to keep Run because "run" is what our UI says

image

prbans: #wontfix for now

PR #203, opened 2020-06-19
Add new allocation-policy to enrollment and enrollment-group

anusapan:
---

### General Guidelines


azext_iot/operations/dps.py
```diff
@@ -621,6 +658,11 @@ def _validate_allocation_policy_for_enrollment(allocation_policy,
                 raise CLIError('Please provide a hub to be assigned with device.')
             if iot_hub_list and len(iot_hub_list) > 1:
                 raise CLIError('Only one hub is required in static allocation policy.')
+        if allocation_policy == AllocationType.custom.value:
+            if webhook_url is None:
```
digimaun, line 662: I think we can check if either webhook_url or api_version is None or empty, then we provide the CLI error message which includes both arguments that must be included...rather than one at a time.

digimaun: Message could be something like "Please provide both the Azure function webhook url and provisioning service api-version when the allocation-policy is defined as Custom."

azext_iot/_params.py
```diff
@@ -762,6 +762,16 @@ def load_arguments(self, _):
...
         )
+        context.argument(
+            "webhook_url",
+            options_list=["--webhook-url", "--wh"],
+            help="The webhook URL used for custom allocation requests.",
+        )
+        context.argument(
+            "api_version",
+            options_list=["--api-version", "--av"],
+            help="The API version of the provisioning service in the custom allocation request.",
```
digimaun, line 773: I think we should add a tiny bit more detail for this argument. Specifically mentioning 'types' and the minimum supported version.

Something like:

"The API version of the provisioning service **types** sent in the custom allocation request. Minimum supported version: 2018-09-01-preview"

azext_iot/tests/test_iot_dps_int.py
```diff
@@ -163,12 +163,15 @@ def test_dps_enrollment_x509_lifecycle(self):
...
         enrollment_id = self.create_random_name('enrollment-for-test', length=48)
+        enrollment_id2 = self.create_random_name('enrollment-for-test', length=48)
         attestation_type = AttestationType.symmetricKey.value
         primary_key = 'x3XNu1HeSw93rmtDXduRUZjhqdGbcqR/zloWYiyPUzw='
         secondary_key = 'PahMnOSBblv9CRn5B765iK35jTvnjDUjYP9hKBZa4Ug='
         device_id = self.create_random_name('device-id-for-test', length=48)
         reprovisionPolicy_reprovisionandresetdata = 'reprovisionandresetdata'
         hub_host_name = '{}.azure-devices.net'.format(hub)
+        webhook_url = 'https://www.test.test'
+        api_version = '2018-09-01-preview'
```
digimaun, line 174: Maybe we test with the same version of the current embedded DPS SDK? The version is `2019-03-31`

azext_iot/tests/test_iot_dps_int.py
```diff
@@ -163,12 +163,15 @@ def test_dps_enrollment_x509_lifecycle(self):
...
     def test_dps_enrollment_symmetrickey_lifecycle(self):
         enrollment_id = self.create_random_name('enrollment-for-test', length=48)
+        enrollment_id2 = self.create_random_name('enrollment-for-test', length=48)
         attestation_type = AttestationType.symmetricKey.value
         primary_key = 'x3XNu1HeSw93rmtDXduRUZjhqdGbcqR/zloWYiyPUzw='
         secondary_key = 'PahMnOSBblv9CRn5B765iK35jTvnjDUjYP9hKBZa4Ug='
         device_id = self.create_random_name('device-id-for-test', length=48)
         reprovisionPolicy_reprovisionandresetdata = 'reprovisionandresetdata'
         hub_host_name = '{}.azure-devices.net'.format(hub)
+        webhook_url = 'https://www.test.test'
```
digimaun, line 173: Can we use GUID formed URL?

azext_iot/operations/dps.py
anusapan: Updated

azext_iot/tests/test_iot_dps_int.py
anusapan: Updated

azext_iot/tests/test_iot_dps_int.py
anusapan: Updated

azext_iot/_params.py
anusapan: Updated

azext_iot/tests/test_iot_dps_int.py
anusapan: This is the test data..

digimaun: Totally understand. My concern is, if during tests, any packets get sent to what is not a real URL for now...but might be a real URL later (concern is if the DPS service sends any ack or other info to a fake webhook url).

If DPS doesn't do anything like that on setting the custom allocation policy then its a moot point.

digimaun: Ok, looks like .test domain can't be registed https://en.wikipedia.org/wiki/.test

anusapan:
image

PR #210, opened 2020-06-25
Added ability to monitor property updates for devices connected to IoT Central

valluriraj:
---

### General Guidelines


prbans:
image

azext_iot/central/_help.py
```diff
@@ -316,6 +316,20 @@ def _load_central_monitors_help():
             az iot central app validate-messages --app-id {app_id} -d {device_id} --cg {consumer_group_name}
     """
 
+    helps[
+        "iot central app monitor-properties"
+    ] = """
+        type: command
+        short-summary: Monitor desired and reported  sent to/from the IoT Hub for an IoT Central app.
```
digimaun, line 323: nit: double space between "reported" and "sent"
related nit: Monitor desired and reported what? Looks like a word (properties?) is missing.

azext_iot/central/_help.py
```diff
@@ -316,6 +316,20 @@ def _load_central_monitors_help():
...
+        long-summary: |
+                    EXPERIMENTAL requires Python 3.5+
```
digimaun, line 325: Is this command really experimental? The use of experimental in the monitoring commands in the IoT extension has historically been more due to the reliance of a C extension for amqp and managing it out-of-band compared to typical python package dependencies. Maybe associating this as a preview command is good enough?

azext_iot/monitor/property.py
```diff
@@ -0,0 +1,96 @@
...
+        seconds=DEVICETWIN_MONITOR_TIME_SEC
+    )
+
+    last_updated = isodate.parse_datetime(val["$lastUpdated"])
+    return last_updated.timestamp() >= updated_within.timestamp()
+
+
+def changed_props(prop, metadata, property_name):
+    # not an interface - whole thing is change log
+    if "$iotin" not in property_name:
```
digimaun, line 90: nit: `$iotin` could be extracted and put as a constant variable somewhere.

prbans: #wontfix this iteration

Will fix next iteration

Raj is working on a follow up PR where he needs more complex logic to detect if something is an interface/property

azext_iot/monitor/property.py
```diff
@@ -0,0 +1,96 @@
...
+        return prop
+    # iterate over property in the interface
+    # if the property is not an exact match for what is present in the previous set of properties
```
digimaun, line 93: Is this comment accurate? The comparison does not look like its by value, instead by the properties last updated time with respect to the last time the twin was pulled.

prbans: Good catch - logic change was made to address a bug and the comment lagged. Will update

azext_iot/central/_help.py
```diff
@@ -316,6 +316,20 @@ def _load_central_monitors_help():
             az iot central app validate-messages --app-id {app_id} -d {device_id} --cg {consumer_group_name}
     """
 
+    helps[
+        "iot central app monitor-properties"
+    ] = """
+        type: command
+        short-summary: Monitor desired and reported  sent to/from the IoT Hub for an IoT Central app.
+        long-summary: |
```
digimaun, line 324: For long description, it would be helpful for a brief/high-level summary of how the property monitoring works - users know what to expect from reading the description.

azext_iot/monitor/property.py
```diff
@@ -0,0 +1,96 @@
...
+def start_property_monitor(
+    cmd,
+    device_id,
+    app_id,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    polling_interval_seconds=DEVICETWIN_POLLING_INTERVAL_SEC,
+):
+    prev_twin = None
+
+    device_twin_provider = CentralDeviceTwinProvider(
```
digimaun, line 29: While not blocking this PR, this functionality can and should be generalized to work for more than Central.

azext_iot/monitor/property.py
```diff
@@ -0,0 +1,96 @@
...
+        if prev_twin:
+            change_d = compare_properties(
+                prev_twin.desired_property, twin.desired_property
+            )
+            change_r = compare_properties(
+                prev_twin.reported_property, twin.reported_property
+            )
+
+            if change_d:
+                print("Changes in desired properties:")
```
digimaun, line 49: Is the usage of this functionality specifically for the inner dev loop? 

Consider that it is easier to parse standard payload formats (like json, yaml etc) - what if someone wanted to build scripts around this functionality for a particular use case. In this case they would have to parse out/handle the basic string texts everytime there is a change.  If the change notifications were for example entirely json it would be easier to consume.

azext_iot/central/providers/devicetwin_provider.py
```diff
@@ -0,0 +1,57 @@
...
+        exception = None
+
+        # The device could be in any hub associated with the given app.
+        # We must search through each IoT Hub until device is found.
+        for token_group in tokens.values():
+            sas_token = token_group["iothubTenantSasToken"]["sasToken"]
+            endpoint = find_between(sas_token, "SharedAccessSignature sr=", "&sig=")
+            target = {"entity": endpoint}
+            auth = BasicSasTokenAuthentication(sas_token=sas_token)
+            service_sdk, errors = _bind_sdk(target, SdkType.service_sdk, auth=auth)
```
digimaun, line 50: Not blocking...I know its how it worked before - just a mental note, `_bind_sdk` is planned to go away. I can take care of this later with some planned refactors.

prbans: Sounds good, happy to assist with that whenever the work comes down the...pipeline

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -197,3 +201,53 @@ def test_should_return_device_template(
...
+        twin_next.reported_property.version = twin.reported_property.version + 1
+        result = compare_properties(twin_next.reported_property, twin.reported_property)
+        assert len(result) == 3
+        assert len(result.get("$iotin:urn_azureiot_Client_SDKInformation")) == 3
+        assert result.get("$iotin:urn_azureiot_Client_SDKInformation").get("language")
+        assert result.get("$iotin:urn_azureiot_Client_SDKInformation").get("version")
+        assert result.get("$iotin:urn_azureiot_Client_SDKInformation").get("vendor")
+
+        assert len(result.get("$iotin:deviceinfo")) == 8
+        assert result.get("$iotin:deviceinfo").get("manufacturer")
```
digimaun, line 227: Seems like you don't need the first `.get()` in a lot of these?

azext_iot/monitor/property.py
prbans: The first iteration is intended for inner dev-loop only. We probably won't change this right now since I think our plan is to use it for customer evaluation/gathering feedback.

Happy to revisit the output style later if we are seeing this used as part of bigger workflows

#wontfix

azext_iot/central/models/devicetwin.py
```diff
@@ -0,0 +1,34 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
```
digimaun, line 6: Cool, good idea

azext_iot/monitor/property.py
prbans: #wontfix this iteration

azext_iot/monitor/property.py
digimaun, line 90: Can you explain more on this decision? `if "$iotin" not in property_name: return prop`....with non-pnp devices all their twin properties are changes?

valluriraj: twin data can be represented as property or collection of properties under interfaces. This version of the PR treats non-pnp updates as property updates and for pnp devices it will check to see if update is a interface update and in which cases iterate thru the properties under interface to identify the updated properties. Next PR will make this generic where it will check if the update is a interface or property and do the needful.

digimaun: Got it thanks. I like the sound of the last sentence :)

prbans:
image

PR #212, opened 2020-07-03
Central property monitor V2

valluriraj:
---

### General Guidelines


azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
-
-            if change_r:
-                print("Changes in reported properties:")
-                print("version :", twin.reported_property.version)
-                print(change_r)
+    def _validate_duplicate_properties(self, property_name):
+        value = (
+            sum(
+                property_name in idnumber
+                for idnumber in self._template.interfaces.values()
```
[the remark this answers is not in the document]
prbans: I'd even argue that this "find interfaces that have a schema with name matching" is useful enough to include in the template class

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+            details=details,
+            message=None,
+            device_id=self._device_id,
+            template_id=self._template.id,
+        )
 
-        time.sleep(polling_interval_seconds)
+    def _is_interface(self, interface_name):
+        # Remove PNP interface prefix to get the actual interface name
+        interface_name = interface_name.replace(PNP_INTERFACE_PREFIX, "")
```
prbans, line 136: I'm a little paranoid, I'd prefer if we just did 

```
return interface_name.replace(PNP_INTERFACE_PREFIX, "") in self._template.interfaces
```

Reason: since strings are mutable, so the caller might see their string changed after this function call

prbans: or used a different variable name to denote the interface_name_modificaiton

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+        return interface_name in self._template.interfaces
 
-        prev_twin = twin
+    def _get_device_template(self):
 
+        central_device_provider = CentralDeviceProvider(self._cmd, self._app_id)
```
prbans, line 141: I would suggest not taking a dependency on other providers within the PropertyProvider, or we risk future circular dependency issues.

We should pass down the template as part of construction (i.e. Controller level).

If we absolutely must take a dependency on the providers, I'd suggest creating the providers in the constructor, then referencing them as needed

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+            raw_twin = device_twin_provider.get_device_twin(
+                central_dns_suffix=self._central_dns_suffix
+            )
 
-def is_relevant(key, val):
-    if key in {"$lastUpdated", "$lastUpdatedVersion"}:
-        return False
+            twin = DeviceTwin(raw_twin)
+            if prev_twin:
+                change_d = self._compare_properties(
```
prbans, line 165: nit: extract code in `if prev_twin` to a function

```
if prev_twin: 
    self._detect_and_print_changes(prev_twin, twin)
```

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+                if change_d:
+                    print("Changes in desired properties:")
+                    print("version :", twin.desired_property.version)
+                    print(change_d)
+
+                if change_r:
+                    print("Changes in reported properties:")
+                    print("version :", twin.reported_property.version)
+                    print(change_r)
+            time.sleep(DEVICETWIN_POLLING_INTERVAL_SEC)
```
prbans, line 181: nit: Make this configurable by requesting it in the constructor

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+
+            prev_twin = twin
+
+    def start_validate_property_monitor(self, minimum_severity):
+        prev_twin = None
+
+        device_twin_provider = CentralDeviceTwinProvider(
```
prbans, line 188: maintenance/pattern: avoid creating providers in functions, try to create them in constructor

prbans: ideally - don't construct other providers in providers, instead have a controller construct them and pass info as required

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -216,7 +219,12 @@ def test_should_return_updated_properties(
...
-        result = compare_properties(twin_next.reported_property, twin.reported_property)
+        monitor = PropertyMonitor(
+            cmd=None, app_id=app_id, device_id=device_id, central_dns_suffix=None,
+        )
+        result = monitor._compare_properties(
+            twin_next.reported_property, twin.reported_property
+        )
         assert len(result) == 3
         assert len(result["$iotin:urn_azureiot_Client_SDKInformation"]) == 3
         assert result["$iotin:urn_azureiot_Client_SDKInformation"]["language"]
```
prbans, line 230: nit: it would be nice if these asserts tested for something more than just existence

azext_iot/central/commands_monitor.py
```diff
@@ -15,7 +15,7 @@
     CentralHandlerArguments,
     TelemetryArguments,
 )
-from azext_iot.monitor.property import start_property_monitor
+from azext_iot.monitor.property import PropertyMonitor
```
prbans, line 18: Action: do NOT import anything from `monitor` namespace until there has been a call to `init_monitoring`

Details:

Since this file is importing stuff from `monitor/` folder, we have to be really careful about users without uamqp. 

Even if your code doesn't use UAMQP, the way imports are done, theres a valid risk that we could break users without uamqp

To do this, we have to do an `init_monitor` call.

So we have 2 paths:

1) init_monitor call before importing anything from `.monitor`. This function gets eventually called as part of instantiating `TelemetryArguments` object.
2) Move this code to `central/` folder to avoid importing from monitor at all (not preferred)

I'd like to eventually delete all references to monitor in the commands_monitor.py class, since our dependency on uamqp makes us more fragile

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,202 @@
...
+        }
+
+        return changes
+
+    def _is_relevant(self, key, val):
+        if key in {"$lastUpdated", "$lastUpdatedVersion"}:
+            return False
+
+        updated_within = datetime.datetime.now() - datetime.timedelta(
+            seconds=DEVICETWIN_MONITOR_TIME_SEC
```
prbans, line 56: nit - might be worth making this configurable (i.e. use a function param that defaults to this value instead)

azext_iot/central/_help.py
```diff
@@ -376,6 +376,20 @@ def _load_central_monitors_help():
             az iot central app monitor-properties --app-id {app_id} -d {device_id}
     """
 
+    helps[
+        "iot central app validate-properties"
+    ] = """
+        type: command
+        short-summary: Validate reported properties sent to the IoT Hub for an IoT Central app.
+        long-summary: |
+                    EXPERIMENTAL requires Python 3.5+
```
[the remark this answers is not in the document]
prbans: Also - lets remove the reference to IoTHub

`Validate reported/desired properties sent to/from an IoT Central app`

azext_iot/monitor/property.py
prbans: #important

azext_iot/monitor/property.py
```diff
@@ -75,15 +74,16 @@ def _changed_props(self, prop, metadata, property_name):
...
-            self._validate_payload_against_interfaces(changes[value], value)
-
-        issues = self._issues_handler.get_issues_with_minimum_severity(minimum_severity)
-
-        for issue in issues:
-            issue.log()
+            issues = self._validate_payload_against_interfaces(
+                changes[value], value, minimum_severity
+            )
+            [issue.log() for issue in issues]
```
prbans, line 80: when not creating a list do the old way:

```
        for issue in issues:
            issue.log()
```

azext_iot/monitor/property.py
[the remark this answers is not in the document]
prbans: This will also remove a minor bug in the code. 

Currently here: https://github.com/Azure/azure-iot-cli-extension/pull/212/files#diff-209395f6b9ec07ce7b03dd88c1122426R105

If a duplicate property is detected, the code will print out all interfaces, rather than just the interfaces that have the duplicate property.

e.g.:

```
interfaces = {'a': 'bool', 'b': 'bool', 'c': 'date'}
dups = _validate_duplicate_properties('bool')
if dups: print...
```

the print result will be:

['a', 'b', 'c']

rather than

['a', 'b']

azext_iot/central/commands_monitor.py
valluriraj: verified that there is no dependency on uamqp with current flow.

azext_iot/monitor/property.py
valluriraj: wont fix

azext_iot/central/_help.py
```diff
@@ -376,6 +376,19 @@ def _load_central_monitors_help():
             az iot central app monitor-properties --app-id {app_id} -d {device_id}
     """
 
+    helps[
+        "iot central app validate-properties"
+    ] = """
+        type: command
+        short-summary: Validate reported properties sent to IoT Central app.
+        long-summary: |
```
prbans, line 384: remove this long summary all together if theres nothing useful in it

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,207 @@
...
+            if self._is_relevant(key, val)
+        }
+        return diff
+
+    def _validate_payload(self, changes, minimum_severity):
+        for value in changes:
+            issues = self._validate_payload_against_interfaces(
+                changes[value], value, minimum_severity
+            )
+            for issue in issues:
```
prbans, line 87: bug: move `for issue in issues` back one indentation, its currently being done for each `value in changes`

azext_iot/monitor/property.py
valluriraj: wont fix

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,207 @@
...
+
+    def _validate_payload_against_interfaces(
+        self, payload: dict, name, minimum_severity
+    ):
+        name_miss = []
+        issues_handler = IssueHandler()
+        interface_name = name.replace(CENTRAL_PNP_INTERFACE_PREFIX, "")
+        if self._is_interface(interface_name):
+            # if the payload is an interface then iterate thru the properties under the interface
+            for property_name in payload.keys():
```
prbans, line 98: nit: no need for `.keys()`

```
for property_name in payload
```

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,207 @@
...
+                schema = self._template.get_schema(
+                    name=property_name, interface_name=interface_name
+                )
+                if not schema:
+                    name_miss.append(property_name)
+        else:
+            # if the payload is a property then process the payload as a single unit.
+            schema = self._template.get_schema(name=name, interface_name="")
```
prbans, line 106: nit: no need for `, interface_name=""`, its already defaulted to empty string

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,207 @@
...
+
+            if not schema:
+                name_miss.append(name)
+
+            interfaces_with_specified_property = self._validate_duplicate_properties(
```
prbans, line 111: maintenance - this isn't validating duplicate properties, its finding a list of interfaces that contain a schema matching "name"

Requested action: rename the function
Suggested action: move function to Template class

azext_iot/monitor/property.py
```diff
@@ -7,91 +7,207 @@
...
+                name
+            )
+            if len(interfaces_with_specified_property) > 1:
```
prbans, line 114: love it, this is super clear now

azext_iot/tests/test_iot_central_unit.py
valluriraj: wont fix

azext_iot/monitor/property.py
prbans: Chatted offline, this is intentional. No bug #closed

azext_iot/central/_help.py
```diff
@@ -390,6 +390,18 @@ def _load_central_monitors_help():
             az iot central app monitor-properties --app-id {app_id} -d {device_id}
     """
 
+    helps[
+        "iot central app validate-properties"
+    ] = """
+        type: command
+        short-summary: Validate reported properties sent to IoT Central app.
```
digimaun, line 397: Validate in what way? How is this command supposed to work/what is the user expecting? Our commands should be as self describing as possible.

azext_iot/monitor/parsers/strings.py
```diff
@@ -73,6 +73,14 @@ def invalid_field_name_mismatch_template(
     ).format(unmodeled_capabilities, modeled_capabilities)
 
 
+# warning
+def duplicate_property_name(duplicate_prop_name, interfaces: list):
+    return (
+        "Duplicate property: '{}' found under following interfaces {} in the device model."
```
digimaun, line 79: nit: Add space at the end of the first sentence.

PR #214, opened 2020-07-11
Eventhub compatible connection-string command

anusapan:
---

### General Guidelines


anusapan:
image

anusapan:
image

azext_iot/_help.py
```diff
@@ -96,6 +96,42 @@
...
+    - name: Show the connection string of an IoT Hub using policy 'service' and secondary key.
+      text: >
+          az iot hub connection-string show -n MyIotHub --policy-name service --key-type secondary
+    - name: Show the connection strings for all IoT hubs in a resource group.
+      text: >
+          az iot hub connection-string show --resource-group MyResourceGroup
+    - name: Show the connection strings for all IoT hubs in a subscription.
+      text: >
+          az iot hub connection-string show
+    - name: Show the connection strings for an IoT hub. (autogenerated)
```
digimaun, line 130: autogenerated help stuff is not something we explicitly put in our help. There is another component responsible for creating the autogenerated help.

anusapan: fixed

azext_iot/_help.py
```diff
@@ -96,6 +96,42 @@
...
+          az iot hub connection-string show -n MyIotHub --policy-name service --key-type secondary
+    - name: Show the connection strings for all IoT hubs in a resource group.
+      text: >
+          az iot hub connection-string show --resource-group MyResourceGroup
+    - name: Show the connection strings for all IoT hubs in a subscription.
+      text: >
+          az iot hub connection-string show
+    - name: Show the connection strings for an IoT hub.
+      text: az iot hub connection-string show --key primary --policy-name MyPolicy
+      crafted: true
```
digimaun, line 132: `crafted: true`? Should be deleted

azext_iot/_help.py
```diff
@@ -96,6 +96,42 @@
...
+    - name: Show the connection string of an IoT Hub using policy 'service' and secondary key.
+      text: >
+          az iot hub connection-string show -n MyIotHub --policy-name service --key-type secondary
+    - name: Show the connection strings for all IoT hubs in a resource group.
+      text: >
+          az iot hub connection-string show --resource-group MyResourceGroup
+    - name: Show the connection strings for all IoT hubs in a subscription.
+      text: >
+          az iot hub connection-string show
+    - name: Show the connection strings for an IoT hub.
```
digimaun, line 130: can we re-arrange these and sort by increasing complexity for the command?

azext_iot/_params.py
```diff
@@ -263,6 +263,18 @@ def load_arguments(self, _):
...
 
+    with self.argument_context("iot hub connection-string") as context:
+        context.argument(
+            "show_all",
+            options_list=["--show-all", "--all"],
+            help="Allow to show all shared access policies."
+        )
+        context.argument(
+            "include_eventhub",
+            options_list=["--include-eventhub", "--eh"],
```
digimaun, line 274: Instead of `--include-eventhub` let's do `--default-eventhub`

azext_iot/operations/hub.py
```diff
@@ -2434,6 +2434,96 @@ def iot_hub_distributed_tracing_update(
...
+        def conn_str_getter(hub):
+            return _get_hub_connection_string(
+                discovery,
+                hub,
+                policy_name,
+                key_type,
+                show_all,
+                include_eventhub
+            )
+        return [{'name': hub.name, 'connectionString': conn_str_getter(hub)} for hub in hubs]
```
digimaun, line 2462: The return format when not specifying a hub is not consistent with the return format when given a target hub. In this case `connectionString` value is always a list.

If there is a target hub, then `connectionString` becomes a string value unless `--show-all` is provided in which case it becomes a list.

azext_iot/operations/hub.py
```diff
@@ -2434,6 +2434,96 @@ def iot_hub_distributed_tracing_update(
...
+        cs_template_eventhub = 'Endpoint={};SharedAccessKeyName={};SharedAccessKey={};EntityPath={}'
+        endpoint = hub.properties.event_hub_endpoints['events'].endpoint
+        entityPath = hub.properties.event_hub_endpoints['events'].path
+        return [
+            cs_template_eventhub.format(
+                endpoint,
+                p.key_name,
+                p.secondary_key if key_type == KeyType.secondary.value else p.primary_key,
+                entityPath
+            ) for p in policies if "serviceconnect" in p.rights.value.lower()
```
digimaun, line 2512: We should state this filtering behavior in the help description for the command so users aren't confused why a policy connection string doesn't show up when --include-eventhub (to be --default-eventhub)  is passed in.

azext_iot/_help.py
anusapan: removed

azext_iot/_params.py
anusapan: updated

azext_iot/_help.py
anusapan: Done

azext_iot/operations/hub.py
anusapan: Mentioned in one of the examples in help content.

azext_iot/operations/hub.py
anusapan: Fixed

azext_iot/_help.py
```diff
@@ -96,6 +96,39 @@
...
+    - name: Show the connection strings for all IoT hubs in a subscription.
+      text: >
+          az iot hub connection-string show
+    - name: Show the connection strings for all IoT hubs in a resource group.
+      text: >
+          az iot hub connection-string show --resource-group MyResourceGroup
+    - name: Show all the connection string of an IoT hub using default key.
+      text: >
+          az iot hub connection-string show -n MyIotHub --all
+    - name: Show the connection string of an IoT hub using default policy and primary key.
```
digimaun, line 121: Because this becomes rendered for end users, can we be consistent with the capitalization of 'IoT Hub'?  In some cases, like this line its 'IoT **h**ub'.

azext_iot/_help.py
```diff
@@ -96,6 +96,39 @@
...
+    short-summary: Manage IoT hub\'s connection string.
+"""
+
+helps[
+    "iot hub connection-string show"
+] = """
+    type: command
+    short-summary: Show the connection strings for an IoT hub.
+    examples:
+    - name: Show the connection strings for all IoT hubs in a subscription.
```
digimaun, line 112: We should append ' using the default policy and primary key.' (like in below examples)  For the by subscription/by resource group examples when --all is not provided.

azext_iot/_help.py
anusapan: Fixed

azext_iot/_help.py
anusapan: Fixed

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -71,6 +71,33 @@ def test_hub(self):
             expect_failure=True,
         )
 
+        # Test 'az iot hub connection-string show'
+        conn_str_pattern = r'^HostName={0}.azure-devices.net;SharedAccessKeyName=iothubowner;SharedAccessKey='.format(
+            LIVE_HUB)
+        conn_str_eventhub_pattern = r'^Endpoint=sb://'
+        self.cmd('iot hub connection-string show -n {0}'.format(LIVE_HUB), checks=[
```
digimaun, line 78: Can we also please add tests for showing connection-string by resource group or by sub? We know the sub or group should have at least 1  IoT Hub and the count of iothubs in a group should be <= count of iothubs in sub.

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -71,6 +71,33 @@ def test_hub(self):
...
+            self.check_pattern('connectionString', conn_str_eventhub_pattern)
+        ])
+
+        self.cmd('iot hub connection-string show -n {0} -g {1}'.format(LIVE_HUB, LIVE_RG), checks=[
+            self.check('length(@)', 1),
+            self.check_pattern('connectionString', conn_str_pattern)
+        ])
+
+        self.cmd('iot hub connection-string show -n {0} -g {1} --all'.format(LIVE_HUB, LIVE_RG), checks=[
+            self.check('length(connectionString[*])', 5),
```
digimaun, line 92: I'm uneasy asserting a particular length for the connectionString list, because for example if your test hub has an extra policy (for whatever reason) this test will fail. I think its safer to check if the length of the list is >= 1, because at least one policy should exist to interact with the iothub dataplane.

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -71,6 +71,33 @@ def test_hub(self):
...
+            self.check_pattern('connectionString[0]', conn_str_pattern)
+        ])
+
+        self.cmd('iot hub connection-string show -n {0} -g {1} --all --eh'.format(LIVE_HUB, LIVE_RG), checks=[
+            self.check('length(connectionString[*])', 2),
```
digimaun, line 97: Same thoughts for checking the length of the connectionString list here. Also in the default eventhub case we can expect at least 1 policy by design.

anusapan: Done

anusapan:
image

azext_iot/tests/test_iot_ext_int.py
anusapan: Done

azext_iot/tests/test_iot_ext_int.py
anusapan: Added new test case

PR #220, opened 2020-07-21
validate-messages enhancements

valluriraj:
---

### General Guidelines


azext_iot/monitor/parsers/strings.py
```diff
@@ -83,11 +83,8 @@ def duplicate_property_name(duplicate_prop_name, interfaces: list):
 
 # error
 def invalid_primitive_schema_mismatch_template(field_name: str, data_type: str, data):
-    return (
-        "Datatype of field '{}' does not match the datatype '{}'. Data '{}'. "
-        "All dates/times/datetimes/durations must be ISO 8601 compliant.".format(
-            field_name, data_type, data,
-        )
+    return "Datatype of field '{}' does not match the datatype {}. Data sent by the device - {}. ".format(
```
digimaun, line 86: Is "field" the right language? Any other references I have seen refers to these elements as "property".

prbans: I went with this because I thought there could be confusion with "device_properties". But its a nit, so feel free to change to property if we want to do that instead

azext_iot/monitor/parsers/strings.py
```diff
@@ -84,11 +84,9 @@ def duplicate_property_name(duplicate_prop_name, interfaces: list):
 # error
 def invalid_primitive_schema_mismatch_template(field_name: str, data_type: str, data):
     return (
-        "Datatype of field '{}' does not match the datatype '{}'. Data '{}'. "
-        "All dates/times/datetimes/durations must be ISO 8601 compliant.".format(
-            field_name, data_type, data,
-        )
-    )
+        "Datatype of telemetry field '{}' does not match the datatype {}. Data sent by the device - {}. "
+        "For information about format of the payload refer: https://aka.ms/iotcentral-payloads"
```
prbans, line 88: Maybe "For more information about expected JSON payload shape and contents, please refer to the information found here: {}".format(field_name, data_type, data, ARTICLE_LINK)

so that we can just update the link if we need to

valluriraj:
image

azext_iot/monitor/parsers/strings.py
valluriraj: updated per feedback from PM

PR #224, opened 2020-07-31
Allow blob sas uri from file as input

anusapan:
---

### General Guidelines

### DIFF
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -198,10 +198,10 @@ helps[
     long-summary: For more information, see
                   https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#import-and-export-device-identities
     examples:
-    - name: Export all device identities to blob container using SAS uri.
+    - name: Export all device identities to a configured blob container using an inline SAS uri.
       text: >
         az iot hub device-identity export -n {iothub_name} --bcu {sas_uri}
-    - name: Export all device identities to blob container using file which contains Blob SAS uri.
+    - name: Export all device identities to a configured blob container using a file path which contains SAS uri.
       text: >
         az iot hub device-identity export -n {iothub_name} --bcu {sas_uri_filepath}
 """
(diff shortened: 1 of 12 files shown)


anusapan:
Integration Test

image

anusapan:
Unit test

image

azext_iot/_params.py
```diff
@@ -382,7 +382,7 @@ def load_arguments(self, _):
             help="Blob Shared Access Signature URI with write, read, and delete access to "
             "a blob container. This is used to output the status of the "
             "job and the results. Note: when using Identity-based authentication, you must supply an "
```
digimaun, line 384: I think this makes the last part of the text more clear (same application for all the descriptions) 

"Note: when using Identity-based authentication an https:// URI is still required. Input for this argument can be inline or from a file path."

azext_iot/_help.py
```diff
@@ -197,6 +197,13 @@
     short-summary: Export all device identities from an IoT Hub to an Azure Storage blob container.
     long-summary: For more information, see
                   https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#import-and-export-device-identities
+    examples:
+    - name: Export all device identities to blob container using SAS uri.
+      text: >
+        az iot hub device-identity export -n {iothub_name} --bcu {sas_uri}
+    - name: Export all device identities to blob container using file which contains Blob SAS uri.
```
digimaun, line 204: super nit: can we lowercase "Blob" near the end

azext_iot/_help.py
```diff
@@ -197,6 +197,13 @@
     short-summary: Export all device identities from an IoT Hub to an Azure Storage blob container.
     long-summary: For more information, see
                   https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#import-and-export-device-identities
+    examples:
+    - name: Export all device identities to blob container using SAS uri.
```
digimaun, line 201: "Export all device identities to a configured blob container using an inline SAS uri."

PR #226, opened 2020-08-03
Add api token commands

valluriraj:
---

### General Guidelines


azext_iot/central/_help.py
```diff
@@ -253,6 +254,68 @@ def _load_central_users_help():
...
+        "iot central app api-token show"
+    ] = """
+    type: command
+    short-summary: Get an API token by ID
+    examples:
+      - name: Get API token
+        text: >
+          az iot central app api-token show
+          --app-id {appid}
+          --token-id {tokenId}
```
prbans, line 289: Can we double check that this API actually returns a token to the developer? If not - lets remove it, I think it misleads our customers into thinking they can programmatically fetch a token. This call essentially just checks for existence of the token.

Let me know if PM has strong desire to keep this, we can work with them internally

azext_iot/central/_help.py
```diff
@@ -253,6 +254,68 @@ def _load_central_users_help():
...
+          az iot central app api-token delete
+          --app-id {appid}
+          --token-id {tokenId}
+    """
+
+    helps[
+        "iot central app api-token list"
+    ] = """
+    type: command
+    short-summary: Get list of API tokens in an application
```
prbans, line 309: same issue with List as with Get

azext_iot/tests/test_iot_central_int.py
```diff
@@ -241,6 +241,32 @@ def test_central_user_methods_CRD(self):
...
+            "iot central app api-token show --app-id {} --token-id {}".format(
+                APP_ID, tokens[0].get("id")
+            ),
+        )
+
+        result = self.cmd(
+            "iot central app api-token list --app-id {}".format(APP_ID,),
+        ).get_output_in_json()
+
+        token_list = result.get("value")
```
prbans, line 257: does this imply pagination on tokens?

prbans:
@valluriraj

azext_iot/central/_help.py
lmasieri: @prbans @valluriraj The use case for this is to list the tokens you've generated, and then be able to revoke them if needed. We can talk to Vivian if more info is needed. Let's keep parity with the API.

valluriraj:
image

azext_iot/tests/test_iot_central_int.py
valluriraj: pagination is not supported on this API

azext_iot/central/_help.py
```diff
@@ -253,6 +254,70 @@ def _load_central_users_help():
...
+    ] = """
+        type: group
+        short-summary: Create and Manage API tokens .
+    """
+
+    helps[
+        "iot central app api-token create"
+    ] = """
+        type: command
+        short-summary: Create a new API token in the application
```
prbans, line 269: nit:

```
long-summary: The only time you will see the value of this token is when creating the token. Ensure you store this token somewhere securely, as if you lose it, you will need to create another.
```

azext_iot/central/_help.py
```diff
@@ -253,6 +254,70 @@ def _load_central_users_help():
...
+            --token-id {tokenId}
+            --app-id {appId}
+            --role admin
+
+    """
+    helps[
+        "iot central app api-token show"
+    ] = """
+    type: command
+    short-summary: Get an API token by ID
```
prbans, line 283: nit:

```
short-summary: Get token meta data (e.g. role as a GUID, expiration)
```

azext_iot/central/_help.py
```diff
@@ -253,6 +254,70 @@ def _load_central_users_help():
...
+    """
+
+    helps[
+        "iot central app api-token list"
+    ] = """
+    type: command
+    short-summary: Get list of API tokens in an application
+    long-summary: Information in the list contains basic information about the tokens in the application and does not include token values.
+    examples:
+      - name: List of API tokens
```
prbans, line 313: nit: 

```
short-summary: Get a list of all token meta data (e.g. Role as a GUID and expiration)
```

valluriraj:
@digimaun

azext_iot/central/_help.py
digimaun: Agree on adding additional info like this, long-summary is a good place. These are the kind of additional infos that really help user understanding and may reduce support tickets/issues down the line.

azext_iot/central/params.py
```diff
@@ -120,6 +120,12 @@ def load_central_arguments(self, _):
             help="Tenant ID for service principal to be added to the app. Object ID must also be specified. ",
         )
 
+        context.argument(
+            "token_id",
+            options_list=["--token-id", "--tkid"],
```
digimaun, line 125: Specifically for the short aliases of "-id" based arguments, as a guideline for other command groups (and trying to adhere moving forward), we want to try and make them single character where possible.

I'm not necessarily saying make the change here (and there are multiple -id args for this command group) but keep in mind the guideline. The ease of understanding and usability of args is high pri.

azext_iot/central/services/api_token.py
```diff
@@ -0,0 +1,125 @@
...
+        token_id: Unique ID for the API token.
+        role : permission level to access the application
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+    token: dict
+    """
+    url = "https://{}.{}/{}/{}".format(app_id, central_dns_suffix, BASE_PATH, token_id)
```
digimaun, line 42: Are there no api versions set for the central service calls?

prbans: Its in the `BASE_PATH`, we currently only have the `/preview/` version

digimaun: I see, thanks. I hope there is not an inadvertent API behavior change that could break something on clients. For ARM services typically there is an api version query string with an exact version (even for preview) to avoid/reduce integration issues.

digimaun:
I would like to see at least the additional `long-summary` info suggested for `az iot central app api-token create` and then will approve.

PR #227, opened 2020-08-07
Removal of _bind_sdk and related components + misc improvements.

digimaun:
image

---

### General Guidelines


azext_iot/operations/dps.py
```diff
@@ -14,7 +14,7 @@
 from azext_iot.common.utility import shell_safe_json_parse
 from azext_iot.common.certops import open_certificate
 from azext_iot.operations.generic import _execute_query
-from azext_iot._factory import _bind_sdk
+from azext_iot._factory import SdkResolver
```
prbans, line 17: nit - should this _factory be changed to factory since the _ implies private?

azext_iot/sdk/service/models/x509_thumbprint.py
```diff
@@ -1,32 +0,0 @@
...
-#
-# Code generated by Microsoft (R) AutoRest Code Generator.
-# Changes may cause incorrect behavior and will be lost if the code is
-# regenerated.
-# --------------------------------------------------------------------------
-
-from msrest.serialization import Model
-
-
-class X509Thumbprint(Model):
```
prbans, line 15: why did we get rid of all the models?

azext_iot/tests/test_iot_ext_int.py
```diff
@@ -719,9 +719,7 @@ def test_hub_device_twins(self):
 
         # Region specific test
         if self.region not in ["West US 2", "North Europe", "Southeast Asia"]:
-            warnings.warn(
-                "Skipping distributed-tracing tests. IoT Hub not in supported region!"
-            )
+            warnings.warn(UserWarning("Skipping distributed-tracing tests. IoT Hub not in supported region!"))
```
prbans, line 722: should we be using UserWarning elsewhere too? What benefits does it give us?

azext_iot/sdk/service/models/x509_thumbprint.py
digimaun: These models were for the old service SDK which nothing uses anymore after this PR -- there are "refreshed" models (as in refreshed for the IoT Hub API version 2019-10-01, in the sdk/iothub/service directory.

azext_iot/operations/dps.py
digimaun: Yea agreed there is a peculiarity with the underscore here, but will punt for later.

azext_iot/tests/test_iot_ext_int.py
digimaun: The benefit here is pytest will automatically captures warnings from tests and shows them after the run. This was put in place to explain why some inner tests were skipped and the reason why. This pattern was used rather than say a `pytest.skip` or similar because of the way the tests were written and not wanting to show the whole test skipped when only the last portion is skipped.

digimaun: At the user interaction layer, if you want to show a warning logger.warning() is what you want to use.

PR #230, opened 2020-08-10
Add cde commands

valluriraj:
---

### General Guidelines


azext_iot/central/params.py
```diff
@@ -126,6 +134,44 @@ def load_central_arguments(self, _):
             help="Unique ID for the API token. ",
         )
 
+        context.argument(
+            "export_id",
+            options_list=["--export-id", "-d"],
```
prbans, line 139: nit - remove "-d" flag

azext_iot/central/params.py
```diff
@@ -42,6 +49,7 @@ def load_central_arguments(self, _):
         context.argument("app_id", options_list=["--app-id", "-n"], help="Target App.")
         context.argument("minimum_severity", arg_type=severity_type)
         context.argument("role", arg_type=role_type)
+        context.argument("sources", arg_type=sources)
```
prbans, line 52: IMPT:

Since we are starting to add a bunch of commands, could you actually reorganize this file as part of this or next PR?

We should group all the command specific params to their parent group. See the following example:

```
with self.argument_context("iot central app validate-messages") as context:
        context.argument("timeout", arg_type=event_timeout_type)
        context.argument("properties", arg_type=event_msg_prop_type)
        context.argument("style", arg_type=style_type)
        context.argument(
            "duration",
            options_list=["--duration", "--dr"],
            type=int,
            help="Maximum duration to receive messages from target device before terminating connection."
            "Use 0 for infinity.",
        )
        context.argument(
            "max_messages",
            options_list=["--max-messages", "--mm"],
            type=int,
            help="Maximum number of messages to recieve from target device before terminating connection."
            "Use 0 for infinity.",
        )
```

For example, if we wanted to do this with CDE:

```
with self.argument_context("iot central app cde") as context:
        context.argument(
            "ep_type",
            options_list=["--ep-type", "-t"],
            arg_type=get_enum_type(EndpointType),
            help="Type of endpoint where exported data should be sent to. ",
        )
```

If we push this till later - we run a bigger risk of having clashing param names/arguments

azext_iot/tests/test_iot_central_int.py
```diff
@@ -267,6 +274,45 @@ def test_central_api_token_methods_CRD(self):
             }
             assert token_info_basic in token_list
 
+    @pytest.mark.skipif(not STORAGE_CS, reason="empty azext_iot_central_sa_cs env var")
```
prbans, line 277: Since we are adding a bunch of IT's and this file is growing to ~500lines now, should we think about splitting the IT classes up a bit more?

IoTCentralDeviceAndDeviceTemplateIntegrationTests
IoTCentralUserTests
IoTCentralCDETests
IoTCentral...

azext_iot/central/_help.py
```diff
@@ -319,6 +320,95 @@ def _load_central_api_token_help():
     """
 
 
+def _load_central_cde_help():
+    helps[
+        "iot central app cde"
+    ] = """
+        type: group
+        short-summary: Create and Manage Continuous Data Export (CDE) for an application.
```
prbans, line 328: maybe an aka.ms link or a `long-summary` explaining what CDE is?

azext_iot/central/models/enum.py
```diff
@@ -41,3 +41,24 @@ class UserType(Enum):
...
+
+
+class DataSourceType(Enum):
+    """
+    Type of Data Sources to export to the endpoint
+    """
+
+    devices = "devices"
+    deviceTemplates = "deviceTemplates"
+    telemetry = "telemetry"
```
prbans, line 64: do we want to specify an "all" flag?

azext_iot/tests/test_iot_central_int.py
prbans: this one is optional, only do this if you think its easier to debug issues when splitting the tests up by use case

valluriraj:
@digimaun

azext_iot/central/_help.py
```diff
@@ -319,6 +320,96 @@ def _load_central_api_token_help():
...
+        long-summary: Continuous Data Export enables exporting IoT Central measurements, devices, and device templates data to storage account .
```
prbans, line 329: `account .` nit extra space between . and last work

prbans: we probably need this long-summary for each CDE command we are introducing

azext_iot/central/commands_cde.py
```diff
@@ -0,0 +1,94 @@
...
+    enable,
+    ep_type: EndpointType,
+    ep_conn,
+    entity_name,
+    sources,
+    app_id: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+):
+    provider = CentralContDataExportProvider(cmd=cmd, app_id=app_id, token=token)
```
prbans, line 78: I'd prefer CentralCdeProvider or the full name (CentralContinuousDataExportProvider)

azext_iot/central/params.py
```diff
@@ -99,6 +106,8 @@ def load_central_arguments(self, _):
         context.argument(
             "module_id", options_list=["--module-id", "-m"], help="Iot Edge Module ID",
         )
+
+    with self.argument_context("iot central app user") as context:
         context.argument(
             "assignee",
             options_list=["--user-id", "--assignee"],
```
prbans, line 113: @digimaun @valluriraj  how do you feel about removing the individual "ids" we have everywhere (user-id, export-id, token-id, etc) and combining them all under a single "--id" and "-i" flag?

azext_iot/central/commands_cde.py
```diff
@@ -0,0 +1,94 @@
...
+    cmd,
+    export_id,
+    enable,
+    ep_type: EndpointType,
+    ep_conn,
+    entity_name,
+    sources,
+    app_id: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
```
prbans, line 76: you'll need to add an exception to the linter for this:

https://github.com/Azure/azure-iot-cli-extension/blob/dev/linter_exclusions.yml

prbans: related error: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=227&view=logs&j=0c36c994-537b-5a38-1422-171fa1397b97&t=9c4e5aac-17ae-589a-22eb-fa659363fcd7&l=130

azext_iot/central/params.py
digimaun: Can you describe a modified command example to make sure we understand the proposal? Here are my initial thoughts:

I am not a fan of using plain "--id" because "--id" or "--ids" typically refer to fully qualified Azure resource id's. Also some of the central commands, such as user creation have lots of *-id arguments. For example user create could potentially have 4 different required Id's.  If the idea is cut down to a single arg that is expecting json, I think that would increase usage difficulty by forcing users to deal with json. Also a user may find it easier to understand a command that has separated arguments in this category from the discoverability of documentation/--help per argument.

linter_exclusions.yml
```diff
@@ -13,3 +13,10 @@ dt endpoint create eventhub:
     eventhub_resource_group:
       rule_exclusions:
       - parameter_should_not_end_in_resource_group
+central app cde create:
```
digimaun, line 16: It's the update command that is specifically giving the command table linter issue.

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+
+    helps[
+        "iot central app cde list"
+    ] = """
+    type: command
+    short-summary: Get the list of continuous data exports in an application.
+    examples:
+      - name: List of continuous data exports
+        text: >
+          az iot central app api-token list
```
digimaun, line 385: This example entry is invalid for this command (the other thing the command table linter raised).

linter_exclusions.yml
```diff
@@ -13,3 +13,10 @@ dt endpoint create eventhub:
     eventhub_resource_group:
       rule_exclusions:
       - parameter_should_not_end_in_resource_group
+central app cde update:
```
digimaun, line 16: You will need to add the "iot" prefix for the command group. In the digital twin case, dt is the root group.

linter_exclusions.yml
```diff
@@ -13,3 +13,10 @@ dt endpoint create eventhub:
...
+  parameters:
+    central_dns_suffix:
+      rule_exclusions:
+      - no_parameter_defaults_for_update_commands
+
+
```
digimaun, line 22: nit: end of this file should contain a single newline

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+        type: command
+        short-summary: Create a new continuous data export
+        examples:
+        - name: Add new continuous data export
+          text: >
+            az iot central app cde create
+            --app-id {appid}
+            --display-name {displayName}
+            --sources "devices, deviceTemplates, telemetry"
+            --enable True,
```
digimaun, line 344: With the default three state flag, you don't need to include the boolean value for true. Just `--enable` will return `True` for your command handlers arg.

Also you are able to use lowercase true/false.

azext_iot/tests/test_iot_central_int.py
```diff
@@ -8,18 +8,25 @@
...
+    EndpointType,
+)
 from azext_iot.monitor.parsers import strings
 
 from . import CaptureOutputLiveScenarioTest, helpers
 
 APP_ID = os.environ.get("azext_iot_central_app_id")
-
+STORAGE_CS = os.environ.get("azext_iot_central_sa_cs")
+STORAGE_CONTRAINER = os.environ.get("azext_iot_central_sa_cn")
```
digimaun, line 29: Spelling for STORAGE_CONTAINER

azext_iot/central/params.py
```diff
@@ -120,12 +129,48 @@ def load_central_arguments(self, _):
...
+        )
+        context.argument(
+            "ep_conn",
+            options_list=["--ep-conn", "-c"],
+            help="Connection string for the endpoint. ",
+        )
+        context.argument(
+            "entity_name",
+            options_list=["--entity-name", "--en"],
+            help="Name of entity pointing at Eg: container_name, queue_name, etc.. ",
```
digimaun, line 160: I think this description for --entity-name needs to be more clear and related the grammar seems off.

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+            --ep-type {epType}
+            --ep-conn {epConnStr}
+            --entity-name {entityName}
```
digimaun, line 347: Just looking at examples here and reading the --entity-name description (from a users perspective), I am confused how the arg should be used. Maybe some actual value sample(s) instead of placeholders would help?

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+
+    """
+
+    helps[
+        "iot central app cde update"
+    ] = """
+    type: command
+    short-summary: Update continuous data export in an application.
+    examples:
+      - name: update continuous data export
```
digimaun, line 396: nit: Capitalize 'update'

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+        text: >
+          az iot central app cde update
+          --app-id {appid}
+          --sources "devices,deviceTemplates,telemetry"
+          --enable True,
```
digimaun, line 401: Just double checking, should --enable be false by default? Is it more common to create a CDE entry where you don't want it to be enabled?

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+    ] = """
+        type: command
+        short-summary: Create a new continuous data export
+        examples:
+        - name: Add new continuous data export
+          text: >
+            az iot central app cde create
+            --app-id {appid}
+            --display-name {displayName}
+            --sources "devices, deviceTemplates, telemetry"
```
digimaun, line 343: With the way the sources arg is defined, this example with quotes around it is going to be taken as a single value.

azext_iot/tests/test_iot_central_int.py
```diff
@@ -475,6 +521,40 @@ def _delete_user(self, user_id) -> None:
             checks=[self.check("result", "success")],
         )
 
+    def _create_cde(self,):
+
+        cde = []
+        for source in DataSourceType:
+            display_name = self.create_random_name(prefix="aztest", length=24)
+            export_id = display_name
+            command = "iot central app cde create --app-id {} --dn {} -d {} -s {} -e False -t {} -c {}  --en {} ".format(
```
digimaun, line 530: I would like to see a test with multiple DataSourceType options at once.

azext_iot/central/_help.py
digimaun: Q's: What can I put in --entity-name if I am not using a storage endpoint? Does the 'entity' need to exist already?

For example, if I _am_ using storage do I have to point to an existing thing -- i.e. if I point to a blob container that doesn't exist, what happens?

azext_iot/tests/test_iot_central_int.py
digimaun, line 530: Why not create the CDE as enabled since there is a real storage account?

azext_iot/central/_help.py
digimaun: EventHub, ServiceBus Queue and ServiceBus Topic all have connection strings that identify themselves in the parent namespace with the "EntityPath" part. So that part could be parsed out and fed into the name/entity-name request property.

_If_ I understand `--entity-name` correctly, it seems `--entity-name` can be optional, and possibly renamed to `--blob-container-name` or similar since only the Storage Account connection string does not include nested resources (a target blob container to export to).

azext_iot/tests/test_iot_central_int.py
valluriraj: Status of the  CDE could be in one of the two possible states  - starting / running. Validation will have to account for possible states and will introduce slight complexity.  From a testing standpoint running state of stopped is more deterministic way of validation.

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
     """
 
 
+def _load_central_cde_help():
+    helps[
+        "iot central app cde"
+    ] = """
+        type: group
+        short-summary: Create and Manage Continuous Data Export (CDE) for an application.
```
rafarhat, line 328: With current release of CDE V2 we are going to start calling this CDE , CDE (legacy)

azext_iot/central/_help.py
```diff
@@ -319,6 +320,92 @@ def _load_central_api_token_help():
...
+        short-summary: Create a new continuous data export
+        examples:
+        - name: Add new continuous data export
+          text: >
+            az iot central app cde create
+            --app-id {appid}
+            --display-name {displayName}
+            --sources "devices, deviceTemplates, telemetry"
+            --enable True,
+            --ep-type {epType}
```
rafarhat, line 345: nit: would rather spell out endpoint instead of ep

azext_iot/central/_help.py
valluriraj: EventHub, ServiceBus Queue and ServiceBus Topic have connection string and entity names - similar model as StorageEndpoint. Would prefer to keep the option set similar

azext_iot/central/params.py
valluriraj: Agree with using separate identifiers

PR #236, opened 2020-08-14
Add keygen command

valluriraj:
---

### General Guidelines


azext_iot/central/_help.py
```diff
@@ -396,6 +397,28 @@ def _load_central_device_twin_help():
...
+        type: group
+        short-summary: Generate a device SAS key
+    """
+
+    helps[
+        "iot central dps-keygen create"
+    ] = """
+        type: command
+        short-summary: Create a device SAS key
+        Long-summary: Generate device key from primary group SAS token. Generated device SAS key is used to connect the device to central.
```
prbans, line 413: Might be worth mentioning a big use case for this functionality is offline key generation for factory scenarios

azext_iot/central/commands_device.py
```diff
@@ -120,3 +120,10 @@ def get_credentials(
     return provider.get_device_credentials(
         device_id=device_id, central_dns_suffix=central_dns_suffix,
     )
+
+
+def generate_device_key(
+    cmd, masterkey, device_id,
+):
+    provider = CentralDeviceProvider(cmd=cmd, app_id=None, token=None,)
```
prbans, line 128: Since a big use case for this is offline generation of dps keys, lets remove the provider pattern in this one case

azext_iot/central/providers/device_provider.py
```diff
@@ -240,3 +240,10 @@ def _dps_populate_essential_info(self, dps_info, device_status: DeviceStatus):
             "error": error.get(device_status),
         }
         return filtered_dps_info
+
+    def generate_device_key(
+        self, masterkey, device_id,
+    ):
+        return central_services.auth.generate_device_key(
```
prbans, line 247: this doesnt need to be in the service layer since it isnt making any network calls

azext_iot/central/services/auth.py
```diff
@@ -0,0 +1,27 @@
...
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+import base64
+import hmac
+import hashlib
+
+
+def generate_device_key(masterkey, device_id):
```
prbans, line 11: I'd suggest maybe creating a new folder under central like /offline or something and placing this there instead

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -123,9 +123,21 @@ class Cmd:
         }
 
 
+class TestCentralDeviceAuth:
```
prbans, line 126: Important - Add an integration test:

1) Fetch SAS group token from API (if exists), or loads from config file
2) Create key using dps-keygen
3) Provision the device using the device sdk that we use to send messages to IoT Central
4) connect the device
5) Send a message to central, check if it got there using `monitor-events`, or check that the device_state is "registered" or something

azext_iot/central/params.py
```diff
@@ -148,3 +148,11 @@ def load_central_arguments(self, _):
             help="Maximum number of messages to recieve from target device before terminating connection."
             "Use 0 for infinity.",
         )
+
+    with self.argument_context("iot central dps-keygen") as context:
+
+        context.argument(
+            "masterkey",
+            options_list=["--masterkey", "-k"],
+            help="Masterkey - Primary group SAS token to generate device keys",
```
prbans, line 157: In the interest of being more inclusive I've heard things like Microsoft is moving away from "master" in their terminology. We should update this to be something like "basekey", "rootkey", "trustedkey", or "primarykey" or similar

azext_iot/central/params.py
```diff
@@ -148,3 +148,11 @@ def load_central_arguments(self, _):
             help="Maximum number of messages to recieve from target device before terminating connection."
             "Use 0 for infinity.",
         )
+
+    with self.argument_context("iot central dps-keygen") as context:
```
prbans, line 152: is this supposed to be at the top level? I agree with having it at the top level since its so different from our other commands and to emphasize it being offline, but the code is structured such that it looks like it belongs under the 'device' command-subgroup

azext_iot/central/providers/device_provider.py
```diff
@@ -240,3 +240,10 @@ def _dps_populate_essential_info(self, dps_info, device_status: DeviceStatus):
             "error": error.get(device_status),
         }
         return filtered_dps_info
+
+    def generate_device_key(
```
prbans, line 244: I'd suggest as a one off not putting this inside a provider, and just calling directly into utils for now

azext_iot/central/params.py
prbans: This will also be more consistent with our docs terminology, since we don't seem to use the word master there at all

image

azext_iot/central/services/auth.py
```diff
@@ -0,0 +1,27 @@
...
+    """
+    Generate device SAS key
+
+    Args:
+        masterkey: Primary group SAS token to generate device keys
+        device_id: unique case-sensitive device id
+    Returns:
+        device key
+    """
+    secret = base64.b64decode(masterkey)
```
prbans, line 21: Fantastic work on getting this to work btw!! Definitely not an easy issue to debug

azext_iot/central/providers/device_provider.py
prbans: or offline, whichever route you go down

azext_iot/central/_help.py
```diff
@@ -396,6 +397,28 @@ def _load_central_device_twin_help():
     """
 
 
+def _load_central_dps_keygen_help():
+    helps[
+        "iot central dps-keygen"
```
prbans, line 402: actually on second thought - we should actually make the entry point for this be in `dps` shouldn't we? cc: @digimaun 

Something like

`az iot dps compute-derived-sas-key`

prbans: I also think this further adds to the argument that this shuold be placed in utility or somewhere more common for consumption

digimaun: If this is a general solution then I agree it shouldn't be in the iot central command space. I could see it landing in `az iot dps` or `az iot device`. Need to brainstorm more on the actual command name, not crazy about `compute-derived-sas-key`.

PR #250, opened 2020-09-02
Implement convenience arguments for device update

anusapan:
---

### General Guidelines


anusapan:
Unit Test
image

Integration Test 
image

anusapan:
Fix for issue https://github.com/Azure/azure-cli/issues/2531

azext_iot/operations/hub.py
```diff
@@ -246,6 +246,38 @@ def _create_self_signed_cert(subject, valid_days, output_path=None):
...
+        if auth_method == DeviceAuthType.shared_private_key.name:
+            auth = 'sas'
+            if (primary_key and not secondary_key) or (not primary_key and secondary_key):
+                raise CLIError("primary + secondary Key required with sas auth")
+            instance['authentication']['symmetricKey']['primaryKey'] = primary_key
+            instance['authentication']['symmetricKey']['secondaryKey'] = secondary_key
+        elif auth_method == DeviceAuthType.x509_thumbprint.name:
+            auth = 'selfSigned'
+            if not any([primary_thumbprint, secondary_thumbprint]):
+                raise CLIError("primary + secondary Thumbprint required with selfSigned auth")
```
digimaun, line 268: If this is doing `not any([primary_thumbprint, secondary_thumbprint])` then the error should say Primary **or** Secondary Thumbprint required with selfSigned auth. I would also double check if you can have just a secondary thumbprint. As in check if the primary thumbprint is minimum required or not.

azext_iot/operations/hub.py
```diff
@@ -246,6 +246,38 @@ def _create_self_signed_cert(subject, valid_days, output_path=None):
...
+    if edge_enabled is not None:
+        instance['capabilities']['iotEdge'] = edge_enabled
+    if status is not None:
+        instance['status'] = status
+    if status_reason is not None:
+        instance['statusReason'] = status_reason
+    if auth_method is not None:
+        if auth_method == DeviceAuthType.shared_private_key.name:
+            auth = 'sas'
+            if (primary_key and not secondary_key) or (not primary_key and secondary_key):
```
digimaun, line 261: Does the service support providing your own SAS key for devices? From what I last recall, you are able to regenerate the primary or secondary key by setting it to an empty value. Let's validate whether you are able to provide a custom key here, and if not when the user changes from some other auth mode to shared private key, this functionality should automatically regenerate keys for them.

If custom shared private key is not possible, then we don't need the `primary_key` and `secondary_key` parameters. Instead we can have a `regenerate_keys` / `--regenerate-keys` where if `auth_method == DeviceAuthType.shared_private_key.name` this function will set both `instance['authentication']['symmetricKey']['primaryKey']=''` and `instance['authentication']['symmetricKey']['secondaryKey'] = ''` which _should_ regenerate them.

azext_iot/operations/hub.py
anusapan: Done

azext_iot/operations/hub.py
anusapan: Yes, we can provide custom keys

digimaun: OK. We should still support regenerating keys easily like in the Portal. 
image

azext_iot/operations/hub.py
digimaun: Double checking - so you can have a secondary thumbprint but not primary? Is that right?

anusapan: Yes

azext_iot/operations/hub.py
anusapan: Shouldn't we create a separate command for renewing the keys for the device-identity like https://docs.microsoft.com/en-us/cli/azure/iot/hub/policy?view=azure-cli-latest#az_iot_hub_policy_renew_key

digimaun: Yea thats fair.

digimaun (approved):

PR #255, opened 2020-09-25
Increase device/module configuration list limit to 1000

c-ryan-k:
Always retrieves the `CONFIG_LIST_UPPER_LIMIT` amount of configs from the service - and adds checks to include configs in final output that have empty objects as `deviceContent` or `modulesContent` (previous check was skipping these as `{}` is falsy)

Also includes some formatting updates in `hub.py`

---

### General Guidelines


c-ryan-k:
Ran all integration tests before fixing config upper limit checks (go me):
image

Config tests are passing but still troubleshooting the failing test in event monitoring:
image

azext_iot/operations/hub.py
```diff
@@ -1172,14 +1193,14 @@ def iot_edge_deployment_list(
...
+    filtered = [c for c in result if c["content"].get("modulesContent") is not None]
     return filtered[:top]
 
 
 def _iot_hub_configuration_list(
-    cmd, hub_name=None, top=10, resource_group_name=None, login=None
+    cmd, hub_name=None, top=None, resource_group_name=None, login=None
 ):
-    top = _process_top(top, upper_limit=100)
+    top = _process_top(top, upper_limit=CONFIG_LIST_UPPER_LIMIT)
```
digimaun, line 1203: Looking at this now, I feel like we shouldn't bother setting a top upper limit. Instead set the default top to 1000. There is cap on configurations (but a soft cap), and would be good to not have to change this client side rule in the future.

digimaun:
Superceded by #257

PR #265, opened 2020-10-26
Nested edge

YingXue:
Update hub service sdk
Implement edge offline (nested edge) feature and update tests

### General Guidelines


azext_iot/sdk/iothub/service/models/job_request_py3.py
```diff
@@ -15,37 +15,35 @@
...
+     standard.
     :type start_time: datetime
-    :param max_execution_time_in_seconds: Max execution time in secounds (ttl
-     duration)
+    :param max_execution_time_in_seconds: The maximum execution time in
+     secounds.
     :type max_execution_time_in_seconds: long
     """
 
-    # @digimaun - altered update_twin type from Twin to {object}
```
digimaun, line 43: Anything that is tagged like this `@digimaun` there was a reason for making the change, so we need to ensure that reason is gone before removing the customization.

azext_iot/sdk/iothub/service/operations/jobs_operations.py
```diff
@@ -529,23 +511,12 @@ def query_jobs(
 
         deserialized = None
 
-        # @digimaun
-        header_dict = {}
-
         if response.status_code == 200:
-            # @digimaun - changed from 'QueryResult' to [object]. Also note link tokens are in the headers.
```
digimaun, line 536: Please ensure these customizations are no longer needed before deleting them.

azext_iot/sdk/iothub/service/iot_hub_gateway_service_ap_is.py
```diff
@@ -85,21 +93,25 @@ def __init__(
         super(IotHubGatewayServiceAPIs, self).__init__(self.config.credentials, self.config)
 
         client_models = {k: v for k, v in models.__dict__.items() if isinstance(v, type)}
-        self.api_version = '2019-10-01'
+        self.api_version = '2020-06-30-preview'
```
digimaun, line 96: There is a mix of API versions? `2020-09-30` and `2020-06-30-preview`? Can we just stick with `2020-09-30`?

azext_iot/sdk/iothub/service/models/job_request_py3.py
YingXue: this is exactly why the tests are failing. I am adding them back, as I didn't realize the hub's operation has impact other than hub.py earlier.

azext_iot/sdk/iothub/service/operations/jobs_operations.py
YingXue: same as above

azext_iot/sdk/iothub/service/iot_hub_gateway_service_ap_is.py
YingXue: my bad. this folder went through api update twice, but the second time i left out the parent folder

azext_iot/_params.py
```diff
@@ -387,17 +387,23 @@ def load_arguments(self, _):
         context.argument(
             "force",
             options_list=["--force", "-f"],
-            help="Overwrites the device's parent device.",
+            help="Overwrites the device's parent device. "
+            "This command parameter has been deprecated and will be removed "
+            "in a future release. Use 'az iot hub device-identity parent set' instead.",
```
digimaun, line 392: Arguments also have the option including a deprecate_info object.

For example:

```python
with self.argument_context('test show-parameters') as c:
  c.argument('junk_flag', help='Something we no longer want to support.' deprecate_info=c.deprecate(expiration='2.1.0'))
```

```python
with self.argument_context('test show-parameters') as c:
  c.argument('resource', options_list=['--resource', c.deprecate(target='--resource-id', redirect='--target')])
```

azext_iot/commands.py
```diff
@@ -50,11 +50,54 @@ def load_command_table(self, _):
...
+        cmd_group.command(
+            "set-parent",
+            "iot_device_set_parent",
+            deprecate_info=self.deprecate(
+                redirect="az iot hub device-identity parent set"
+            ),
+        )
+
+    with self.command_group(
+        "iot hub device-identity children", command_type=iothub_ops
```
digimaun, line 90: We need at least integration tests for these new commands.

azext_iot/commands.py
```diff
@@ -50,11 +50,54 @@ def load_command_table(self, _):
...
+    ) as cmd_group:
+        cmd_group.show_command("add", "iot_device_children_add")
+        cmd_group.show_command("remove", "iot_device_children_remove")
+        cmd_group.show_command("list", "iot_device_children_list")
+
+    with self.command_group(
+        "iot hub device-identity parent", command_type=iothub_ops
+    ) as cmd_group:
+        cmd_group.show_command("get", "iot_device_get_parent")
```
digimaun, line 99: `get` and `set` sound like a nice pair, but overall we want the single resource retrievals to be `show` to align with Az CLI verbs.

digimaun: the `add` and `remove` above are OK because that is adding or removing entities from a collection.

azext_iot/commands.py
digimaun: You could organize the tests in a new module in the tests/iothub directory

azext_iot/_help.py
```diff
@@ -298,48 +298,81 @@
...
+helps[
+    "iot hub device-identity parent set"
+] = """
+    type: command
+    short-summary: Set the parent device of the specified device.
+    examples:
+    - name: Set the parent device of the specified device.
+      text: >
+        az iot hub device-identity parent set -d {device_id} --pd {edge_device_id} -n {iothub_name}
+    - name: Set the parent device of the specified device irrespectively the device is
```
digimaun, line 346: This use of `irrespectively` (which was there before) doesn't sound right. Consider an improvement like:

"Set the parent device of a specified device regardless of if the device already has a parent."

azext_iot/operations/hub.py
```diff
@@ -62,7 +62,7 @@ def iot_query(
 
     try:
         query_args = [query_command]
-        query_method = service_sdk.registry_manager.query_iot_hub
+        query_method = service_sdk.query.get_twins
```
digimaun, line 65: `get_twins` is what the query API is called?

YingXue: yes. i compared the description and param, and verified the output, that's what has been renamed to

azext_iot/operations/hub.py
```diff
@@ -141,18 +141,21 @@ def iot_device_create(
     resolver = SdkResolver(target=target)
     service_sdk = resolver.get_sdk(SdkType.service_sdk)
 
+    if add_children:
+        if edge_enabled:
```
digimaun, line 145: This `if` could be inverted for simplification/less branches. Like:

```python
if add_children:
  if not edge_enabled:
    raise CLIError('The device "{}" should be edge device in order to add children.'.format(device_id))

  for child_device_id in add_children.split(","):
    child_device = _iot_device_show(target, child_device_id.strip())
    _validate_parent_child_relation(child_device, force)
```

digimaun (approved):

PR #280, opened 2020-11-23
Align SSL verification options with core for Central commands

lucadruda:
IoT Central commands don't follow CLI core settings for SSL certificate validation

- Added code to handle "AZURE_CLI_DISABLE_CONNECTION_VERIFICATION" environment variable
- Use of "Session" for request module to share options between requests

---

### General Guidelines

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+0.10.8
++++++++++++++++
+**IoT Central fixes**
+* Align SSL certificate verification with core behavior
+
+
 0.10.7
 +++++++++++++++
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -6,7 +6,7 @@ Release History
 0.10.8
 +++++++++++++++
 **IoT Central fixes**
-* Align SSL certificate verification with core behavior
+* Align SSL certificate verification with core behavior. "AZURE_CLI_DISABLE_CONNECTION_VERIFICATION" environment variable is now honored.
 
 
 0.10.7
--- a/azext_iot/central/services/api_token.py
+++ b/azext_iot/central/services/api_token.py
@@ -18,6 +18,7 @@ BASE_PATH = "api/preview/apiTokens"
 http = requests.Session()
 http.verify = not should_disable_connection_verify()
 
+
 def add_api_token(
     cmd,
     app_id: str,
--- a/azext_iot/central/services/api_token.py
+++ b/azext_iot/central/services/api_token.py
@@ -5,7 +5,7 @@
 # --------------------------------------------------------------------------------------------
 
 
-from requests import Session
+import requests
 from knack.log import get_logger
 from azext_iot.constants import CENTRAL_ENDPOINT
 from azext_iot.central.services import _utility
@@ -15,8 +15,6 @@ from azure.cli.core.util import should_disable_connection_verify
 logger = get_logger(__name__)
 
 BASE_PATH = "api/preview/apiTokens"
-http = Session()
-http.verify = not should_disable_connection_verify()
 
 
 def add_api_token(
@@ -50,7 +48,7 @@ def add_api_token(
 
     headers = _utility.get_headers(token, cmd, has_json_payload=True)
 
-    response = http.put(url, headers=headers, json=payload)
+    response = requests.put(url, headers=headers, json=payload, verify=not should_disable_connection_verify())
     return _utility.try_extract_result(response)
 
 
@@ -74,7 +72,7 @@ def get_api_token_list(
 
     headers = _utility.get_headers(token, cmd)
 
-    response = http.get(url, headers=headers)
+    response = requests.get(url, headers=headers, verify=not should_disable_connection_verify())
     return _utility.try_extract_result(response)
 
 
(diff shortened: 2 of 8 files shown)


azext_iot/central/services/api_token.py
```diff
@@ -10,10 +10,13 @@
 from azext_iot.constants import CENTRAL_ENDPOINT
 from azext_iot.central.services import _utility
 from azext_iot.central.models.enum import Role
+from azure.cli.core.util import should_disable_connection_verify
```
prbans, line 13: nit - remove unused `requests` import

azext_iot/central/services/api_token.py
```diff
@@ -10,10 +10,13 @@
...
 
 logger = get_logger(__name__)
 
 BASE_PATH = "api/preview/apiTokens"
+http = requests.Session()
```
prbans, line 18: Do we need to make similar changes to how the `monitor-events` work? 

https://github.com/Azure/azure-iot-cli-extension/blob/a5e3accbb036511e618114a988ce19f8af22c9db/azext_iot/common/_azure.py#L122

cc: @digimaun

HISTORY.rst
```diff
@@ -3,6 +3,12 @@
 Release History
 ===============
 
+0.10.8
```
prbans, line 6: I think we typically wait for a full release before we put something in the release history - work with @digimaun on when/how to add release notes

azext_iot/central/services/device.py
```diff
@@ -15,10 +15,14 @@
 from azext_iot.central.services import _utility
 from azext_iot.central.models.device import Device
 from azext_iot.central.models.enum import DeviceStatus
+from azure.cli.core.util import should_disable_connection_verify
+
 
 logger = get_logger(__name__)
 
 BASE_PATH = "api/preview/devices"
+http = requests.Session()
```
prbans, line 24: Is there a way to write IT's to validate this behavior?

Is there a way to force a change in the session and verify something breaks?

Is there a way to force the SSL to remain the same and see behavior corrected?

azext_iot/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.10.7"
+VERSION = "0.10.8"
```
prbans, line 10: same version bumping comment - check with @digimaun

azext_iot/central/services/device.py
prbans: request changes on review is for this - all else are nits, feel free to #wontfix them (other than if Paymaun has guidance otherwise)

prbans (changes requested):
Great work! Would be good to back this behavior change with IT's

HISTORY.rst
digimaun: This is OK - we now like to update the release history (if there are user facing, or otherwise notable changes) per PR. It helps to more quickly inform the GitHub release notes.

HISTORY.rst
```diff
@@ -3,6 +3,12 @@
...
++++++++++++++++
+**IoT Central fixes**
+* Align SSL certificate verification with core behavior
```
digimaun, line 9: I would also include the specific configuration value the user can expect to be aligned i.e. the env var `AZURE_CLI_DISABLE_CONNECTION_VERIFICATION` would be honored.

azext_iot/constants.py
digimaun: Yea its a good rule of thumb to check before committing the version in PR. For now we are good with the changes in `0.10.8`.  But keep in mind we could do an out-of-band release (due to the need of an immediate behavior change or fix) before the PR gets merged which would throw this off.

azext_iot/central/services/api_token.py
digimaun: I think we should be consistent and that function should be at play due to the `requests.post` against a central app endpoint. Good catch.

HISTORY.rst
lucadruda: added

azext_iot/common/_azure.py
```diff
@@ -120,7 +120,9 @@ def _find_iot_dps_from_list(all_dps, dps_name):
 
 
 def get_iot_central_tokens(cmd, app_id, token, central_dns_suffix):
-    import requests
+    from requests import post
```
prbans, line 123: how come we arent using `http` here?

prbans, line 123: @digimaun - does this need to be done for the hub monitor events too?

lucadruda (commented):
Added comments

azext_iot/common/_azure.py
lucadruda: "http" is just a variable name. Also there is no reason to use Session here as there's just one call and verification settings can be put in the post options. Session is useful in other files where multiple requests need to share same settings

lucadruda:
Change session back to single execution. The verify option is passed to every requests.
This helps when testing fine grained.
Converting to draft until tests ready.

azext_iot/common/_azure.py
digimaun: I am not aware of a requirement for this functionality in hub monitor-events. So I'm leaning towards no in this PR. We also need to verify `uamqp` supports the functionality of a TLS connection without verification.

digimaun:
Closing due to inactivity.

PR #289, opened 2020-12-11
Add dtdlv2 support

valluriraj:
---

### General Guidelines


azext_iot/central/models/template.py
```diff
@@ -15,30 +15,89 @@ def __init__(self, template: dict):
...
         except:
             raise CLIError("Could not parse iot central device template.")
 
-    def get_schema(self, name, interface_name=""):
-        # interface_name has been specified, do a pointed lookup
-        if interface_name:
-            interface = self.interfaces.get(interface_name, {})
-            return interface.get(name)
+    def get_schema(self, name, is_component=False, identifier=""):
+        if is_component:
```
prbans, line 28: can consolidate a lot of this:

```
to_search = self.components if is_component else self.interfaces
to_return = to_search.get(name) if identifier else self._search(to_search)
```

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,89 @@ def __init__(self, template: dict):
...
+
         except:
             raise CLIError("Could not parse iot central device template.")
 
-    def get_schema(self, name, interface_name=""):
-        # interface_name has been specified, do a pointed lookup
-        if interface_name:
-            interface = self.interfaces.get(interface_name, {})
-            return interface.get(name)
+    def get_schema(self, name, is_component=False, identifier=""):
```
prbans, line 27: please define return types on all functions:

```
def foo(some_obj: dict, some_var: string) -> dict:
```

azext_iot/monitor/property.py
```diff
@@ -134,11 +132,22 @@ def _validate_payload_against_interfaces(
                     device_id=self._device_id,
                     template_id=self._template.id,
                 )
+        else:
+            # Property update is part of a component perform additional validations under component list.
+            component_property_updates = [
+                property_name for property_name in payload if property_name != "__t"
```
prbans, line 138: nit - we should move these things ("__t") to a constants file

azext_iot/tests/central/json/device_template_int_test.json
```diff
@@ -8,6 +8,15 @@
         "@type": [
```
prbans, line 8: does this file also need the "@type": "CapabilityModel" change?

azext_iot/tests/central/json/device_template.json
```diff
@@ -7,9 +7,7 @@
     "displayName": "duplicate-field-name",
     "capabilityModel": {
         "@id": "urn:sampleApp:modelOne_bz:2",
-        "@type": [
-            "CapabilityModel"
-        ],
+        "@type": "CapabilityModel",
```
prbans, line 10: should we move the old ones to something like a /v1 folder, and the new ones to the /v2 folder?

azext_iot/central/_help.py
```diff
@@ -39,8 +39,6 @@ def load_central_help():
     _load_central_monitors_help()
     _load_central_command_help()
     _load_central_compute_device_key()
-    # TODO: Delete this by end of Dec 2020
-    _load_central_deprecated_commands()
```
prbans, line 43: yesssssssss +1

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,89 @@ def __init__(self, template: dict):
...
         except:
             raise CLIError("Could not parse iot central device template.")
 
-    def get_schema(self, name, interface_name=""):
-        # interface_name has been specified, do a pointed lookup
-        if interface_name:
-            interface = self.interfaces.get(interface_name, {})
-            return interface.get(name)
+    def get_schema(self, name, is_component=False, identifier="") -> dict:
+        if is_component:
```
prbans, line 28: nit - this can be consolidated a lot:

```
to_search = this.components if is_component else self.interfaces
return to_search.get(name, None)
```

azext_iot/monitor/property.py
```diff
@@ -90,34 +92,31 @@ def _changed_props(self, prop, metadata, property_name):
         }
         return diff
 
+    def _is_component(self, prop):
+        return type(prop) == dict and prop.get(COMPONENT_MARKER) == "c"
+
     def _validate_payload(self, changes, minimum_severity):
         for value in changes:
-            issues = self._validate_payload_against_interfaces(
+            issues = self._validate_payload_against_interfaces_components(
```
prbans, line 100: i liked your earlier name - does naming this `_validate_payload_against_entities` work?

If not, lets just rename it to components for parity with dtdlv2 and remove all references to interfaces outside of user facing strings and dtdlv1 template internal logic

azext_iot/tests/central/json/property_validation_template.json
```diff
@@ -7,8 +7,129 @@
...
+              "displayName": "component1Telemetry",
+              "name": "component1Telemetry",
+              "schema": "double"
+            }
+          ],
+          "displayName": "Component"
+        }
+      },
+      {
+        "@id": "urn:rigado:RS40_Occupancy_Sensor:addRootPropertyReadOnly:5",
```
prbans, line 118: nit - why are some of these URN's and others DTMI's?

azext_iot/tests/test_iot_central_unit.py
```diff
@@ -392,23 +373,64 @@ def test_validate_properties_severity_level(
...
         assert len(issues) == 0
+
+    @mock.patch("azext_iot.central.services.device_template")
+    @mock.patch("azext_iot.central.services.device")
+    def test_validate_properties_name_miss_under_component(
+        self, mock_device_svc, mock_device_template_svc
+    ):
+
+        # setup
+        mock_device_template_svc.get_device_template.return_value = Template(
```
prbans, line 406: loving all these tests

azext_iot/tests/central/json/property_validation_template.json
digimaun: If central supports only v2, then it should always be dtmi.

HISTORY.rst
```diff
@@ -2,6 +2,13 @@
 
 Release History
 ===============
+0.10.8
++++++++++++++++
+
+**IoT Central updates**
+
+* update parsing template logic to support  DTDLV2 models.
```
digimaun, line 10: 0.10.8 already has changes staged for release, combine the central notes with the existing.

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,89 @@ def __init__(self, template: dict):
             self.name = template.get("displayName")
             self.interfaces = self._extract_interfaces(template)
             self.schema_names = self._extract_schema_names(self.interfaces)
+            self.components = self._extract_components(template)
```
prbans, line 18: I think having all these mangled under the same class might introduce some very difficult to see bugs/create maintenance issues. But depending on time constraints I can see why we did this.

I see 3 options:

1) Keep as is
2) Split template parsing into 2 files (templateV1, templateV2), and have a factory that builds them appropriately
3) Split parsing of templates into individual components and have interface/component/etc level parsers.

I think doing (2) makes sense here, since that would also make deprecation of v1 templates super trivial.

I'll let you file backlog items/think about this further, but I won't block on it.

azext_iot/constants.py
```diff
@@ -44,7 +44,7 @@
 DEVICETWIN_MONITOR_TIME_SEC = 15
 # (Lib name, minimum version (including), maximum version (excluding))
 EVENT_LIB = ("uamqp", "1.2", "1.3")
-CENTRAL_PNP_INTERFACE_PREFIX = "$iotin:"
+COMPONENT_MARKER = "__t"
```
digimaun, line 47: nit - I think we should have a more descriptive name here for COMPONENT_MARKER,  like add PnP and the version other wise its a very generic variable name.

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,89 @@ def __init__(self, template: dict):
...
+            entity for entity in rootContents if entity.get("@type") != "Component"
+        ]
+
+        return {"@id": dcm.get("@id", {}), "schema": {"contents": contents}}
+
     def _extract_interfaces(self, template: dict) -> dict:
         try:
+
+            interfaces = []
             dcm = template.get("capabilityModel", {})
```
digimaun, line 89: capabilityModel is removed in DTDL v2.

azext_iot/tests/central/json/device_template_int_test.json
prbans: I see this got resolved, but I don't see the updates, was this a won't fix? Please only resolve the comment after updating the code, or after providing a #wontfix with a reason

azext_iot/tests/central/json/device_template.json
prbans: Same here

valluriraj: currently we support both v1 and v2 dont see additional value in separating them out as v1 and v2. V2 is a super set of

azext_iot/monitor/property.py
valluriraj: _validate_payload_against_entities  works - components and interfaces co-exist - cannot remove interfaces from the reference here.

azext_iot/tests/central/json/property_validation_template.json
valluriraj: Wont  fix - dtdlv1 and v2 are supported today by public API and will continue to do so till end of june.-

azext_iot/central/models/template.py
valluriraj: option 3 is better since components are type of interfaces and v2 has additive components. Adding separate parsers will make sense if the content is completely different and is exclusive.  Backlog item exists to address this for Jan.

azext_iot/central/models/template.py
valluriraj: wont fix - current public api supports both V1 and V2. Templates that are prior to DTDLV2 service side update will continue to remain as V1 and  newer templates created will be V2 format.

valluriraj:
image
central only tests

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,78 @@ def __init__(self, template: dict):
...
             if schema:
                 return schema
 
         # not found
         return None
 
+    def _extract_components(self, template: dict) -> dict:
+        try:
+            dcm = template.get("capabilityModel", {})
+            if dcm.get("contents"):
```
prbans, line 46: can simplify here - 

```
# doing this is safe without an if check because you are defaulting to an empty dictionary
rootContents = dcm.get("contents", {}) 
components = [
                    entity
                    for entity in rootContents
                    if entity.get("@type") == "Component"
]
```

azext_iot/central/models/template.py
```diff
@@ -15,30 +15,78 @@ def __init__(self, template: dict):
...
+                    }
+
+        except Exception:
+            details = "Unable to extract schema for component from template '{}'.".format(
+                self.id
+            )
+            raise CLIError(details)
+
+    def _extract_root_interface_contents(self, dcm: dict):
+        rootContents = dcm.get("contents", {})
```
prbans, line 67: hmm - we seem to do the same thing in extract components, can we extract getting the contents of a rootComponent into a method? Perhaps call it on construction of the Template?

azext_iot/monitor/parsers/central_parser.py
```diff
@@ -92,17 +92,29 @@ def _perform_dynamic_validations(self, payload: dict):
...
-                self.interface_name, list(template.interfaces.keys())
+        # if component name is not defined then data should be mapped to root/inherited interfaces
+        if not self.component_name:
+            self._validate_payload(
+                payload=payload, template=template, is_component=False
             )
+            return
+
+        if not template.components:
+            # tempate does not have any valid components
```
prbans, line 103: nit - template

azext_iot/monitor/parsers/central_parser.py
```diff
@@ -92,17 +92,29 @@ def _perform_dynamic_validations(self, payload: dict):
...
+            details = strings.invalid_component_name(self.component_name, list())
```
prbans, line 104: We should have a separate error here - one where it tells the user the template doesn't have any components

valluriraj: error mapping is per PM recommendation. Error message will have an empty list indicating that there are no components defined in template.

azext_iot/central/models/template.py
valluriraj: wont fix - for readability it is good to keep the calls as is.  call under _extract_components is for components where as call under _extract_root_interface_contents is to add under interfaces.

prbans (approved):
Looks good! Thanks for putting this hard work in. Good to get the validator back up and running

azext_iot/central/models/template.py
valluriraj: fixed

digimaun (approved):

PR #319, opened 2021-03-11
Extend generate-sas-token to take a device connection string

nanowireUK:
New feature: generate a SAS token based on only device connection string.

This extends the az iot cli to provide the functionality we already have in VS Code and as basic scripts in samples. We can directly generate a SAS token based only on the supplied device connection string without a requirement for any IoT Hub authentication. This is particularly useful for customers wanting to connect devices directly over MQTT to IoT Central since in this case you are not able to get an IoT Hub connection string.

---

### General Guidelines

### DIFF
--- a/azext_iot/_validators.py
+++ b/azext_iot/_validators.py
@@ -17,12 +17,14 @@ def mode2_iot_login_handler(cmd, namespace):
             iot_cmd_type = None
             entity_value = None
 
-            if 'hub_name' in args:
+            if 'device_connection_string' in args:
+                entity_value = 'NA'
+            elif 'hub_name' in args:
                 iot_cmd_type = 'IoT Hub'
                 entity_value = args['hub_name']
             elif 'dps_name' in args:
                 iot_cmd_type = 'DPS'
-                entity_value = args['dps_name']
+                entity_value = args['dps_name']            
 
             if not any([login_value, entity_value]):
                 raise CLIError(error_no_hub_or_login_on_input(iot_cmd_type))
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -1611,6 +1611,7 @@ def iot_get_sas_token(
     cmd,
     hub_name=None,
     device_id=None,
+    device_connection_string=None,
     policy_name="iothubowner",
     key_type="primary",
     duration=3600,
@@ -1639,6 +1640,7 @@ def iot_get_sas_token(
             cmd,
             hub_name,
             device_id,
+            device_connection_string,
             module_id,
             policy_name,
             key_type,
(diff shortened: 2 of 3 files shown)


digimaun:
Thank you for this contribution and bringing up the gap in functionality as we aim for the IoT CLI to be a comprehensive tool.

Will get back to you on this PR...we may want to pivot in a different implementation (that covers the same functionality).

azext_iot/operations/hub.py
```diff
@@ -1611,6 +1611,7 @@ def iot_get_sas_token(
     cmd,
     hub_name=None,
     device_id=None,
+    device_connection_string=None,
```
digimaun, line 1614: Instead of supporting an offline mode just for device connection strings, I'm thinking it would be valuable to support either hub, device or module offline SAS token  generation via generic  `connection_string` corresponding to a new `--connection-string` argument on this command.

azext_iot/_validators.py
```diff
@@ -17,12 +17,14 @@ def mode2_iot_login_handler(cmd, namespace):
...
-            if 'hub_name' in args:
+            if 'device_connection_string' in args:
+                entity_value = 'NA'
+            elif 'hub_name' in args:
                 iot_cmd_type = 'IoT Hub'
                 entity_value = args['hub_name']
             elif 'dps_name' in args:
                 iot_cmd_type = 'DPS'
-                entity_value = args['dps_name']
+                entity_value = args['dps_name']
```
digimaun, line 27: Let's remove the extra spaces.

azext_iot/operations/hub.py
```diff
@@ -1665,16 +1668,30 @@ def _iot_build_sas_token(
         parse_iot_device_module_connection_string,
     )
 
+    uri = None
+    policy = None
+    key = None
+
+    if device_connection_string:
```
digimaun, line 1675: Lets add tests to exercise the scenario

digimaun:
@nanowireUK let us know if you want to continue working on the PR, or we can cover this functionality in a different PR.

Right now my mental model for the feature is we should cover more than device connection strings.

azext_iot/operations/hub.py
nanowireUK: @digimaun absolutely agree. This just happened to cover my use case for IoT Central but agree that there's no reason to restrict to only device connection strings

nanowireUK:
@digimaun happy to keep working on the PR and absolutely support the idea of making the implementation more generic. I'll have a go at that and update the PR.

azext_iot/operations/hub.py
nanowireUK: I've noticed it creates a bunch of regression issues in my az iot cli which I need to look at before I go further with it

digimaun:
Superseded by https://github.com/Azure/azure-iot-cli-extension/pull/375

PR #329, opened 2021-04-16
Added support for track 2 SDKs (azure-mgmt-iothub >= 1.0.0)

c-ryan-k:
---

### General Guidelines


c-ryan-k:
Will need to add an explicit dependency for `packaging` in `setup.py`, but waiting for #328 to get merged so I can clean up the conflicts

setup.py
```diff
@@ -48,7 +48,7 @@
 # though that is installed out of band (managed by the extension)
 # for compatibility reasons.
 
-DEPENDENCIES = ["paho-mqtt==1.5.0", "jsonschema==3.2.0", "setuptools"]
+DEPENDENCIES = ["paho-mqtt==1.5.0", "jsonschema==3.2.0", "setuptools", "packaging"]
```
digimaun, line 51: Do we need setuptools anymore?

azext_iot/tests/iothub/jobs/test_iothub_jobs_int.py
```diff
@@ -182,6 +182,21 @@ def test_jobs(self):
             checks=[self.check("jobId", self.job_ids[2])],
         )
 
+        # Allow time for job to transfer to scheduled state (cannot cancel job in running state)
+        from time import sleep
+        sleep(5)
```
digimaun, line 187: Should we make this longer? Or is 5 second sleep consistently working.

azext_iot/constants.py
```diff
@@ -48,3 +48,6 @@
 
 # Config Key's
 CONFIG_KEY_UAMQP_EXT_VERSION = "uamqp_ext_version"
+
+# Initial Track 2 SDK version
+TRACK_2_SDK_MIN_VERSION = '1.0.0'
```
digimaun, line 53: We should prefix this with `IOTHUB_` or re-organize like creating dictionary that holds IoT Hub constants

azext_iot/common/utility.py
```diff
@@ -438,10 +438,14 @@ def is_iso8601_time(self, to_validate: str) -> bool:
             return False
 
 
-def ensure_min_version(cur_ver, min_ver):
-    from pkg_resources._vendor.packaging import version
+def ensure_min_version(min_ver):
```
digimaun, line 441: It would be nice either through docstrings and/or slightly altering the name that this is to ensure the min version of the iothub mgmt sdk.

azext_iot/iothub/providers/discovery.py
```diff
@@ -60,23 +67,27 @@ def get_policies(self, hub_name: str, rg: str) -> List:
...
         self._initialize_client()
 
-        from azure.mgmt.iothub.models import ErrorDetailsException
+        from azure.core.exceptions import HttpResponseError
 
         if rg:
             try:
                 return self.client.get(resource_group_name=rg, resource_name=hub_name)
-            except ErrorDetailsException:
+            except HttpResponseError:
```
digimaun, line 90: Will this exception work for both SDK types? And our error processing behaves the same?

azext_iot/iothub/providers/discovery.py
```diff
@@ -30,9 +33,9 @@ def _initialize_client(self):
         if not self.client:
             if getattr(self.cmd, "cli_ctx", None):
                 self.client = iot_hub_service_factory(self.cmd.cli_ctx)
+                self.sub_id = get_subscription_id(self.cmd.cli_ctx)
             else:
                 self.client = self.cmd
-            self.sub_id = self.client.config.subscription_id
```
digimaun, line 35: We _may_ hit a regression here since before `sub_id` was extracted when the client was initialized and used to build the `target` dictionary.  Might help determine problems if we change this https://github.com/c-ryan-k/azure-iot-cli-extension/blob/sdk_track_2_support/azext_iot/tests/iothub/test_iothub_discovery_int.py#L102 to assert sub_id is not the "unknown" placeholder value.

azext_iot/iothub/providers/discovery.py
digimaun: Btw python supports an exception tuple...for example `except (ErrorDetailsException, HttpResponseError):`. But one exception type would be preferred.

c-ryan-k: Good catch, this is an issue I was working on yesterday - the `azure.core` exception import will cause an issue with the older CLI (2.3.1) - I'll have an update on this in a bit.

azext_iot/tests/iothub/jobs/test_iothub_jobs_int.py
c-ryan-k: Consistently working for me, but I also haven't seen this issue pop up in the build pipelines, only in local testing - one of my hubs got stuck with a job pending in a `scheduled` state, and couldn't add a new job.

The error was because our test tried to cancel a job before it transitioned from `running` to `scheduled` and this seemed to fix it for me (at least for now)

azext_iot/iothub/providers/discovery.py
c-ryan-k: So just to follow up, this issue is a bit complex.

- We switched from `ErrorDetailsException` to the `azure.core.HttpResponseError` (base error class) as `ErrorDetailsException` no longer exists in our Track 2 SDKs.
- In AZ CLI 2.3.1, however - we don't have the `azure.core` exceptions
- The actual exception this will return is a [ResourceNotFoundError](https://github.com/Azure/azure-sdk-for-python/blob/bfb7119dc7ea11aaf25d0f859f28c249b73312fd/sdk/core/azure-core/azure/core/exceptions.py#L317) which also comes from `azure.core`

Is it pertinent here to catch the specific error type if we're not utilizing the error/exception itself?
I'd like to avoid blanket `except:` statements (or more try/catch imports) but that might be the easiest fix here.

digimaun: Good details...my initial thought was if there was a common base exception like `Exception` we could use but I think the most pragmatic solution is the one you suggest in this specific case having `except:` since we are not using the exception itself.

digimaun (approved):

PR #331, opened 2021-04-19
Add twin delete all and twin relationship delete all

vilit1:
---

### General Guidelines

### DIFF
--- a/azext_iot/digitaltwins/_help.py
+++ b/azext_iot/digitaltwins/_help.py
@@ -540,6 +540,16 @@ def load_digitaltwins_help():
             az dt twin delete -n {instance_or_hostname} --twin-id {twin_id} --etag {etag}
     """
 
+    helps["dt twin delete-all"] = """
+        type: command
+        short-summary: Deletes all digital twins within the specified resource, including all relationships for those twins.
+
+        examples:
+        - name: Delete all digital twins. Any relationships referencing the twins will also be deleted.
+          text: >
+            az dt twin delete-all -n {instance_or_hostname}
+    """
+
     helps["dt twin relationship"] = """
         type: group
         short-summary: Manage and configure the digital twin relationships of a Digital Twins instance.
@@ -643,6 +653,20 @@ def load_digitaltwins_help():
         - name: Delete a digital twin relationship using the etag.
           text: >
             az dt twin relationship delete -n {instance_or_hostname} --twin-id {twin_id} --relationship-id {relationship_id} --etag {etag}
+
+        - name: Delete all digital twin relationships associated with the twin.
+          text: >
+            az dt twin relationship delete -n {instance_or_hostname} --twin-id {twin_id} --all
+    """
+
+    helps["dt twin relationship delete-all"] = """
+        type: command
+        short-summary: Deletes all digital twin relationships on a Digital Twins instance, including incomming relationships.
+
+        examples:
+        - name: Delete all digital twin relationships associated with the twin.
+          text: >
+            az dt twin relationship delete-all -n {instance_or_hostname} --twin-id {twin_id}
     """
 
     helps["dt twin telemetry"] = """
--- a/azext_iot/digitaltwins/_help.py
+++ b/azext_iot/digitaltwins/_help.py
@@ -542,7 +542,7 @@ def load_digitaltwins_help():
 
     helps["dt twin delete-all"] = """
         type: command
-        short-summary: Deletes all digital twins within the specified resource, including all relationships for those twins.
+        short-summary: Deletes all digital twins within a Digital Twins instance, including all relationships for those twins.
 
         examples:
         - name: Delete all digital twins. Any relationships referencing the twins will also be deleted.
@@ -657,12 +657,16 @@ def load_digitaltwins_help():
 
     helps["dt twin relationship delete-all"] = """
         type: command
-        short-summary: Deletes all digital twin relationships on a Digital Twins instance, including incomming relationships.
+        short-summary: Deletes all digital twin relationships within a Digital Twins instance, including incoming relationships.
 
         examples:
         - name: Delete all digital twin relationships associated with the twin.
           text: >
             az dt twin relationship delete-all -n {instance_or_hostname} --twin-id {twin_id}
+
+        - name: Delete all digital twin relationships within the Digital Twins instace.
+          text: >
+            az dt twin relationship delete-all -n {instance_or_hostname}
     """
 
     helps["dt twin telemetry"] = """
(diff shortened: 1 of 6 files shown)


azext_iot/digitaltwins/command_map.py
```diff
@@ -100,6 +100,7 @@ def load_digitaltwins_commands(self, _):
         cmd_group.show_command("show", "show_twin")
         cmd_group.command("update", "update_twin")
         cmd_group.command("delete", "delete_twin")
+        cmd_group.command("delete-all", "delete_all_twin", confirmation=True, supports_no_wait=True)
```
digimaun, line 103: I don't think `supports_no_wait=True` is accurate here as that implies the service is doing work async.

azext_iot/digitaltwins/command_map.py
```diff
@@ -115,6 +116,7 @@ def load_digitaltwins_commands(self, _):
         cmd_group.command("list", "list_relationships")
         cmd_group.command("update", "update_relationship")
         cmd_group.command("delete", "delete_relationship")
+        cmd_group.command("delete-all", "delete_all_relationship", confirmation=True, supports_no_wait=True)
```
digimaun, line 119: See top comment on `supports_no_wait=True`.

azext_iot/digitaltwins/_help.py
```diff
@@ -540,6 +540,16 @@ def load_digitaltwins_help():
             az dt twin delete -n {instance_or_hostname} --twin-id {twin_id} --etag {etag}
     """
 
+    helps["dt twin delete-all"] = """
+        type: command
+        short-summary: Deletes all digital twins within the specified resource, including all relationships for those twins.
```
digimaun, line 545: Instead of saying "within the specified resource", I think the description used in relationship delete-all is a bit more clear where it says "... on/within a Digital Twins instance,"

azext_iot/digitaltwins/_help.py
```diff
@@ -645,6 +655,16 @@ def load_digitaltwins_help():
             az dt twin relationship delete -n {instance_or_hostname} --twin-id {twin_id} --relationship-id {relationship_id} --etag {etag}
     """
 
+    helps["dt twin relationship delete-all"] = """
+        type: command
+        short-summary: Deletes all digital twin relationships on a Digital Twins instance, including incomming relationships.
```
digimaun, line 660: Small typo for `incoming`.

azext_iot/digitaltwins/_help.py
```diff
@@ -645,6 +655,16 @@ def load_digitaltwins_help():
...
+
+        examples:
+        - name: Delete all digital twin relationships associated with the twin.
+          text: >
+            az dt twin relationship delete-all -n {instance_or_hostname} --twin-id {twin_id}
```
digimaun, line 665: This is the only use case supported in the command right? As in you have to always provide a twin id. What about a scenario where I would like to clear all relationships but keep the twins in place? Also the command description sort of implies this is possible with the "Delete all digital twin relationships  on a Digital Twins instance," part. In a nutshell I am wondering if we could make the `--twin-id` optional.

azext_iot/digitaltwins/commands_twins.py
```diff
@@ -44,7 +44,12 @@ def update_twin(cmd, name_or_hostname, twin_id, json_patch, resource_group_name=
 
 def delete_twin(cmd, name_or_hostname, twin_id, resource_group_name=None, etag=None):
     twin_provider = TwinProvider(cmd=cmd, name=name_or_hostname, rg=resource_group_name)
-    return twin_provider.delete(twin_id, etag=etag)
+    return twin_provider.delete(twin_id=twin_id, etag=etag)
+
+
+def delete_all_twin(cmd, name_or_hostname, resource_group_name=None, etag=None):
```
digimaun, line 50: I think usage of etag here is awkward since each twin and relationship has its own etag value. The lower level delete operation would fail on the first mismatched etag.

azext_iot/digitaltwins/commands_twins.py
```diff
@@ -118,6 +123,13 @@ def delete_relationship(
     )
 
 
+def delete_all_relationship(
+    cmd, name_or_hostname, twin_id, resource_group_name=None, etag=None
```
digimaun, line 127: Please see earlier comment on etag for bulk removals.

azext_iot/digitaltwins/providers/twin.py
```diff
@@ -214,6 +230,36 @@ def delete_relationship(self, twin_id, relationship_id, etag=None):
...
+            while True:
+                relationships.extend(incoming_pager.advance_page())
+        except StopIteration:
+            pass
+
+        print(f"Found {len(relationships)} relationship(s) associated with twin {twin_id}.")
+
+        for relationship in relationships:
+            try:
+                if type(relationship) == dict:
```
digimaun, line 248: `isinstance()` is preferred to `type()`. It's faster and reflects inheritance.

azext_iot/digitaltwins/providers/twin.py
```diff
@@ -214,6 +230,36 @@ def delete_relationship(self, twin_id, relationship_id, etag=None):
...
+                    self.delete_relationship(
+                        twin_id=twin_id,
+                        relationship_id=relationship['$relationshipId'],
+                        etag=etag
+                    )
+                else:
```
digimaun, line 254: We should look into optimizing `list_relationships` to always return the same type so we don't need to branch here. End user usage I believe its always a list of dictionaries/list of objects returned.

vilit1: This issue stems from incoming_relationships=True being unpacked by the sdk (and giving a list of twins) vs incoming_relationships=False giving a pager and forcing us to unpack.

I also use is this a dict or twin object as a lazy way to differentiate if I should be using the twin_id or source_id when deleting the relationship. Otherwise I would have two for loops with with almost identical code structure for two lists.

azext_iot/digitaltwins/commands_twins.py
```diff
@@ -118,6 +123,16 @@ def delete_relationship(
     )
 
 
+def delete_all_relationship(
+    cmd, name_or_hostname, twin_id=None, resource_group_name=None
+):
+    twin_provider = TwinProvider(cmd=cmd, name=name_or_hostname, rg=resource_group_name)
+    if twin_id:
+        return twin_provider.delete_all_relationship(twin_id=twin_id)
+    else:
```
digimaun, line 132: nit - this `else` could be eliminated

digimaun (approved):

azext_iot/tests/digitaltwins/test_dt_twin_lifecycle_int.py
```diff
@@ -500,16 +501,181 @@ def test_dt_twin(self):
...
+
+        twin_query_result = self.cmd(
+            "dt twin query -n {} -g {} -q 'select * from digitaltwins' --cost".format(
+                instance_name, self.rg
+            )
+        ).get_output_in_json()
+        assert len(twin_query_result["result"]) == 0
+        assert twin_query_result["cost"]
+
+    def test_dt_twin_delete(self):
```
digimaun, line 519: nit - I think we should label these tests as bulk delete

vilit1:
image

PR #334, opened 2021-04-28
IOT Central multi-hub failover commands

gtrifonov:
---

### General Guidelines


prbans:
Looks like style checks are failing - take a look at this section of the contributing doc

https://github.com/Azure/azure-iot-cli-extension/blob/dev/CONTRIBUTING.md#formatting-and-linting

azext_iot/central/_help.py
```diff
@@ -85,6 +85,35 @@ def _load_central_devices_help():
             --device-id {deviceid}
     """
 
+    helps[
+        "iot central device manual-failover"
+    ] = """
+        type: command
+        short-summary: Execute a manual failover of device across multiple IoT Hubs to validate device firmware's ability to reconnect using DPS to a different IoT Hub.
```
prbans, line 92: Can we include a link to a public doc explaining what a failover/failback is?

azext_iot/central/services/_utility.py
```diff
@@ -42,3 +43,9 @@ def try_extract_result(response: Response):
         raise CLIError(body["error"])
 
     return body
+
+def log_response_debug(response: Response, logger: logging.Logger):
```
prbans, line 47: +1 big thanks

azext_iot/central/services/device.py
```diff
@@ -42,7 +43,7 @@ def get_device(
     url = "https://{}.{}/{}/{}".format(app_id, central_dns_suffix, BASE_PATH, device_id)
     headers = _utility.get_headers(token, cmd)
 
-    response = requests.get(url, headers=headers)
+    response = requests.get(url, headers=headers, verify=not should_disable_connection_verify())
```
prbans, line 46: +1 big thanks

azext_iot/central/services/device.py
```diff
@@ -294,5 +295,75 @@ def get_component_command_history(
...
+    Returns:
+        result (currently a 200)
+    """
+
+    url = "https://{}.{}/{}/{}/manual-failover".format(
+        app_id, central_dns_suffix, "system/iothub/devices", device_id
+    )
+    headers = _utility.get_headers(token, cmd)
+    json = {}
+    if ttl_minutes :
```
prbans, line 332: is TTL 0 valid? because if so this won't get populated on 0 case

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -472,6 +479,80 @@ def test_central_device_registration_summary(self):
         assert json_result[DeviceStatus.blocked.value] is not None
         assert len(json_result) == 4
 
+    def test_central_device_should_start_failover_and_failback(self):
+
+        # created device template & device
```
prbans, line 484: awesome, i love IT's

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -582,8 +665,8 @@ def _create_device_template(self):
         template_id = template_name + "id"
 
         self.cmd(
-            "iot central device-template create --app-id {} --device-template-id {} -k '{}'".format(
-                APP_ID, template_id, device_template_path
+            "iot central device-template create --app-id {} --device-template-id {} -k '{}' --token \"{}\"  --central-dns-suffix \"{}\" ".format(
+                APP_ID, template_id, device_template_path, TOKEN, DNS_SUFFIX
```
prbans, line 669: @digimaun do we need to revert these DNS_SUFFIX/TOKEN params? Or should we update the IT pipeline to include them?

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -647,3 +730,11 @@ def _get_monitor_events_output(self, device_id, enqueued_time, asserts=None):
             output = ""
 
         return output
+
+    def _connect_gettwin_disconnect_wait_tobeprovisioned(self, device_id, credentials):
+        device_client = helpers.dps_connect_device(device_id, credentials)
+        assert device_client.connected
```
prbans, line 736: Don't assert until after the test passes or you won't clean up things in the IT pipeline on errors

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -472,6 +479,80 @@ def test_central_device_registration_summary(self):
...
+            " --device-id {}"
+            " --token \"{}\""
+            " --central-dns-suffix \"{}\""
+            .format(APP_ID, device_id, TOKEN, DNS_SUFFIX)
+        )
+
+        # checking if failover has been done by comparing original hub identifier with hub identifier after failover is done
+        fb_json_result = fb_result.get_output_in_json()
+        hubIdentifierFailOver = fb_json_result["hubIdentifier"]
+        assert len(hubIdentifierFailOver) > 0
```
prbans, line 529: please leave all assertions to the very end of the test, otherwise if there is a failure we won't do any of the cleanup ops

azext_iot/central/params.py
```diff
@@ -150,6 +150,13 @@ def load_central_arguments(self, _):
             " A case-sensitive string (up to 128 characters long) of ASCII 7-bit alphanumeric characters plus"
             " certain special characters: - . + % _ # * ? ! ( ) , : = @ $ '",
         )
+    with self.argument_context("iot central device manual-failover") as context:
+        context.argument(
+            "ttl_minutes",
+            type=int,
```
prbans, line 156: Can we add some validation ranges to the arguments passed here? Right now the user can pass -1

azext_iot/central/providers/device_provider.py
```diff
@@ -225,7 +225,35 @@ def get_component_command_history(
             command_name=command_name,
             central_dns_suffix=central_dns_suffix,
         )
+        
+    def run_manual_failover(
+        self,
+        device_id: str,
+        ttl_minutes:int=None,
```
prbans, line 232: We should have defaulting to the same value we specify in the help text, i.e. this should be 30 not None

azext_iot/central/commands_device.py
```diff
@@ -52,7 +52,9 @@ def delete_device(
     cmd, app_id: str, device_id: str, token=None, central_dns_suffix=CENTRAL_ENDPOINT,
 ):
     provider = CentralDeviceProvider(cmd=cmd, app_id=app_id, token=token)
-    return provider.delete_device(device_id)
+    return provider.delete_device(
+        device_id=device_id,
```
prbans, line 56: awesome, thanks for fixing these bugs

prbans (changes requested):
Looks good overall! Needs a few minor things around docs/user input validation done

azext_iot/tests/central/test_iot_central_int.py
digimaun: I value ease of test configuration & maintaining tests -- if they are not strictly needed I would lean on removing it. Especially if they are not commonly used parameters by end users - it would be good to run command IT's in the most realistic way an end user would. We could have unit tests that cover DNS_SUFFIX & TOKEN plumbing. Open to more thoughts.

azext_iot/central/services/device.py
gtrifonov: TTLL is optional parameter for backend, if not specified than default value will be assigned by backend. default can be updated over time in backend

azext_iot/central/params.py
```diff
@@ -150,6 +150,13 @@ def load_central_arguments(self, _):
...
+            options_list=["--ttl-minutes","--ttl"],
+            help="A positive integer. TTL in minutes to move device back to the original hub. Defaults to 30 minutes if not specified.",
```
prbans, line 158: if this can change async in the backend then remove the defaults to statement or CLI users might get confused

azext_iot/central/commands_device.py
```diff
@@ -88,6 +90,36 @@ def run_command(
...
 
+def run_manual_failover(
+    cmd,
+    app_id: str,
+    device_id: str,
+    ttl_minutes=None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+):
+    if ttl_minutes and ttl_minutes < 1:
```
prbans, line 101: we should have this validation in the `params` layer not in the `controller` layer

azext_iot/tests/central/test_iot_central_int.py
gtrifonov: These parameters have been introduced before and i fixed their consistency usage. domain prefix and token usage is valid for multi cloud configuration such as mooncake, fairfax, blackforest and allows to run integration tests in pre production enviroments

digimaun: But `--token` & `--central-dns-suffix` are not being applied consistently in `test_iot_central_int.py`. For example (arbitrarily choosing) 

https://github.com/Azure/azure-iot-cli-extension/blob/43b50797108368d0815ec2abfd157bb894188977/azext_iot/tests/central/test_iot_central_int.py#L93

https://github.com/Azure/azure-iot-cli-extension/blob/43b50797108368d0815ec2abfd157bb894188977/azext_iot/tests/central/test_iot_central_int.py#L336

https://github.com/Azure/azure-iot-cli-extension/blob/43b50797108368d0815ec2abfd157bb894188977/azext_iot/tests/central/test_iot_central_int.py#L597

Even if they were applied across the board, these parameters are optional. So we wouldn't be testing behavior without them which is probably common (not to include them).

There are several ways we could support this. We can do a whole pass of the central integration tests without them then a pass with them. This could be supported with a fixture addition. Another is to expand unit tests to assert the dns suffix and token plumbing.

azext_iot/central/_help.py
```diff
@@ -85,6 +85,37 @@ def _load_central_devices_help():
...
+        long-summary: For more information about high availability visit https://github.com/iot-for-all/iot-central-high-availability-clients#readme
```
prbans, line 93: While I do love this doc (its hilarious), should we perhaps considering linking to an official azure docs page instead?

This is the official azure CLI, we shouldn't be linking to one off repos

azext_iot/central/params.py
prbans: handled in controller layer

azext_iot/central/params.py
```diff
@@ -150,6 +150,13 @@ def load_central_arguments(self, _):
             " A case-sensitive string (up to 128 characters long) of ASCII 7-bit alphanumeric characters plus"
             " certain special characters: - . + % _ # * ? ! ( ) , : = @ $ '",
         )
+    with self.argument_context("iot central device manual-failover") as context:
+        context.argument(
+            "ttl_minutes",
+            type=int,
+            options_list=["--ttl-minutes", "--ttl"],
+            help="A positive integer. TTL in minutes to move device back to the original hub.",
```
prbans, line 158: We should add information here that there is a default value, even if the client doesn't know about it, we should document it publically or it may customer confusion.

e.g. if a customer doesn't set this, they might expect the failover to never expire

prbans: something like "... Has default value in backend. Client may become out of sync, see docs here on what the latest backend default time to live is aka.ms/link-here"

prbans (approved):
Ok looks good to me now, last few things I'd ask are:

1) Post passing IT's
2) Update the verbiage around 30min TTL, and notify user that they are getting an unknown to client default value if not passed to CLI

azext_iot/central/_help.py
gtrifonov: it's official one which i got from team

gtrifonov:
image

prbans (dismissed):
Awesome!! Good work

digimaun (approved):

PR #336, opened 2021-04-29
Add support for direct method in simulator and integrate IoT Python SDK

avagraw:
---

### General Guidelines


azext_iot/operations/hub.py
```diff
@@ -2187,6 +2189,11 @@ def iot_simulate_device(
     )
     token = None
 
+    if method_response_payload:
```
digimaun, line 2192: We should throw an error if custom values are provided for `method_response_code` and `method_response_payload` when simulating via http so users aren't confused. Similar to what we have when trying to change c2d receive settlement when we are simulating an mqtt protocol device.

digimaun: And a test for this.

azext_iot/operations/_mqtt.py
```diff
@@ -5,92 +5,70 @@
...
+
+        # set response payload
+        if self.method_response_payload:
+            payload = self.method_response_payload
+        else:
+            payload = {
+                "device_id": self.device_id,
+                "method_name": method_request.name,
+                "method_request_id": method_request.request_id,
+                "data": "Method executed successfully"
```
digimaun, line 52: If no custom payload is requested, the returned payload needs to show the request payload

digimaun: Also this needs to be formatted with json casing

digimaun: Example:

```
{
  "methodRequestName": "mymethod",
  "methodRequestId": 12345,
  "methodRequestPayload": { ... }
}
```

I didn't include device Id because I don't think it is that useful. If a target device does not have a method callback setup the `invoke-device-method` command will fail or timeout (which includes the target device you are making a request to anyway).

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -217,6 +254,61 @@ def test_uamqp_device_messaging(self):
             expect_failure=True,
         )
 
+    def test_uamqp_device_direct_method_with_custom_response_status_payload(self):
```
digimaun, line 257: this is not a `uamqp` test

azext_iot/operations/hub.py
avagraw: I think instead of checking for http and failing we should check if the protocol is not mqtt - and then fail if custom values are provided for method_response_code and method_response_payload. That makes us future proof.

azext_iot/operations/_mqtt.py
```diff
@@ -5,92 +5,87 @@
...
+
+        # set response payload
+        if self.method_response_payload:
+            payload = self.method_response_payload
+        else:
+            payload = {
+                "methodName": method_request.name,
+                "methodRequestId": method_request.request_id,
+                "methodRequestPayload": method_request.payload,
+                "result": "Method executed successfully"
```
digimaun, line 64: Why include this static `result` property in the default response?

avagraw:
All integration tests are passing:

image

image

image

image

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -426,14 +514,15 @@ def test_hub_monitor_events(self):
...
+                    None
                 ],
                 max_runs=1,
+                return_handle=True,
             )
+
         # Monitor events for all devices and include sys, anno, app
         self.command_execute_assert(
-            "iot hub monitor-events -n {} -g {} --cg {} --et {} -t 5 -y -p sys anno app".format(
+            "iot hub monitor-events -n {} -g {} --cg {} --et {} -t 7 -y -p sys anno app".format(
```
c-ryan-k, line 525: No need to change or fix, just curious if the results of this test are consistently working with a timeout of `7` but not `5`.  Still better than `30` :)

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -508,33 +597,28 @@ def test_hub_monitor_events(self):
...
-                LIVE_RG,
-                None,
-                1,
-            ],
-            max_runs=1,
+        iot_device_send_message(
+            client,
+            device_ids[i],
+            LIVE_HUB,
+            '{\r\n"payload_data1":"payload_value1"\r\n}',
```
c-ryan-k, line 604: Like mentioned earlier - can we parametrize this?

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -543,31 +627,27 @@ def test_hub_monitor_events(self):
...
-                "$.ct=application/json",
-                1,
-                LIVE_RG,
-            ],
-            max_runs=1,
+        iot_device_send_message(
+            client,
+            device_ids[i],
+            LIVE_HUB,
+            '{\r\n"payload_data1":"payload_value1"\r\n}',
```
c-ryan-k, line 634: And we can reuse the same value here

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -576,23 +656,19 @@ def test_hub_monitor_events(self):
...
-                "$.ct=application/json",
-                1,
-                LIVE_RG,
-            ],
-            max_runs=1,
+        iot_device_send_message(
+            client,
+            device_ids[i],
+            LIVE_HUB,
+            '{\r\n"payload_data1""payload_value1"\r\n}',
```
c-ryan-k, line 663: and here :)

c-ryan-k (commented):
Glad to see we caught the issue regarding timeouts here - I have a few nitpicky-type comments but overall it looks good. Going to validate this against a few different local versions of the azure CLI to ensure we don't run into any issues regarding package versioning / scoping - and of course to do some more performance validation.

azext_iot/operations/_mqtt.py
```diff
@@ -5,92 +5,87 @@
...
+    def send_d2c_message(self, message_text, properties=None):
+        message = Message(message_text)
+        message.custom_properties = properties
+        self.device_client.send_message(message)
+
+    def message_handler(self, message):
         six.print_()
-        six.print_("_Received C2D message with topic_: {}".format(msg.topic))
-        six.print_("_Payload_: {}".format(msg.payload))
+        six.print_("_Received C2D message with topic_: /devices/{}/messages/devicebound".format(self.device_id))
```
digimaun, line 33: All references to `six` should be removed. But we can get to it in the next iteration.

avagraw: Yeah I will address this in the next iteration

azext_iot/_params.py
```diff
@@ -63,7 +63,8 @@
     type=str,
     nargs="?",
     choices=["0", "1"],
-    help="Quality of Service. 0 = At most once, 1 = At least once. 2 (Exactly once) is not supported.",
+    help="Quality of Service. 0 = At most once, 1 = At least once. 2 (Exactly once) is not supported."
+    "This command parameter has been deprecated and will be removed in a future release.",
```
digimaun, line 67: For deprecating we use `deprecate_info` like so

image

And the deprecation message needs to be in a released version prior to the version where the functionality stops working or is removed.

azext_iot/operations/_mqtt.py
```diff
@@ -5,92 +5,87 @@
...
+        self.device_client = mqtt_device_client.create_from_connection_string(device_conn_string)
+        self.device_client.connect()
+        self.device_client.on_message_received = self.message_handler
+        self.device_client.on_method_request_received = self.method_request_handler
+        self.method_response_code = method_response_code
+        self.method_response_payload = method_response_payload
+
+    def send_d2c_message(self, message_text, properties=None):
+        message = Message(message_text)
+        message.custom_properties = properties
```
digimaun, line 28: This handles system properties as well right?

azext_iot/operations/hub.py
```diff
@@ -1806,62 +1804,22 @@ def iot_device_send_message(
...
-            hostname=target["entity"],
-            auth=auth,
-            port=8883,
-            protocol=mqtt.MQTTv311,
-            tls=tls,
-        )
-        return
-    except Exception as x:
-        raise CLIError(x)
+        client_mqtt.send_d2c_message(message_text=data, properties=properties)
```
digimaun, line 1822: This handles system properties right?

azext_iot/_params.py
digimaun: Point being we will need to add this in a different branch

digimaun (approved):

azext_iot/_params.py
avagraw: As discussed, I am removing QoS in this PR. The deprecation will be done in a separate version which will be released before these changes are released

c-ryan-k (approved):

PR #345, opened 2021-05-19
Digital Twin wait commands

vilit1:
---

### General Guidelines


vilit1:
image

azext_iot/digitaltwins/_help.py
```diff
@@ -113,6 +113,19 @@ def load_digitaltwins_help():
...
+        type: command
+        short-summary: Wait until an operation on an Digital Twins instance is complete.
+
+        examples:
+        - name: Wait until an arbitrary instance is created.
+          text: >
+            az dt wait -n {instance_name} --created
+        - name: Wait until an existing instance is deleted.
+          text: >
+            az dt wait -n {instance_name} --deleted
```
c-ryan-k, line 126: Can we add an example for `--custom`? I think the `--created` and `--deleted` are pretty self-explanatory, but an example for `--custom` would be super helpful

azext_iot/tests/digitaltwins/__init__.py
```diff
@@ -160,23 +160,22 @@ def tearDown(self):
...
     def wait_for_hostname(
-        self, instance: dict, wait_in_sec: int = 10, interval: int = 4
+        self, instance: dict, wait_in_sec: int = 10, interval: int = 7
     ):
         from time import sleep
-
-        while interval >= 1:
-            logger.info(
-                "Waiting :{} (sec) for provisioning to complete.".format(wait_in_sec)
+        sleep(wait_in_sec)
```
c-ryan-k, line 166: Just curious - if we're adding the `dt wait` command to our tests - do we need to actually sleep first? In other words, does the `dt wait` command do an immediate poll, or does it wait first before polling? Not a huge deal, just always looking for a way to shave a few seconds off of the tests :)

vilit1: Sadly it polls right away so we can increase the number of polls or let the command sleep first. Since this helper function is used after functions that take a while to complete, it makes sense to sleep first

azext_iot/digitaltwins/_help.py
vilit1: done

c-ryan-k (approved):

digimaun (approved):

PR #353, opened 2021-05-27
Device Twin: Sync reported properties with desired properties during simulation 

avagraw:
---

### General Guidelines

### DIFF
--- a/azext_iot/operations/_mqtt.py
+++ b/azext_iot/operations/_mqtt.py
@@ -76,7 +76,7 @@ class mqtt_client(object):
     def execute(self, data, properties={}, publish_delay=2, msg_count=100):
         from tqdm import tqdm
         try:
-            for _ in tqdm(range(msg_count), desc='Simulation in progress'):
+            for _ in tqdm(range(msg_count), desc='Device simulation in progress'):
                 self.send_d2c_message(message_text=data.generate(True), properties=properties)
                 sleep(publish_delay)
 
--- a/azext_iot/operations/_mqtt.py
+++ b/azext_iot/operations/_mqtt.py
@@ -84,7 +84,7 @@ class mqtt_client(object):
                     desired_properties = device_twin.get("properties").get("desired")
                     reported_properties = device_twin.get("properties").get("reported")
                     twin_properties_to_update = {}
-                    
+
                     for prop in desired_properties:
                         if not prop.startswith("$"):
                             if prop not in reported_properties or desired_properties[prop] != reported_properties[prop]:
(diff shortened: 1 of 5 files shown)


azext_iot/operations/_mqtt.py
```diff
@@ -77,6 +79,20 @@ def execute(self, data, properties={}, publish_delay=2, msg_count=100):
         from tqdm import tqdm
         try:
             for _ in tqdm(range(msg_count), desc='Device simulation in progress'):
+                device_twin = _iot_device_twin_show(self.target, self.device_id)
+                if device_twin:
+                    desired_properties = device_twin.get("properties").get("desired")
```
digimaun, line 84: This double `get()` usage will throw an exception if the first `get()` returns `None`.

azext_iot/operations/_mqtt.py
```diff
@@ -77,6 +79,20 @@ def execute(self, data, properties={}, publish_delay=2, msg_count=100):
...
+                    reported_properties = device_twin.get("properties").get("reported")
+                    twin_properties_to_update = {}
+
+                    for prop in desired_properties:
```
digimaun, line 88: Placeholder

azext_iot/operations/_mqtt.py
digimaun: You can do something like this to avoid the exception `device_twin.get("properties", {}).get("desired", {})`   (note also modified the last `get()` to return an empty `dict` if the key doesn't exist).

digimaun (approved):

avagraw:
image
Integration tests pass

azext_iot/operations/_mqtt.py
avagraw: Will address this in the next PR

azext_iot/operations/_mqtt.py
avagraw: Will address this in the next PR

azext_iot/operations/_mqtt.py
```diff
@@ -77,6 +79,20 @@ def execute(self, data, properties={}, publish_delay=2, msg_count=100):
         from tqdm import tqdm
         try:
             for _ in tqdm(range(msg_count), desc='Device simulation in progress'):
+                device_twin = _iot_device_twin_show(self.target, self.device_id)
```
digimaun, line 82: This should be refactored in a way that is more efficient/the intended usage pattern for the device client to receive desired properties. Its not efficient that for every d2c message we are making an API request to the service for the twin.

PR #374, opened 2021-06-17
Dt create provisioning

vilit1:
---

### General Guidelines

### DIFF
--- a/azext_iot/digitaltwins/providers/resource.py
+++ b/azext_iot/digitaltwins/providers/resource.py
@@ -3,7 +3,6 @@
 # Copyright (c) Microsoft Corporation. All rights reserved.
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
-from msrestazure.polling.arm_polling import ARMPolling
 from azext_iot.digitaltwins.common import (
     ADTEndpointAuthType,
     ADTPublicNetworkAccessType,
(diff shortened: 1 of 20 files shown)


azext_iot/digitaltwins/providers/resource.py
```diff
@@ -76,11 +77,17 @@ def check_state(lro):
...
+                retries = 0
+                while state == "Provisioning" and retries < MAX_ADT_CREATE_RETRIES:
+                    retries += 1
                     sleep(1)
                     lro.update_status()
                     instance = lro.resource().as_dict()
                     state = instance.get('provisioning_state', None)
+                if retries == MAX_ADT_CREATE_RETRIES:
+                    raise CLIError(
+                        "Never finished provisioning the Digital Twins instance. Please try again with a different name."
```
c-ryan-k, line 89: Just a note - this doesn't necessarily mean that they'll need to create another instance - I think we should just warn users that their resource was created (since the poller returned success / `201`), but that it may not be finished provisioning. Perhaps we could even give them a hand with an example `az dt show -n {name} -g {group}` example command.

azext_iot/tests/digitaltwins/__init__.py
```diff
@@ -166,7 +166,7 @@ def wait_for_hostname(
         sleep(wait_in_sec)
 
         self.embedded_cli.invoke(
-            "dt wait -n {} -g {} --custom \"hostName && provisioningState=='Succeeded'\" --interval {} --timeout {}".format(
+            "dt wait -n {} -g {} --created --interval {} --timeout {}".format(
```
c-ryan-k, line 169: Just to jog my memory - can you explain the difference here? Does `--created` return when the poller does (possibly still `provisioning`) or does it wait for our create method to finish?

vilit1: wait --created checks if the get call returns a provisioningState that is not Provisioning. Thus, this will depend on the digital twins get calls and if that returned has a provisioningState of completed. This is assuming that the get call will only return a digital twins object that has provisioningState == Succeeded if there is a hostname.

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -64,13 +65,33 @@ def create(
...
+                while state == "Provisioning" and retries < MAX_ADT_CREATE_RETRIES:
+                    retries += 1
+                    sleep(1)
+                    lro.update_status()
+                    instance = lro.resource().as_dict()
+                    state = instance.get('provisioning_state', None)
+                if retries == MAX_ADT_CREATE_RETRIES:
+                    raise CLIError(
+                        "The resource has been created and has not finished provisioning. Please monitor the status of "
+                        "the Digital Twin instance using az dt show -n {} -g {}".format(
```
c-ryan-k, line 90: Just a minor suggestion to help make the command stand out from the error
```suggestion
                        "the Digital Twins instance using: 'az dt show -n {} -g {}'".format(
```

c-ryan-k (commented):
This looks good to me (besides my last minor string change), just want to make sure - have we tested integration tests for DT locally, and have we ensured that RBAC gets handled appropriately after the `provisioning_state` issue we're handling here?

PR #377, opened 2021-06-22
Feature to support setting twin reported properties during simulation

avagraw:
---

### General Guidelines


avagraw:
Integration tests pass:

image

avagraw:
Integration tests pass, except a couple:

image

Failed integration tests also pass:

image
image
image

c-ryan-k (approved):

azext_iot/operations/_mqtt.py
```diff
@@ -105,8 +109,19 @@ def twin_patch_handler(self, patch):
...
+
         try:
+            if self.init_reported_properties:
+                twin_properties = {
+                    "properties": {
+                        "desired": self.init_reported_properties
+                    }
+                }
+
+                _iot_device_twin_update(self.target, self.device_id, twin_properties)
```
digimaun, line 123: The goal of this functionality is to be able to init device side (reported) properties. The implementation here is setting service side desired properties.

avagraw: Update to the desired properties is picked up by the MQTT listener, which automatically updates the reported properties, hence bringing the desired and reported properties in sync.

digimaun: But you don't **need** to set desired properties to apply the initial reported state. Devices connecting to IoT Hub will provide their initial state through reported properties without needing to make a modification to desired. Desired twin properties are a solution side interaction and its strange that we would force an update to the desired twin state in order to update the reported side when using facilities of a device simulation command. The primary justification that I see right now is implementation convenience.

You can take a [look at this device sample](https://github.com/Azure/azure-iot-sdk-python/blob/master/azure-iot-device/samples/pnp/temp_controller_with_thermostats.py) as an example/datapoint - the device connects to iothub initializing a set of reported properties (reflecting a pnp device) with no modification to the desired/solution side interaction.

PR #378, opened 2021-06-23
Update Uamqp package dependency

vilit1:
---

### General Guidelines


azext_iot/common/deps.py
```diff
@@ -10,14 +10,20 @@
...
 from azext_iot.common._homebrew_patch import HomebrewPipPatch
 
 
 def ensure_uamqp(config, yes=False, repair=False):
-    if get_uamqp_ext_version(config) != EVENT_LIB[1] or repair or not test_import(EVENT_LIB[0]):
+    if (
+        repair or
+        (
+            get_uamqp_ext_version(config) != EVENT_LIB[1] and
+            not test_import_and_version(EVENT_LIB[0], EVENT_LIB[1])
```
c-ryan-k, line 24: Just running through scenarios - this catches a previous situation that the other didn't - in which the dependency was installed (preinstalled, even) but the config value was not set since the command hadn't been run.

However, in this new logic - users can have any version of `uamqp` pre-installed, but we won't write a config value for them unless it's repaired or our extension installs it for them. At this point, do we need the config value? Or would it be safe to assume that if they have a satisfactory version preinstalled, we can just nix the config check?

Maybe it's worthwhile to keep around just to know if our extension was the one who put `uamqp` there...

@digimaun  any thoughts here?

vilit1: To add on (note that has uamqp means has the correct uamqp version):

- I updated the test_import (which is only used in this function) to check for the installed uamqp version too
- Scenario diagram because I confuse myself at times:
image

The issue becomes the left bottom box and if we check for that or have our cli say that is ok. If we say that is not ok then we are just ignoring the azure config (I guess we can add the uamqp value to the azure config if it is not there already). Maybe the left bottom box should be rare and if that box's condition is met, customers should have issues elsewhere (or did not activate the python env) if their config is out of sync with installed packages.

The more I think about this the more I want to agree with Ryan and say let's just ignore the azure config (unless it is a requirement - in which case write to the config value).

digimaun: Yes, lets remove config integration for the `uamqp` install flow and delete the unnecessary code.

PR #388, opened 2021-07-01
Enable integration tests to run in parallel

avagraw:
---

### General Guidelines


azext_iot/tests/iothub/test_iothub_discovery_int.py
```diff
@@ -82,13 +83,13 @@ def test_iothub_targets(self):
         )
         assert_target(desired_target, rg=LIVE_RG, include_events=True)
 
-        sub_targets = discovery.get_targets()
```
digimaun, line 85: Why were these assertions commented out?

azext_iot/operations/_mqtt.py
```diff
@@ -109,18 +109,11 @@ def twin_patch_handler(self, patch):
...
         try:
             if self.init_reported_properties:
-                twin_properties = {
-                    "properties": {
-                        "desired": self.init_reported_properties
-                    }
-                }
-
-                _iot_device_twin_update(self.target, self.device_id, twin_properties)
+                self.device_client.patch_twin_reported_properties(self.init_reported_properties)
```
digimaun, line 116: Are there IT's for this functionality? If not, we need them.

azext_iot/tests/iothub/test_iot_ext_int.py
```diff
@@ -22,7 +23,7 @@
     req_env_set=ENV_SET_TEST_IOTHUB_BASIC, opt_env_set=opt_env_set
 )
 
-LIVE_HUB = settings.env.azext_iot_testhub
+LIVE_HUB = settings.env.azext_iot_testhub if settings.env.azext_iot_testhub else "test-hub-" + str(uuid4())
```
digimaun, line 26: We have a number of tests (most if not all in this module) that require the hub to be configured with a storage account. For new hubs that are created per test, the configuration with storage needs to be done as part of test setup.

azext_iot/tests/iothub/test_iot_messaging_int.py
```diff
@@ -18,7 +18,7 @@
 )
 
 settings = DynamoSettings(ENV_SET_TEST_IOTHUB_BASIC)
-LIVE_HUB = settings.env.azext_iot_testhub
+LIVE_HUB = settings.env.azext_iot_testhub if settings.env.azext_iot_testhub else "test-hub-" + str(uuid4())
```
digimaun, line 21: Instead of `str(uuid4())` you can use `...tests.generators.generate_generic_id()` which we are trying to maintain usage on when this respective need arises.

azext_iot/operations/_mqtt.py
avagraw: Yes, the integration test already exists.

azext_iot/tests/iothub/test_iot_ext_int.py
avagraw: My thought was that why not use a fixed shared storage account, instead of creating a separate one. Do you see any downsides of that

azext_iot/tests/iothub/test_iothub_discovery_int.py
avagraw: I was testing something, this got in by mistake, will fix.

azext_iot/tests/iothub/test_iot_messaging_int.py
avagraw: Makes sense, will make the change

azext_iot/tests/iothub/test_iot_ext_int.py
digimaun: I'm OK with that but the fixed storage account still needs to be configured to work with a new IoT Hub.

avagraw:
All comments addressed

Will soon post screenshots for integration testing evidence as well as time comparisons when running tests parallelly

.azure-devops/templates/set-testenv-sentinel.yml
```diff
@@ -15,7 +15,9 @@ steps:
             "azext_iot_testhub": os.environ.get("AZEXT_IOT_TESTHUB", sentinel_value),
             "azext_iot_testdps": os.environ.get("AZEXT_IOT_TESTDPS", sentinel_value),
             "azext_iot_teststorageuri": os.environ.get("AZEXT_IOT_TESTSTORAGEURI", sentinel_value),
-            "azext_iot_identity_teststorageid": os.environ.get("AZEXT_IOT_IDENTITY_TESTSTORAGEID", sentinel_value),
+            "azext_iot_identity_teststorageid": os.environ.get("AZEXT_IOT_IDENTITY_TESTSTORAGEID", sentinel_value),           
+            "azext_iot_teststorageconnstring": os.environ.get("AZEXT_IOT_TESTSTORAGECONNSTRING", sentinel_value),
```
digimaun, line 19: Can we look into auto-deriving the storage connection string? Looking to have as simple of test configuration as possible, with the least amount of user required info we can get by with.

avagraw: Auto deriving the connection string requires both storage account name as well as storage account key. Hence, instead of taking 2 inputs (storage account name and key) it is better to take only connection string as input.

The SAS URL can be generated using the connection string and container name, hence that input will no longer be needed

digimaun: > Auto deriving the connection string requires both storage account name as well as storage account key.

This is definitely not the case. It is common for resource providers to have/require "get keys" operations which can be used for show connection-string commands.

Storage supports this via the following command:

`az storage account show-connection-string` 

And it supports minimum identifiers...for example with just the storage account name.

```
user@machine:~$ az storage account show-connection-string --name mystorageaccount
{
  "connectionString": "DefaultEndpointsProtocol=https;EndpointSuffix=core.windows.net;AccountName=mystorageaccount;AccountKey=abcd12345"
}
```

.azure-devops/templates/set-testenv-sentinel.yml
```diff
@@ -41,6 +43,8 @@ steps:
       AZEXT_IOT_TESTDPS: $(azext_iot_testdps)
       AZEXT_IOT_TESTSTORAGEURI: $(azext_iot_teststorageuri)
```
digimaun, line 44: You said

> the SAS URL can be generated using the connection string and container name

So why do we still have code that relies on `azext_iot_teststorageuri`?

azext_iot/tests/__init__.py
```diff
@@ -81,6 +84,14 @@ def __init__(self, test_scenario, entity_name, entity_rg):
 
         super(IoTLiveScenarioTest, self).__init__(test_scenario)
 
+        if self.entity_name != settings.env.azext_iot_testhub:
+            self.cmd(
+                "iot hub create --name {} --resource-group {} --fc {} --fcs {} --sku S1 ".format(
```
digimaun, line 89: This is forcing every new iothub created to integrate with a storage account, even for tests that don't have any dependency on storage. Ideally the integration with storage should happen on the storage related tests/tests that have a dependency on iothub configuration with storage.

azext_iot/tests/iothub/devices/test_iothub_device_tracing.py
```diff
@@ -9,10 +9,11 @@
 from azext_iot.tests import IoTLiveScenarioTest
 from azext_iot.tests.settings import DynamoSettings, ENV_SET_TEST_IOTHUB_BASIC
 from azext_iot.common.shared import AuthenticationTypeDataplane
+from azext_iot.tests.generators import generate_generic_id
 
 settings = DynamoSettings(req_env_set=ENV_SET_TEST_IOTHUB_BASIC)
 
-LIVE_HUB = settings.env.azext_iot_testhub
+LIVE_HUB = settings.env.azext_iot_testhub if settings.env.azext_iot_testhub else "test-hub-" + generate_generic_id()
```
digimaun, line 16: Instead of repeating this code for iot hub name generation across the IoT Hub IT modules, we could apply a similar pattern once (reduce code duplication) in the base test class like `IoTLiveScenarioTest` for IoT Hub ITs. First `azext_iot_testhub` must be an optional parameter with respect to its usage with `DynamoSettings`. Then during instance initialization the logic can check if an entity name has been set, if not generate one.  If generated, then create the hub. Similarly during teardown/cleanup process only delete the hub if its generated.

setup.py
```diff
@@ -48,7 +48,7 @@
 # though that is installed out of band (managed by the extension)
 # for compatibility reasons.
 
-DEPENDENCIES = ["jsonschema==3.2.0", "packaging", "azure-iot-device~=2.5", "tqdm"]
+DEPENDENCIES = ["jsonschema==3.2.0", "packaging", "azure-iot-device~=2.5", "tqdm", "azure-storage-blob"]
```
digimaun, line 51: Do we need to take a dependency on `azure-storage-blob` ?

azext_iot/tests/iothub/devices/test_iothub_device_tracing.py
```diff
@@ -42,22 +35,24 @@ def test_iothub_device_distributed_tracing(self):
...
+                    f"iot hub distributed-tracing show -d {device_ids[0]} -n {self.entity_name} -g {self.entity_rg}",
                     auth_type=auth_phase,
                 ),
                 checks=self.is_empty(),
             )
 
             result = self.cmd(
                 self.set_cmd_auth_type(
-                    f"iot hub distributed-tracing update -d {device_ids[0]} -n {LIVE_HUB} -g {LIVE_RG} --sm on --sr 50",
+                    "iot hub distributed-tracing update -d {} -n {} -g {} --sm on --sr 50".format(
```
digimaun, line 53: I'm curious why we have to refactor out of f strings here.

azext_iot/tests/iothub/devices/test_iothub_device_twin_int.py
```diff
@@ -172,7 +167,9 @@ def test_iothub_device_twin_replace(self):
             )
             self.cmd(
                 self.set_cmd_auth_type(
-                    f"iot hub device-twin replace -d {device_ids[0]} -n {LIVE_HUB} -g {LIVE_RG} -j '{replace_twin_content_path}'",
+                    "iot hub device-twin replace -d {} -n {} -g {} -j '{}'".format(
```
digimaun, line 170: Related to refactoring out of f strings question.

azext_iot/tests/__init__.py
```diff
@@ -71,16 +75,37 @@ def command_execute_assert(self, command, asserts):
...
-        assert entity_name
-        assert entity_rg
-
-        self.entity_name = entity_name
-        self.entity_rg = entity_rg
+        self.entity_rg = settings.env.azext_iot_testrg
 
         super(IoTLiveScenarioTest, self).__init__(test_scenario)
 
+        if settings.env.azext_iot_testhub:
```
digimaun, line 84: Solid refactorings from the prior iteration

azext_iot/tests/iothub/configurations/test_iot_config_int.py
```diff
@@ -10,13 +10,8 @@
 from azext_iot.tests import IoTLiveScenarioTest
 from azext_iot.tests.conftest import get_context_path
 from azext_iot.tests.iothub import DATAPLANE_AUTH_TYPES
-from azext_iot.tests.settings import DynamoSettings, ENV_SET_TEST_IOTHUB_BASIC
 from azext_iot.common.utility import read_file_content
 
-settings = DynamoSettings(ENV_SET_TEST_IOTHUB_BASIC)
```
digimaun, line 16: Glad we can reduce boiler plate code across test modules

setup.py
digimaun: I see it is being used for generating a sas uri in tests so it should be limited to `dev_requirements` rather than a general dependency addition

azext_iot/tests/iothub/test_iot_ext_int.py
```diff
@@ -40,8 +34,39 @@
...
-        super(TestIoTStorage, self).__init__(test_case, LIVE_HUB, LIVE_RG)
+
+        super(TestIoTStorage, self).__init__(test_case)
         self.managed_identity = None
+        if LIVE_STORAGE_ACCOUNT:
+            self.live_storage_uri = self.get_container_sas_url()
+
+    def get_container_sas_url(self):
+        from datetime import datetime, timedelta
+        from azure.storage.blob import ResourceTypes, AccountSasPermissions, generate_account_sas, BlobServiceClient
```
digimaun, line 45: Was there no combination of commands in the CLI able to obtain this value? For example `az storage container generate-sas`

avagraw: The SAS token generated using that command was not letting me to run the device-identity export command:

image

Hence, had to use this

setup.py
avagraw: Makes sense, will stick to dev_requirements

azext_iot/tests/iothub/devices/test_iothub_device_tracing.py
avagraw: The line of code was exceeding 130 characters, but we can have multiline f strings

avagraw: Should I change these back to f strings?

avagraw:
I had to merge dev branch because style check was failing. 
My changes are completely independent of the changes coming from dev ( No files are common), in case this is difficult to review I can make 2 PRs

azext_iot/tests/iothub/devices/test_iothub_device_tracing.py
digimaun: Please

avagraw:
When run serially, the IoT Hub tests take ~37 minutes to complete:

image


When run in parallel the longest running test takes ~12 minutes to finish:

image

image

image

digimaun:
We are looking for the tests to actually be run in parallel, rather than one at a time as your screenshots show. This would reveal naming conflicts should they exist. Also please do not skip the storage identity tests. There are an assortment of updates to the tests and we need to ensure _everything_ works.

azext_iot/tests/__init__.py
```diff
@@ -71,16 +86,57 @@ def command_execute_assert(self, command, asserts):
...
+
+                profile = Profile(cli_ctx=DummyCli())
+                user_id = profile.get_current_account_user()
+                new_hub = self.cmd(
+                    "iot hub show -n {} -g {}".format(self.entity_name, self.entity_rg)
+                ).get_output_in_json()
+
+                # assign IoT Hub Data Contributor role to current user
+                self.cmd(
+                    '''role assignment create --assignee {} --role "{}" --scope "{}"'''.format(
```
digimaun, line 135: Role assignments commonly do not propagate immediately. We need to use any combination of wait timers and/or refresh account credentials. We commonly run into RBAC propagation issues on the DT tests side and are looking to improve the RBAC pattern there.

avagraw:
> We are looking for the tests to actually be run in parallel, rather than one at a time as your screenshots show. This would reveal naming or resource conflicts should they exist. Also please do not skip the storage identity tests. There are an assortment of updates to the tests and we need to ensure _everything_ works.

The tests were run in parallel across multiple powershell windows.

azext_iot/tests/__init__.py
avagraw: Makes sense, although I didn't run into issues during my testing. But, it's better to be safe

avagraw:
> > We are looking for the tests to actually be run in parallel, rather than one at a time as your screenshots show. This would reveal naming or resource conflicts should they exist. Also please do not skip the storage identity tests. There are an assortment of updates to the tests and we need to ensure _everything_ works.
> 
> The tests were run in parallel across multiple powershell windows.

In the Azure DevOps pipeline this will be achieved by running a script which kicks off the pytest commands in parallel ( where each pytest command kicks of a different test suite)

digimaun:
> > > We are looking for the tests to actually be run in parallel, rather than one at a time as your screenshots show. This would reveal naming or resource conflicts should they exist. Also please do not skip the storage identity tests. There are an assortment of updates to the tests and we need to ensure _everything_ works.
> > 
> > 
> > The tests were run in parallel across multiple powershell windows.
> 
> In the Azure DevOps pipeline this will be achieved by running a script which kicks off the pytest commands in parallel ( where each pytest command kicks of a different test suite)

Are you suggesting having to run a separate agent job per hub test type? This parallelization would primarily benefit the devops pipeline (at the cost of agents) as trying to execute pytest across local shell windows is not optimal devEx. Why not try to achieve parallelization  which can also benefit local execution and not require numerous jobs?

azext_iot/tests/iothub/test_iothub_discovery_int.py
```diff
@@ -71,24 +67,24 @@ def test_iothub_targets(self):
...
-            hub_name=LIVE_HUB, policy_name=self.desired_policy_name, include_events=True
+            hub_name=self.entity_name, policy_name=self.desired_policy_name, include_events=True
         )
-        assert_target(desired_target, rg=LIVE_RG, include_events=True)
+        assert_target(desired_target, rg=self.entity_rg, include_events=True)
 
-        sub_targets = discovery.get_targets()
-        [assert_target(tar) for tar in sub_targets]
+        # sub_targets = discovery.get_targets()
+        # [assert_target(tar) for tar in sub_targets]
```
digimaun, line 82: Assertion still commented out?

azext_iot/tests/iothub/test_iothub_discovery_int.py
```diff
@@ -71,24 +67,24 @@ def test_iothub_targets(self):
...
 
-        rg_targets = discovery.get_targets(resource_group_name=LIVE_RG, include_events=True)
-        [assert_target(tar, rg=LIVE_RG, include_events=True) for tar in rg_targets]
+        rg_targets = discovery.get_targets(resource_group_name=self.entity_rg, include_events=True)
+        [assert_target(tar, rg=self.entity_rg, include_events=True) for tar in rg_targets]
 
-        assert len(rg_targets) <= len(sub_targets)
+        # assert len(rg_targets) <= len(sub_targets)
```
digimaun, line 87: Assertion still commented out?

digimaun:
> > > > We are looking for the tests to actually be run in parallel, rather than one at a time as your screenshots show. This would reveal naming or resource conflicts should they exist. Also please do not skip the storage identity tests. There are an assortment of updates to the tests and we need to ensure _everything_ works.
> > > 
> > > 
> > > The tests were run in parallel across multiple powershell windows.
> > 
> > 
> > In the Azure DevOps pipeline this will be achieved by running a script which kicks off the pytest commands in parallel ( where each pytest command kicks of a different test suite)
> 
> Are you suggesting having to run a separate agent job per hub test type? This parallelization would primarily benefit the devops pipeline (at the cost of agents) as trying to execute pytest across local shell windows is not optimal devEx. Why not try to achieve parallelization which can also benefit local execution and not require numerous jobs?

We talked about this and next steps offline.

avagraw:
The following screenshots show tests running in parallel using 2 agents (poweshell processes running in parallel). There are flags we can setup to re-run failed tests and set time between re-runs ( the following commands were run with those flags). The re-run was not needed as all tests passed in first go here, but during my testing I was able to test the re-runs as well.

image

image

azext_iot/tests/__init__.py
avagraw: I am able to address this by refreshing accounts

dev_requirements
```diff
@@ -9,4 +9,5 @@ black
 wheel==0.30.0
 pre-commit
 pylint
-flake8
\ No newline at end of file
+flake8
+azure-storage-blob
```
digimaun, line 13: Where is the xdist plugin defined?

azext_iot/tests/__init__.py
```diff
@@ -71,16 +88,59 @@ def command_execute_assert(self, command, asserts):
...
+                        user_id, USER_ROLE, new_hub["id"]
+                    )
+                )
+                profile.refresh_accounts()
+                time.sleep(ROLE_ASSIGNMENT_REFRESH_TIME)
```
digimaun, line 142: Were you able to test if the role assignment sleep step is still needed after the `profile.refresh_accounts()`? Looking for optimization on the ADT side where we have sleep timers in place (but no `profile.refresh_accounts()`).

avagraw: Yes, there is still some time needed for the account to be refreshed, I tried without sleep and saw failures because permissions didn't flow through.

dev_requirements
avagraw: Yeah x-dist should be added here, will make the update

digimaun (dismissed):

azext_iot/tests/__init__.py
```diff
@@ -71,16 +88,59 @@ def command_execute_assert(self, command, asserts):
...
+                        )
+                    ).get_output_in_json()
+
+                    self.cmd(
+                        "iot hub create --name {} --resource-group {} --fc {} --fcs {} --sku S1 ".format(
+                            self.entity_name, self.entity_rg,
+                            STORAGE_CONTAINER, storage_account_connenction["connectionString"]
+                        )
+                    )
+                else:
```
digimaun, line 122: Could conditionally add the storage parameter values to the same `iot hub create` call rather than two `iot hub create` calls differing on whether to include storage elements or not.

azext_iot/tests/dps/test_iot_dps_int.py
```diff
@@ -101,6 +102,13 @@ def __init__(self, test_method):
         self.cmd_shell = Setting()
         setattr(self.cmd_shell, "cli_ctx", self.cli_ctx)
 
+        if hub != os.environ.get("azext_iot_testhub"):
```
digimaun, line 105: IoT Hub tests uses `DynamoSettings` and the same could be used here to support uniform test plumbing across technology areas.

digimaun (approved):

azext_iot/tests/__init__.py
avagraw: Yeah, I just thought the code will be cleaner that way

PR #390, opened 2021-07-12
Fill gaps with IoT Central REST apis

lucadruda:
This adds device groups and role listing to IoT Central preview commands in order to align with REST api offer

### General Guidelines


azext_iot/tests/central/test_iot_central_unit.py
```diff
@@ -19,6 +19,10 @@
     CentralDeviceProviderV1,
     CentralDeviceTemplateProviderV1,
 )
```
valluriraj, line 21: add int tests to the collection.

valluriraj (changes requested):
add integration tests to cover E2E scenarios

valluriraj (approved):

c-ryan-k (approved):

PR #404, opened 2021-08-09
Add missing IoT Central list commands

lucadruda:
This adds device and device template listing to IoT Central commands in order to align with REST api offer.

- List devices
- List templates
- Updated "device-template create" help

### General Guidelines

image

### DIFF
--- a/azext_iot/central/commands_device_template.py
+++ b/azext_iot/central/commands_device_template.py
@@ -50,7 +50,7 @@ def list_device_templates(
         provider = CentralDeviceTemplateProviderV1(cmd=cmd, app_id=app_id, token=token)
 
     templates = provider.list_device_templates(central_dns_suffix=central_dns_suffix)
-    return {template.id: template.raw_template for template in templates.values()}
+    return templates
 
 
 def map_device_templates(
--- a/azext_iot/central/models/templatepreview.py
+++ b/azext_iot/central/models/templatepreview.py
@@ -85,7 +85,7 @@ class TemplatePreview:
                 interfaces.extend(dcm.get("implements"))
 
             return {
-                (interface["schema"]["@id"] if interface.get("@type") else interface["@id"]): self._extract_schemas(interface)
+                self._get_interface_id(interface): self._extract_schemas(interface)
                 for interface in interfaces
             }
         except:
@@ -113,3 +113,6 @@ class TemplatePreview:
             for interface, schema in self.schema_names.items()
             if property_name in schema
         ]
+
+    def _get_interface_id(self, interface) -> list:
+        return interface["schema"]["@id"] if interface.get("@type") else interface["@id"]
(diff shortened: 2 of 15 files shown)


azext_iot/central/services/device.py
```diff
@@ -114,7 +118,7 @@ def list_devices(
                 central_models.DeviceV1(device) for device in result["value"]
```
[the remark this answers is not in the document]
lucadruda: there is no conversion as far as I know. The object gets constructed based on the api-version specified (v1 or preview)

azext_iot/central/commands_device_template.py
```diff
@@ -50,7 +50,7 @@ def list_device_templates(
         provider = CentralDeviceTemplateProviderV1(cmd=cmd, app_id=app_id, token=token)
 
     templates = provider.list_device_templates(central_dns_suffix=central_dns_suffix)
-    return {template.id: template.raw_template for template in templates.values()}
+    return templates
```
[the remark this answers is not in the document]
lucadruda: nope. I just moved the dict generation to providers to align with other "list" commands in the extension

yang-hai-feng (approved):

azext_iot/central/models/templatepreview.py
```diff
@@ -77,18 +77,18 @@ def _extract_interfaces(self, template: dict) -> dict:
...
+            
             if dcm.get("contents"):
                 interfaces.append(self._extract_root_interface_contents(dcm))
 
             if dcm.get("implements"):
                 interfaces.extend(dcm.get("implements"))
 
             return {
-                interface["@id"]: self._extract_schemas(interface)
+                (interface["schema"]["@id"] if interface.get("@type") else interface["@id"]): self._extract_schemas(interface)
```
vilit1, line 88: Can you move the code to determine and get the interface id into a separate helper function? It will help with readability.

azext_iot/central/services/device.py
vilit1, line 118: use devices.extend([devices to add])

vilit1:
Can you post a screenshot with the IT passing?

azext_iot/central/services/device.py
```diff
@@ -97,9 +97,13 @@ def list_devices(
...
     query_parameters["api-version"] = api_version
 
+    logger.warning(
+        "This command may take a long time to complete if your app contains a lot of devices"
+    )
+
     pages_processed = 0
-    while (pages_processed <= max_pages) and url:
-        response = requests.get(url, headers=headers, params=query_parameters)
+    while (max_pages == 0 or pages_processed <= max_pages) and url:
```
vilit1, line 105: Should this be pages_processed < max_pages?

let's say we have max_pages = 1
We start with pages_processed at 0, so we get and process that page, increase the count of pages processed.
Now that pages_processed is 1, we should not process another page (but the code says we will).

lucadruda: good catch. fixed here and in other places with same bug

yang-hai-feng (approved):

vilit1 (approved):

PR #406, opened 2021-08-09
Dt create

vilit1:
image

---

### General Guidelines


azext_iot/digitaltwins/providers/resource.py
```diff
@@ -71,6 +72,25 @@ def create(
...
+                state = instance.get('provisioning_state', None)
+                retries = 0
+                while state == "Provisioning" and retries < MAX_ADT_CREATE_RETRIES:
+                    retries += 1
+                    sleep(1)
+                    lro.update_status()
+                    instance = lro.resource().as_dict()
+                    state = instance.get('provisioning_state', None)
+                if retries == MAX_ADT_CREATE_RETRIES:
+                    logger.warning(
```
vilit1, line 87: I would like to know if it is better to raise an error or keep as a warning.

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -71,6 +72,25 @@ def create(
...
             )
 
+            def check_state(lro):
+                from time import sleep
+                instance = lro.resource().as_dict()
+                state = instance.get('provisioning_state', None)
+                retries = 0
+                while state == "Provisioning" and retries < MAX_ADT_CREATE_RETRIES:
+                    retries += 1
+                    sleep(1)
```
digimaun, line 82: We should wait longer than 1 second, IIRC the 'Retry-After' header value (for which the client should poll the location header URI) for this api version is 10 (seconds).

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -71,6 +72,25 @@ def create(
                 long_running_operation_timeout=timeout,
             )
 
+            def check_state(lro):
+                from time import sleep
+                instance = lro.resource().as_dict()
+                state = instance.get('provisioning_state', None)
+                retries = 0
+                while state == "Provisioning" and retries < MAX_ADT_CREATE_RETRIES:
```
digimaun, line 80: Think it would be good to have the provisioning state values we care about in an enum

azext_iot/digitaltwins/providers/resource.py
digimaun: The msrest ARM poller has the following provisioning state values

`FINISHED = frozenset(['succeeded', 'canceled', 'failed'])`
`FAILED = frozenset(['canceled', 'failed'])`
`SUCCEEDED = frozenset(['succeeded'])`

If its still provisioning (not a terminal state) at the end of our polling then I think a warning is good. However if a failed provisioning state has been reached and nothing else handles it then an error should be thrown.

vilit1: There isn't any try excepts within the code or poller code (armpoller and the lropoller) that would suggest an error would be caught and not thrown again. 

Is there a way to create a dt instance that would fail but not on the initial call? (So we go into the check_state)

azext_iot/digitaltwins/providers/resource.py
vilit1: The reason for the 1 second was through ad-hoc testing with Ryan - it seemed like some instances needed as little as 2 seconds, some up to 4 seconds, with most averaging at 3-4 seconds to propogate the hostname.

Will update to use the retry-after value to align with CLI documentation unless there is merit to keep intervals of 1 second (which I see - at least 6 seconds faster if using 1 second intervals).

azext_iot/digitaltwins/providers/resource.py
c-ryan-k: I believe it's possible - Essentially since the resource is still `provisioning` after create (and the poller returns to us) - it's definitely plausible that something happens inside that final provisioning step where we'd run a GET and get a failed provisioning state.

azext_iot/digitaltwins/providers/resource.py
vilit1: Have discussed this - will keep 10 seconds

vilit1:
image

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -71,6 +74,25 @@ def create(
...
+                instance = lro.resource().as_dict()
+                state = instance.get('provisioning_state', None)
+                retries = 0
+                while (state.lower() not in ProvisioningStateType.FINISHED.value) and retries < MAX_ADT_CREATE_RETRIES:
+                    retries += 1
+                    sleep(ADT_CREATE_RETRY_AFTER)
+                    lro.update_status()
+                    instance = lro.resource().as_dict()
+                    state = instance.get('provisioning_state', None)
+                if state.lower() not in ProvisioningStateType.FINISHED.value:
```
c-ryan-k, line 88: Is it possible that somehow `provisioning_state` doesn't exist on this object? Just a concern that if we retry `MAX_ADT_CREATE_RETRIES` times, and something happens to the `provisioning_state`:

`state = instance.get('provisioning_state', None)` -> assigns `None` to `state`

`if state.lower()` -> `AttributeError: 'NoneType' object has no attribute 'lower'`

Maybe we just need `if state and state.lower()`?

vilit1: Technically, there should always be a provisioning_state. However, will change it to be state and state.lower in case there is some weirdness

c-ryan-k (dismissed):
Looks good to me - @digimaun should we wait for the next release or do you think this one's good to add in now?

azext_iot/digitaltwins/common.py
```diff
@@ -11,6 +11,9 @@
 
 from enum import Enum
 
+MAX_ADT_CREATE_RETRIES = 5
+ADT_CREATE_RETRY_AFTER = 10
```
digimaun, line 15: Why not use the API header value `Retry-After` so the client has a dynamic adherence to a service specified wait time? If `Retry-After` is missing, then we can fallback to a hard coded wait time. The ARM [async operation spec](https://github.com/Azure/azure-resource-manager-rpc/blob/master/v1.0/async-api-reference.md#202-accepted-and-location-headers) states "3. Clients should wait for the Retry-After interval, if it was specified, or the default of **60 seconds** if it was not, before polling again."

vilit1: updated

vilit1:
image

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -97,6 +119,7 @@ def rbac_handler(lro):
                             )
 
             create_or_update.add_done_callback(rbac_handler)
+            create_or_update.add_done_callback(check_state)
```
digimaun, line 122: Do the callbacks run serially (assuming so)? We would want to run check_state prior to rbac_handler since rbac_handler assumes the resource has been fully provisioned.

azext_iot/tests/digitaltwins/test_dt_resource_lifecycle_int.py
```diff
@@ -141,7 +141,8 @@ def test_dt_resource(self):
         sleep(60)
 
         assert_common_resource_attributes(
-            self.wait_for_hostname(create_msi_output),
+            self.wait_for_hostname(
```
digimaun, line 144: We should consider retooling `wait_for_hostname` to not hide LRO issues (now that we know/confirmed the service behavior), so we can be confident the check_state callback is working, without additional wait loops https://github.com/vilit1/azure-iot-cli-extension/blob/dt_create2/azext_iot/tests/digitaltwins/__init__.py#L168

azext_iot/digitaltwins/providers/resource.py
vilit1: image
will switch

azext_iot/tests/digitaltwins/test_dt_resource_lifecycle_int.py
vilit1: Two wait calls will still be required - both are for identity assignments since the service can return a succeeded state without updating the identity

vilit1:
image

azext_iot/tests/digitaltwins/test_dt_resource_unit.py
```diff
@@ -0,0 +1,166 @@
...
+            method=responses.PUT,
+            url=re.compile(
+                "https://(.*)management.azure.com/subscriptions/(.*)/"
+                "resourceGroups/{}/providers/Microsoft.DigitalTwins/digitalTwinsInstances/{}".format(
+                    resource_group_name, name
+                )
+            ),
+            body=provisioning,
+            status=201,
+            headers={
```
digimaun, line 72: We could add the `Retry-After` header to the mock to ensure header value is being honored.

digimaun (approved):

PR #407, opened 2021-08-09
Topic space3

vilit1:
With respect to updates before 8/9/2021 and the bug bash on 8/6/2021

---

### General Guidelines


azext_iot/commands.py
```diff
@@ -132,6 +132,15 @@ def load_command_table(self, _):
     ) as cmd_group:
         cmd_group.show_command("show", "iot_hub_connection_string_show")
 
+    with self.command_group(
+        "iot hub topic-space", command_type=iothub_ops
```
digimaun, line 136: The preview topic-space commands need to be attributed as `is_preview=True` prior to end user distribution.

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3304,173 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+    resolver = SdkResolver(target=target)
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    topic_type = TopicSpaceType(topic_type)
+    if topic_type == TopicSpaceType.HighFanout:
+        raise CLIError("Only LowFanout and PublishOnly topic types are supported right now")
+
+    if topic_template[0].endswith(".json") and os.path.exists(topic_template[0]):
+        try:
+            topic_template = json.loads(read_file_content(topic_template[0]))
```
digimaun, line 3337: We commonly use `shell_safe_json_parse` or a wrapper on it `process_json_arg` to handle json input

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3305,173 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+        hub_name=hub_name,
+        resource_group_name=resource_group_name,
+        login=login,
+        auth_type=auth_type_dataplane,
+    )
+    resolver = SdkResolver(target=target)
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    topic_type = TopicSpaceType(topic_type)
+    if topic_type == TopicSpaceType.HighFanout:
```
digimaun, line 3333: I guess this is point in time. Weird to have a valid arg (HighFanout) that is not actually supported yet.

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3305,173 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+        raise CLIError("Only LowFanout and PublishOnly topic types are supported right now")
+
+    if topic_template[0].endswith(".json") and os.path.exists(topic_template[0]):
```
digimaun, line 3336: We shouldn't limit input files to .json extension only.

azext_iot/_params.py
```diff
@@ -560,6 +561,26 @@ def load_arguments(self, _):
...
+    with self.argument_context("iot hub topic-space") as context:
+        context.argument(
+            "topic_name",
+            options_list=["--topic-space-name", "--tsn"],
+            help="Topic space name.",
+        )
+        context.argument(
+            "topic_template",
+            nargs="*",
+            options_list=["--topic-space-template", "--template"],
```
digimaun, line 573: Plurality changes `--topic-space-templates` and `--templates`

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3305,173 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+        try:
+            topic_template = shell_safe_json_parse(read_file_content(topic_template[0]))
+        except OSError as e:
+            raise CLIError("Could not read file {}. Error: {}".format(topic_template, e))
```
digimaun, line 3340: Take a look at `process_json_arg()` to see a common processing pattern, and also we've added some error hints in case a user puts in a malformed file path.

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3305,173 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+        resource_group_name=resource_group_name,
+        login=login,
+        auth_type=auth_type_dataplane,
+    )
+    resolver = SdkResolver(target=target)
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    try:
+        return service_sdk.topic_space.delete_topic_space(
+            id=topic_name
```
digimaun, line 3474: does delete return a value? if so lets return with `raw=True`

azext_iot/operations/hub.py
vilit1: It will be supported at some point

azext_iot/operations/hub.py
vilit1: Done

vilit1:
image

azext_iot/_help.py
```diff
@@ -295,6 +295,30 @@
         --auth-type identity --identity {managed_identity_resource_id}
 """
 
+helps[
+    "iot hub device-identity generate-user-credentials"
```
digimaun, line 299: Why `generate-user-credentials` and not `generate-mqtt-credentials`?

azext_iot/_help.py
```diff
@@ -295,6 +295,30 @@
         --auth-type identity --identity {managed_identity_resource_id}
 """
 
+helps[
+    "iot hub device-identity generate-mqtt-credentials"
+] = """
+    type: command
+    short-summary: Generate user credentials for MQTT Device sign in.
```
digimaun, line 302: Should be along the lines of "Generate mqtt credentials for device or module connections."

azext_iot/_help.py
```diff
@@ -295,6 +295,30 @@
...
+    long-summary: For more information, see
+                  https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#import-and-export-device-identities
```
digimaun, line 304: This link https://docs.microsoft.com/azure/iot-hub/iot-hub-devguide-identity-registry#import-and-export-device-identities is unrelated to mqtt connections.

azext_iot/_help.py
digimaun: lets also update the example verbiage

azext_iot/_params.py
```diff
@@ -495,6 +496,41 @@ def load_arguments(self, _):
...
+            options_list=["--password-creation-time", "--pct"],
+            help="Integer representing time in seconds representing the password creation time.",
+        )
+        context.argument(
+            "product_info",
+            options_list=["--product-info", "--pi"],
+            help="Description of the product.",
+        )
+        context.argument(
+            "shared_access_key_name",
```
digimaun, line 524: We have an existing parameter that can be used which is policy_name - technically the shared access policy is the entity that contains keys.

azext_iot/_params.py
```diff
@@ -495,6 +496,41 @@ def load_arguments(self, _):
...
+        )
+        context.argument(
+            "password_creation_time",
+            options_list=["--password-creation-time", "--pct"],
+            help="Integer representing time in seconds representing the password creation time.",
+        )
+        context.argument(
+            "product_info",
+            options_list=["--product-info", "--pi"],
+            help="Description of the product.",
```
digimaun, line 521: Do length restrictions exist?

azext_iot/_params.py
```diff
@@ -495,6 +496,41 @@ def load_arguments(self, _):
...
+            options_list=["--connection-string", "--cs"],
+            help="Target connection string. This bypasses the IoT Hub registry and generates the user credentials directly "
+                 "from the supplied symmetric key without further validation. All other command parameters aside from "
+                 "password-creation-time and password-expiry-time will be ignored. Supported connection string types: Device, "
+                 "Module."
+        )
+        context.argument(
+            "password_expiry_time",
+            options_list=["--password-expiry-time", "--pet"],
+            help="Integer representing time in seconds representing the password expiry time.",
```
digimaun, line 511: Would be good to say the default expiry of 1 hour if omitted (as its currently implemented).

azext_iot/_params.py
```diff
@@ -495,6 +496,41 @@ def load_arguments(self, _):
...
+        )
+        context.argument(
+            "password_creation_time",
+            options_list=["--password-creation-time", "--pct"],
+            help="Integer representing time in seconds representing the password creation time.",
```
digimaun, line 516: I don't understand this -- is it supposed to be seconds since unix epoch (utc)?

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3303,290 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+    if topic_templates and looks_like_file(topic_templates[0]):
+        topic_templates = process_json_arg(topic_templates[0], "Topic Space Templates")
+
+    try:
+        return service_sdk.topic_space.put_topic_space(
+            id=topic_name,
+            name=topic_name,
+            properties={
+                "topicspace_type": topic_type,
+                "topic_templates": topic_templates
```
digimaun, line 3347: What happens if topic_templates is `[]` , does the service handle well or should the client handle?

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3303,290 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+        )
+        resolver = SdkResolver(target=target)
+        service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    topic_type = TopicSpaceType(topic_type)
+    if topic_type == TopicSpaceType.HighFanout:
+        raise CLIError("Only LowFanout and PublishOnly topic types are supported right now")
+
+    if topic_templates and looks_like_file(topic_templates[0]):
+        topic_templates = process_json_arg(topic_templates[0], "Topic Space Templates")
```
digimaun, line 3339: If the topic_templates looks like `["a/file/path.json", "d${hello}"]`, then the non-zero index from the initial list gets silently dropped.

azext_iot/operations/hub.py
vilit1: Service handles this well

azext_iot/operations/hub.py
vilit1: This is where it gets more confusing - technically "a/file/path.json" could be a valid template because . are allowed in the template names...

Should I make a rule where if length of the templates is 1 then check if it is a file? Otherwise assume that all are templates?

azext_iot/operations/hub.py
vilit1: Service does not handle duplicates templates well though

azext_iot/_params.py
```diff
@@ -495,6 +496,44 @@ def load_arguments(self, _):
...
+        context.argument(
+            "dtmi",
+            options_list=["--dtmi"],
+            help="Digital Twins model Id. Example: dtmi:com:example:Room;2",
+        )
+        context.argument(
+            "version",
+            options_list=["--format-version", "--fv"],
+            help="MQTT Connect Credentials Format Version to use when generating the credentials. Supported values: v1, v2."
+                 " Default: v2",
```
digimaun, line 534: We shouldn't need to indicate the default text here if a default value is assigned to the respective command handler input.

Also would be good idea to use an enum instead of a string.

azext_iot/operations/hub.py
digimaun: Talked about changes here.

azext_iot/_params.py
```diff
@@ -495,6 +496,44 @@ def load_arguments(self, _):
...
+                 "Module."
+        )
+        context.argument(
+            "password_expiry_in_secs",
+            options_list=["--password-expiry-in-secs", "--pes"],
+            type=int,
+            help="Valid password duration in seconds.",
+        )
+        context.argument(
+            "password_creation_time",
```
digimaun, line 515: Makes sense to eliminate this and simplify the command surface.

azext_iot/_params.py
```diff
@@ -495,6 +496,44 @@ def load_arguments(self, _):
...
+            "connection_string",
+            options_list=["--connection-string", "--cs"],
+            help="Target connection string. This bypasses the IoT Hub registry and generates the user credentials directly "
+                 "from the supplied symmetric key without further validation. All other command parameters aside from "
+                 "password-creation-time and password-expiry-in-secs will be ignored. Supported connection string types: Device, "
+                 "Module."
+        )
+        context.argument(
+            "password_expiry_in_secs",
+            options_list=["--password-expiry-in-secs", "--pes"],
```
digimaun, line 510: We could use `--duration` like we do in `generate-sas-token`.

azext_iot/_params.py
```diff
@@ -560,6 +599,26 @@ def load_arguments(self, _):
...
+        context.argument(
+            "topic_name",
+            options_list=["--topic-space-name", "--tsn"],
+            help="Topic space name.",
+        )
+        context.argument(
+            "topic_templates",
+            nargs="*",
+            options_list=["--topic-space-templates", "--templates"],
+            help="List of topic space template paths or a text file containing a list of topic "
```
digimaun, line 612: Maybe should say something about, if an input is a file, what the file format should be?

azext_iot/_params.py
```diff
@@ -495,6 +496,38 @@ def load_arguments(self, _):
...
+            help="Target connection string. This bypasses the IoT Hub registry and generates the user credentials directly "
+                 "from the supplied symmetric key without further validation. All other command parameters aside from "
+                 "password-creation-time and password-expiry-in-secs will be ignored. Supported connection string types: Device, "
+                 "Module."
+        )
+        context.argument(
+            "duration",
+            options_list=["--duration", "--du"],
+            type=int,
+            help="Valid password duration in seconds.",
```
digimaun, line 512: nit, something like "Indicates the duration in seconds, for which the mqtt credentials will remain valid."

azext_iot/_params.py
```diff
@@ -495,6 +496,38 @@ def load_arguments(self, _):
             "Account and Contributor role for the IoT Hub.",
         )
 
+    with self.argument_context("iot hub device-identity generate-mqtt-credentials") as context:
+        context.argument(
+            "connection_string",
+            options_list=["--connection-string", "--cs"],
+            help="Target connection string. This bypasses the IoT Hub registry and generates the user credentials directly "
```
digimaun, line 503: help should be updated due to updated args

azext_iot/_params.py
digimaun: "MQTT connection credentials format version to use ... ..."

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3303,311 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+def iot_hub_topic_space_create(
+    cmd,
+    topic_name,
+    topic_templates,
+    topic_type,
+    hub_name=None,
+    resource_group_name=None,
+    login=None,
+    auth_type_dataplane=None,
+    target=None,
```
digimaun, line 3320: What is the purpose of target here? Why is it exposed to users?

digimaun: Same thing for service sdk?

azext_iot/_params.py
```diff
@@ -495,6 +497,37 @@ def load_arguments(self, _):
...
+        context.argument(
+            "dtmi",
+            options_list=["--dtmi"],
+            help="Digital Twins model Id. Example: dtmi:com:example:Room;2",
+        )
+        context.argument(
+            "version",
+            arg_type=get_enum_type(MQTTConnectVersionType),
+            options_list=["--format-version", "--fv"],
+            help="MQTT connect credentials format version to use when generating the credentials.",
```
digimaun, line 528: The help should state when to use which format i.e. api versions.

azext_iot/operations/hub.py
vilit1: Was going to reuse the code here for update but will remove this.

azext_iot/operations/hub.py
```diff
@@ -3301,3 +3303,311 @@ def _customize_device_tracing_output(device_id, desired, reported):
...
+    cmd,
+    device_id=None,
+    connection_string=None,
+    dtmi=None,
+    hub_name=None,
+    module_id=None,
+    duration=3600,
+    policy_name=None,
+    product_info=None,
+    version=MQTTConnectVersionType.v2.value,
```
vilit1, line 3491: TODO: remove default value and change to None

azext_iot/_params.py
vilit1: Something like:
Use v1 for MQTT 3.1 or below and v2 for MQTT 3.1.1 or higher

PR #408, opened 2021-08-13
Enable tests against custom IoT Central endpoints

lucadruda:
With this changes, IoT Central integration tests can be run against internal clusters or different cloud.
Technically, it exposes the usual "--central-dns-suffix" as a configuration option for pytest.
No change in any commands.

image


azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -552,14 +539,14 @@ def test_central_device_registration_info_unassociated(self):
             == "Device does not have a valid template associated with it."
         )
 
-    @pytest.mark.skipif(
-        not DEVICE_ID, reason="empty azext_iot_central_primarykey env var"
+    @ pytest.mark.skipif(
```
digimaun, line 542: Unintended space?

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -23,7 +23,7 @@
 APP_PRIMARY_KEY = os.environ.get("azext_iot_central_primarykey")
 APP_SCOPE_ID = os.environ.get("azext_iot_central_scope_id")
 DEVICE_ID = os.environ.get("azext_iot_central_device_id")
-TOKEN = os.environ.get("azext_iot_central_token")
+TOKEN = os.environ.get("azext_iot_central_token", None)
```
digimaun, line 26: `None` should be the default if the `azext_iot_central_token` key does not have a value.

azext_iot/central/providers/monitor_provider.py
```diff
@@ -90,15 +91,15 @@ def _build_targets(
         return targets
 
     def _build_handler(
```
digimaun, line 93: Would recommend using python black for auto formatting.

azext_iot/tests/central/test_iot_central_int.py
lucadruda: good catch. fixed

azext_iot/tests/central/test_iot_central_int.py
lucadruda: removed

azext_iot/central/providers/monitor_provider.py
lucadruda: changed to black and re-run formatting on the folder

digimaun (approved):

PR #440, opened 2021-09-28
Fix tests for central and python 3.9

lucadruda:
### General Guidelines

- [x] Have **all** unit **and** integration tests passed locally? i.e. `pytest <project root> -vv`. Please provide evidence in the form of a screenshot showing a succesful run of integration tests.

Fixes minor bugs in central tests where it assumes device templates can always be deleted while this can't happen if there are associated devices.
Also align with python 3.9 changes in asyncio.

### DIFF
--- a/azext_iot/tests/central/test_iot_central_devices_int.py
+++ b/azext_iot/tests/central/test_iot_central_devices_int.py
@@ -99,9 +99,12 @@ class TestIotCentralDevices(CentralLiveScenarioTest):
             checks=[self.check("deviceId", device_id)],
         )
         self._delete_device(device_id=device_id, api_version=self._api_version)
-        self._delete_device_template(
-            template_id=template_id, api_version=self._api_version
-        )
+        try:
+            self._delete_device_template(
+                template_id=template_id, api_version=self._api_version
+            )
+        except:
+            pass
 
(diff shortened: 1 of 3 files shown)


lucadruda:
Integration tests
image
image

azext_iot/tests/central/test_iot_central_devices_int.py
```diff
@@ -99,9 +99,12 @@ def test_central_device_twin_show_success(self):
             checks=[self.check("deviceId", device_id)],
         )
         self._delete_device(device_id=device_id, api_version=self._api_version)
-        self._delete_device_template(
-            template_id=template_id, api_version=self._api_version
-        )
+        try:
+            self._delete_device_template(
+                template_id=template_id, api_version=self._api_version
```
digimaun, line 104: If there is an issue with deleting device templates, then hide the error? What about introducing a polling element to ensure devices associated with a target template are first deleted?

lucadruda: assuming we always operate with CLI-born device templates, we can delete devices with no worries.
I followed your suggestion and delete associated devices for the template. INT tests screenshots updated

digimaun (approved):

PR #451, opened 2021-10-29
Optional parameter for dps compute-device-key

vilit1:
---

### General Guidelines

- [ ] Have **all** unit **and** integration tests passed locally? i.e. `pytest <project root> -vv`. Please provide evidence in the form of a screenshot showing a succesful run of integration tests.

### DIFF
--- a/azext_iot/constants.py
+++ b/azext_iot/constants.py
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "255.255.1"
+VERSION = "0.11.0"
 EXTENSION_NAME = "azure-iot"
 EXTENSION_ROOT = os.path.dirname(os.path.abspath(__file__))
 EXTENSION_CONFIG_ROOT_KEY = "iotext"
--- a/azext_iot/constants.py
+++ b/azext_iot/constants.py
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.11.0"
+VERSION = "255.255.1"
 EXTENSION_NAME = "azure-iot"
 EXTENSION_ROOT = os.path.dirname(os.path.abspath(__file__))
 EXTENSION_CONFIG_ROOT_KEY = "iotext"
--- a/azext_iot/operations/dps.py
+++ b/azext_iot/operations/dps.py
@@ -558,8 +558,38 @@ def iot_dps_device_enrollment_group_delete(
 
 
 def iot_dps_compute_device_key(
-    cmd, symmetric_key, registration_id,
+    cmd,
+    registration_id,
+    client=None,
+    enrollment_id=None,
+    dps_name=None,
+    resource_group_name=None,
+    symmetric_key=None,
 ):
+    if symmetric_key is None:
+        if not all([client, dps_name, resource_group_name, enrollment_id]):
+            raise CLIError(
+                "Please provide DPS enrollment group idenfitiers (Device Provisioning Service name via "
+                "--dps-name, Enrollment ID via --enrollment-id, and resource group via --resource-group "
+                "or -g) or the enrollment group symmetric key via --symmetric-key or --key."
+            )
+
+        target = get_iot_dps_connection_string(cmd, client, dps_name, resource_group_name)
+        try:
+            resolver = SdkResolver(target=target)
+            sdk = resolver.get_sdk(SdkType.dps_sdk)
+            attestation = sdk.get_enrollment_group_attestation_mechanism(
+                enrollment_id, raw=True
+            ).response.json()
+            if attestation.get("symmetricKey") is None:
+                raise CLIError(
+                    "Requested enrollment group has an attestation type of 'x509'. Currently, compute-device-key "
+                    "is only supported for symmetric key enrollment groups"
+                )
+            symmetric_key = attestation["symmetricKey"]["primaryKey"]
+        except ProvisioningServiceErrorDetailsException as e:
+            raise CLIError(e)
+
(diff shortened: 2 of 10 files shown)


azext_iot/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.11.0"
+VERSION = "255.255.1"
```
digimaun, line 10: Please leave the version value or set to `0.11.1`.

azext_iot/operations/dps.py
```diff
@@ -558,8 +558,38 @@ def iot_dps_device_enrollment_group_delete(
...
+    cmd,
+    registration_id,
+    client=None,
+    enrollment_id=None,
+    dps_name=None,
+    resource_group_name=None,
+    symmetric_key=None,
 ):
+    if symmetric_key is None:
+        if not all([client, dps_name, resource_group_name, enrollment_id]):
```
digimaun, line 570: When is `client`  `None` and what is the user supposed to do about it?

azext_iot/operations/dps.py
```diff
@@ -558,8 +558,38 @@ def iot_dps_device_enrollment_group_delete(
...
+            raise CLIError(
+                "Please provide DPS enrollment group idenfitiers (Device Provisioning Service name via "
```
digimaun, line 572: spelling idenfitiers

azext_iot/operations/dps.py
```diff
@@ -558,8 +558,38 @@ def iot_dps_device_enrollment_group_delete(
...
+            )
+
+        target = get_iot_dps_connection_string(cmd, client, dps_name, resource_group_name)
+        try:
+            resolver = SdkResolver(target=target)
+            sdk = resolver.get_sdk(SdkType.dps_sdk)
+            attestation = sdk.get_enrollment_group_attestation_mechanism(
+                enrollment_id, raw=True
+            ).response.json()
+            if attestation.get("symmetricKey") is None:
```
digimaun, line 584: I think evaluating the attestation record type property to ensure it is of type symmetricKey would be better for this check. We also wouldn't need to hardcode 'x509' or a future attestation type.

azext_iot/tests/dps/test_iot_dps_int.py
```diff
@@ -596,6 +596,95 @@ def test_dps_enrollment_group_lifecycle(self):
             )
         )
 
+    def test_dps_enrollment_group_symmetrickey_lifecycle(self):
```
digimaun, line 599: What about adding a test (unit is fine too) that exercises the  raised CLIError if the attestation type is not symmetricKey?

azext_iot/operations/dps.py
vilit1: Updated the conditional and error message - there is also tpm as a type.

azext_iot/operations/dps.py
vilit1: In the original, the client was not passed in at all because no calls were needed to calculate the key offline.

azext_iot/tests/dps/test_iot_dps_int.py
vilit1: Added to test above where an enrollment group with a certificate is created.

vilit1: Added to test above which already creates an enrollment group with a cert

vilit1:
image

azext_iot/operations/dps.py
digimaun: Gotcha...to clarify, why is client optional i.e. `client=None` vs required, and related, client is not exposed as a user param correct? Yet it is one of the required elements in the  `if not all()` conditional - so if a user environment has client=None they would trigger the CLIError but there is nothing in the message indicating aht to do in that case.

vilit1: I have changed the client to be required + no longer tracked in the all exist bucket.

To answer your questions:
1. Why is client optional/not optional? The client is only needed for online commands, i.e. getting the dps instance which happens if the user does not provide the symmetric key. The client should be autopopulated as a parameter.
2. If a user's environment has client=None? Not sure what happens but in that scenario other dps commands would be affected too. Client gets populated based on cmd so the user should not be trying to mess with this parameter.

And before you ask: Why did I change the client to be required + removed it from the bucket?
Client is required in other commands so the change is more for consistency.  It should be auto-populated based on the cmd, thus it will have a value no matter what.

digimaun: Thanks. The first line in your reply is what I was driving at, and you made the change. I didn't see the reason to have client be optional when the DPS client is configured to be always passed in (if the respective arg is present) , and have it in the `if not all` check when its not a user stemming issue.

digimaun (approved):

azext_iot/tests/dps/test_iot_dps_int.py
```diff
@@ -596,6 +605,95 @@ def test_dps_enrollment_group_lifecycle(self):
...
+        assert offline_device_key == online_device_key
+
+        # Compute Device Key uses primary key
+        offline_device_key = self.cmd(
+            'az iot dps compute-device-key --key "{}" '
+            "--registration-id myarbitrarydeviceId".format(keys["secondaryKey"])
+        ).output
+        assert offline_device_key != online_device_key
+
+        etag = self.cmd(
```
digimaun, line 667: Does this etag extraction get used later?

PR #471, opened 2021-12-03
DPS SDK update + RBAC support

vilit1:
Update to generated sdk. The operations got generated as a group of files rather than one single file with all the commands. DPS.py was updated to reflect that.

- api version is now 10-01-2021
- individual enrollments now have optional device information
- added param shortcuts for enrollment-id
- added support for RBAC credentials

image

---

### General Guidelines

Intent for Production

Basic expectations


vilit1:
image
Top test run is when dps name is specified (and has the correct DPS data Contributor role)
Bottom run is when dps name is not specified so the dps is generated

HISTORY.rst
```diff
@@ -20,6 +20,8 @@ Release History
 
 **IoT DPS updates**
 
+* Added RBAC support with --auth-type login.
+
```
c-ryan-k, line 24: We might want to add something about an API-version update and some of the new added parameters (`--device-info`, etc)

vilit1: done - could not find official documentation for rbac for dps yet.

HISTORY.rst
```diff
@@ -20,6 +20,27 @@ Release History
 
 **IoT DPS updates**
 
+* Added RBAC support for dataplane commands, similar to the RBAC support for IoT Hub.
+  The type of auth used to execute commands can be controlled with the "--auth-type" parameter
+  which accepts the values "key" or "login". The value of "key" is set by default.
+
+  * When "--auth-type" has the value of "key", like before the CLI will auto-discover
+    a suitable policy when interacting with iothub.
```
c-ryan-k, line 28: This is supposed to refer to DPS, right?

c-ryan-k:
@vilit1 - Looks good to me (minus the few nit comments about history), @digimaun would you mind giving the RBAC auth parameter / token generation for DPS a quick once-over since you scaffolded the original IoT Hub scenarios? Looks like we switched to using a separate DPS and Hub auth-type (which I think is the right move) - just making sure I didn't miss anything that might cause issues there.

azext_iot/dps/providers/aad_oauth.py
```diff
@@ -0,0 +1,62 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from msrest.authentication import Authentication
+
+
+class IoTDPSOAuth(Authentication):
```
digimaun, line 10: We don't need another auth module as there are no additional customizations here -- we can probably use the same aad_oauth module across services IoT Hub/DPS/DigitalTwins/Central putting it in a generic location. There is a oneoff auth.py in azext_iot/common used by Central. That particular one could be handled later.

azext_iot/_params.py
```diff
@@ -39,7 +39,23 @@
...
+dps_auth_type_dataplane_param_type = CLIArgumentType(
+    options_list=["--auth-type"],
+    arg_type=get_enum_type(
+        AuthenticationTypeDataplane, AuthenticationTypeDataplane.key.value
+    ),
+    arg_group="Access Control",
+    help="Indicates whether the operation should auto-derive a policy key or use the current Azure AD session. "
+    "You can configure the default using `az configure --defaults iotdps-data-auth-type=<auth-type-value>`",
+    configured_default="iotdps-data-auth-type",
+    local_context_attribute=LocalContextAttribute(
```
digimaun, line 51: Lets remove usage of local_context_attribute here and for hub_auth_type_dataplane_param_type. It is still experimental functionality and we haven't advertised it.

azext_iot/_params.py
```diff
@@ -958,6 +979,11 @@ def load_arguments(self, _):
             help="The secondary symmetric shared access key stored in base64 format. ",
             arg_group="Authentication"
         )
+        context.argument(
+            "device_information",
+            options_list=["--device-information", "--info"],
+            help="Optional device information.",
```
digimaun, line 985: Ideally we could provide a bit more detail, but this is something service docs should cover.

azext_iot/tests/dps/__init__.py
```diff
@@ -108,6 +110,31 @@ def create_dps(self):
...
+            )
+        ).get_output_in_json()
+
+        account = self.cmd("account show").get_output_in_json()
+        user = account["user"]
+
+        if user["name"] is None:
+            raise Exception("User not found")
+
+        while True:
```
digimaun, line 125: I know this is how it works on the IoT Hub side as well, but we should think about a max attempt threshold ....like after 10 applications of the role assignment and waiting for 10 seconds, we abort and raise an exception....something like that.

azext_iot/tests/dps/test_iot_dps_int.py
```diff
@@ -46,7 +46,8 @@ def test_dps_enrollment_tpm_lifecycle(self):
                     "iot dps enrollment create --enrollment-id {} --attestation-type {}"
```
digimaun, line 46: Not in this iteration. We should breakout/separate DPS unit and integration tests similar to the work we have done for IoT Hub.

azext_iot/_params.py
vilit1: It has the same format as tags, which is shown in the help example.

digimaun: I'm talking about the business use case. Just knowing the format doesn't cover this.

azext_iot/_params.py
vilit1: It's been there for a while - I would rather keep both local context attributes (at least the hub one)

There may be users using the local hub variable because the variable has been there. It makes more sense to separate the hub and dps variables (especially as the dps rbac is a newer feature) to let users more granular control. Thus, I lean heavily on both being present.

HISTORY.rst
vilit1: rip

azext_iot/_params.py
digimaun: local context in az CLI is both deprecated and experimental 
image

and is independent from az cli configuration.

digimaun: I also don't understand this comment 

"It makes more sense to separate the hub and dps variables (especially as the dps rbac is a newer feature) to let users more granular control. "

As this feedback item is not making an assertion against separate hub and dps variables.

vilit1:
image

vilit1:
image

azext_iot/_params.py
vilit1: removed.

vilit1:
DPS IT:
image

vilit1:
IOT hub 
image

HISTORY.rst
```diff
@@ -49,6 +49,27 @@ Release History
 
 **IoT DPS updates**
 
+* Added RBAC support for DPS dataplane commands, similar to the RBAC support for IoT Hub.
```
digimaun, line 52: These DPS update notes need to be placed/combined with existing notes under 0.12.2 (unreleased). Then please change the 0.12.2 version in notes -> 0.13.0 and also make the same change to `VERSION` in `constants.py` from `0.12.2` to `0.13.0`.

digimaun (approved):

PR #474, opened 2021-12-13
Adding missing "Update" sub-commands for IoT Central commands

lucadruda:
---
This PR adds the "update" sub-commands for all the commands allowing patches as per API definition (https://docs.microsoft.com/en-us/rest/api/iotcentral/).
The "update" sub-commands are only available for API-Version==1.1-preview.


lucadruda:
Last INT tests run:
inttests
ittests2

HISTORY.rst
```diff
@@ -3,6 +3,14 @@
 Release History
 ===============
 
+0.12.1
```
digimaun, line 6: Can we please plug this change set under `0.12.2`?

lucadruda: sure thing

HISTORY.rst
```diff
@@ -3,6 +3,14 @@
 Release History
 ===============
 
+0.12.2
++++++++++++++++
+
+**IoT Central updates**
+
+* Added missing "update" sub-commands for all commands supporting it 
+  ("device", "device-template", "file-upload-config", "organization", "user")
```
vilit1, line 12: personally I like having the full commands like so:
 * az iot central device
 * az iot central device-template

azext_iot/central/_help.py
```diff
@@ -352,6 +352,28 @@ def _load_central_devices_help():
             --simulated
     """
 
+    helps[
+        "iot central device update"
+    ] = """
+        type: command
+        short-summary: Update a device in IoT Central.
```
vilit1, line 359: Is there a limit to what a user can update? For example, they cannot update the device-id, correct?

azext_iot/central/_help.py
```diff
@@ -542,6 +564,31 @@ def _load_central_users_help():
             --object-id {objectId}
             --role operator
     """
+
+    helps[
+        "iot central user update"
+    ] = """
+        type: command
+        short-summary: Update a user in the application
```
vilit1, line 572: same comment as update for devices - what limits are there?

azext_iot/central/commands_device_template.py
```diff
@@ -38,6 +38,7 @@ def get_device_template(
 def list_device_templates(
     cmd,
     app_id: str,
+    compact=False,
```
vilit1, line 41: Add this change to history

azext_iot/central/commands_user.py
```diff
@@ -36,15 +37,53 @@ def add_user(
         return provider.add_email(
             assignee=assignee,
             email=email,
-            role=Role[role],
+            org_id=org_id,
```
vilit1, line 40: is there a reason for a change from Role to role and org_id?

azext_iot/central/models/preview/device.py
```diff
@@ -9,14 +9,14 @@
...
+        self.approved = device.get("enabled") or device.get("approved")
         self.display_name = device.get("displayName")
         self.etag = device.get("etag")
         self.id = device.get("id")
-        self.instance_of = device.get("instanceOf")
+        self.instance_of = device.get("template") or device.get("instanceOf")
         self.provisioned = device.get("provisioned")
         self.simulated = device.get("simulated")
-        self.device_status = self._parse_device_status()
+        self._device_status = self._parse_device_status()
```
vilit1, line 19: is there a reason for "hiding" device status?

azext_iot/central/providers/organization_provider.py
```diff
@@ -103,6 +108,7 @@ def create_organization(
             org_name=org_name,
             parent_org=parent_org,
             token=self._token,
+            update=update,
```
vilit1, line 111: what is the point of specifying update if you have a check that ensures the org_id does not exist yet?

azext_iot/central/providers/device_template_provider.py
```diff
@@ -72,6 +74,17 @@ def list_device_templates(
             api_version=self._api_version,
         )
 
+        if compact:  # if asked for compact, just keep reduced info
```
vilit1, line 77: Wouldn't it be simpler to say compact means that only fields x, y, z are kept and create a separate template using that? That way you don't need to worry about having to pop unneeded fields if a new feature is added (assuming that the default compact fields stay the same).

Furthermore, this is weird that list can modify the template object here.

azext_iot/central/services/_utility.py
```diff
@@ -121,3 +121,11 @@ def get_object(data: dict, model: str, api_version) -> object:
         raise CLIError(
             "{} is not available for api version == {}".format(model, api_version)
         )
+
+
+def to_camel_dict(data: dict) -> dict:
+    keys = list(data.keys())
+    for key in keys:
+        data[to_camel_case(key)] = data.pop(key)
```
vilit1, line 129: this assumes that there won't be a mixed case dictionary with conflicting keys, correct?

Also weird that this function modifies the given data dictionary and returns it - maybe have this function just return a new dictionary without modifiying the dictionary from the argument?

HISTORY.rst
lucadruda: done

azext_iot/central/_help.py
lucadruda: added long-summary

azext_iot/central/_help.py
lucadruda: added long summary

azext_iot/central/commands_user.py
lucadruda: org_id allows to specify organizations for the roles and it's available for api 1.1-preview. That was a missing parameter in earlier versions.

role has been changed to be a string representing the role Id (GUID) or one of the built-in roles ("admin","operator","builder").
Current CLI only supports built-in roles and any other value passed into the "role" parameters causes early failure because of the enum. I moved the enum logic deeper in the code (at service level) and added a simple validation:
- role is a GUID formatted string -> pass it as is
- role is one of the built-in -> lookup enum

azext_iot/central/models/preview/device.py
lucadruda: device status is an internal concept not to be outputted to the user as per API specs. In previous versions the output dictionary was manually constructed to only specify valid keys. After code refactoring we now delegate to the framework (knack) the job of converting objects to dicts before output. The guideline in this case is to have "private" and "public" properties.

azext_iot/central/services/_utility.py
lucadruda: assumption is correct. dictionaries are not custom entities for us.
Agree on second point. Refactored.

azext_iot/central/providers/organization_provider.py
lucadruda: renamed to have "create_or_update" for better understandability.

We'll have a separate PR to remove all the checks like this "exist in" as we're going to remove all legacy caching in central commands.

lucadruda:
INT tests updated

azext_iot/central/providers/device_template_provider.py
lucadruda: followed you suggestions and refactored.

azext_iot/central/commands_device_template.py
lucadruda: added

azext_iot/central/services/user.py
```diff
@@ -57,19 +65,21 @@ def _make_call(
...
 
-def add_service_principal(
+def addorupdate_service_principal(
     cmd,
     app_id: str,
     assignee: str,
     tenant_id: str,
     object_id: str,
-    role: Role,
+    role: str,
```
yang-hai-feng, line 74: roles

azext_iot/central/services/user.py
```diff
@@ -110,18 +130,20 @@ def add_service_principal(
     return _utility.get_object(result, MODEL, api_version)
 
 
-def add_email(
+def addorupdate_email(
     cmd,
     app_id: str,
     assignee: str,
     email: str,
     role: Role,
```
yang-hai-feng, line 138: roles

azext_iot/central/services/user.py
```diff
@@ -110,18 +130,20 @@ def add_service_principal(
     return _utility.get_object(result, MODEL, api_version)
 
 
-def add_email(
+def addorupdate_email(
```
yang-hai-feng, line 133: nit: addorupdate_email_user

azext_iot/central/services/user.py
```diff
@@ -57,19 +65,21 @@ def _make_call(
     return _utility.try_extract_result(response)
 
 
-def add_service_principal(
+def addorupdate_service_principal(
```
yang-hai-feng, line 68: nit: addorupdate_service_principal_user

azext_iot/central/services/user.py
```diff
@@ -141,15 +163,23 @@ def add_email(
...
 
+    if role:
+        role_id = Role[role].value if role in get_enum_keys(Role) else role
+        if org_id and api_version == ApiVersion.v1_1_preview.value:
+            payload["roles"] = [{"role": role_id, "organization": org_id}]
+        else:
+            payload["roles"] = [{"role": role_id}]
+
+    if email:
+        payload["email"] = email
```
yang-hai-feng, line 177: Can email be updated?

azext_iot/central/models/preview/device.py
vilit1: Will there be any breaking changes/issues with respect to having this attribute now be private? This is assuming some customer decided to rely on device_status and now it is technically _device_status

azext_iot/central/_help.py
```diff
@@ -542,6 +582,36 @@ def _load_central_users_help():
...
+                    Updating the tenantId or objectId for a service principal user is not allowed.
+                    Updating the email address for an email user is not allowed.
+
+        examples:
+        - name: Update a user by email in the application
+          text: >
+            az iot central user update
+            --user-id {userId}
+            --app-id {appId}
+            --email {emailAddress}
```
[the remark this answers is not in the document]
vilit1: To add on, can we generate an email address using the user-id if we still want the email address?

azext_iot/central/params.py
```diff
@@ -96,6 +96,13 @@ def load_central_arguments(self, _):
             " Learn more at https://aka.ms/iotcentraldtmi.",
         )
 
+    with self.argument_context("iot central device-template list") as context:
+        context.argument(
+            "compact",
+            options_list=["--compact", "-c"],
+            help="Shot templates in compact mode",
```
vilit1, line 103: shot or short? Also update with what is shown in compact mode.

azext_iot/central/models/preview/device.py
lucadruda: no breaking changes as previous version did not return device_status too. we did code refactoring to simplify things a little bit and this required non-public fields to be marked explicitly with underscore to prevent them appearing in the output

azext_iot/central/_help.py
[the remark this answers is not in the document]
lucadruda: good point. I just keep user-id as the only parameter to pick the user. since "roles" is the only allowed field to be updated, both "email" and "servicePrincipal" users will have the same update syntax

azext_iot/central/_help.py
```diff
@@ -542,6 +582,36 @@ def _load_central_users_help():
...
+            --role admin
```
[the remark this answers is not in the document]
lucadruda: I won't breaking change the "create" command for now so I implemented 1:n mapping for update only.
New parameter is called "roles" and accepts comma-separated roles in the format "org_id\role_id" with org_id as optional.
The parameter syntax is explained in the help.

azext_iot/central/_help.py
```diff
@@ -542,6 +582,36 @@ def _load_central_users_help():
...
+
+        - name: Update a service-principal user in the application
+          text: >
+            az iot central user update
+            --user-id {userId}
+            --app-id {appId}
+            --tenant-id {tenantId}
```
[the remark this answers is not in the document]
lucadruda: removed

azext_iot/central/_help.py
```diff
@@ -542,6 +582,36 @@ def _load_central_users_help():
...
+            --object-id {objectId}
+            --role operator
```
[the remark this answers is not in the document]
lucadruda: see above

azext_iot/central/commands_user.py
```diff
@@ -36,15 +37,53 @@ def add_user(
...
+        role=role,
+        central_dns_suffix=central_dns_suffix,
+    )
+
+
+def update_user(
+    cmd,
+    app_id: str,
+    assignee: str,
+    role=None,
```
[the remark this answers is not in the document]
lucadruda: same as above

azext_iot/central/params.py
lucadruda: typo. "show"

yang-hai-feng (dismissed):

model.json
```diff
@@ -0,0 +1,39 @@
+{
```
digimaun, line 1: Un-intended placement of this model file?

digimaun:
Let's please close resolved feedback items.

model.json
lucadruda: good catch. thank you

azext_iot/central/params.py
```diff
@@ -23,7 +23,7 @@
 
 role_type = CLIArgumentType(
     options_list=["--role", "-r"],
```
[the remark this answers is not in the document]
lucadruda: used different parameters

azext_iot/central/providers/user_provider.py
```diff
@@ -39,31 +38,62 @@ def add_service_principal(
         assignee: str,
         tenant_id: str,
         object_id: str,
-        role: Role,
+        role: str,
```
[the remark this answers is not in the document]
lucadruda: only added on update for backward compatibility

azext_iot/central/providers/user_provider.py
```diff
@@ -104,18 +134,44 @@ def add_email(
         self,
         assignee: str,
         email: str,
-        role: Role,
+        role: str,
```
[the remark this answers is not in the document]
lucadruda: same as above

lucadruda (commented):
fix comments

azext_iot/central/services/user.py
lucadruda: added condition for update only

azext_iot/central/commands_organization.py
```diff
@@ -74,9 +74,32 @@ def create_org(
...
+    cmd,
+    app_id: str,
+    org_id: str,
+    org_name=None,
+    parent_org=None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+) -> OrganizationV1_1_preview:
+    provider = CentralOrganizationProvider(
```
vilit1, line 95: Is there any code that will block a user from trying to update org when the provided api version is not 1.1 preview? Is there at least an appropriate error from the API?

azext_iot/central/commands_user.py
```diff
@@ -36,15 +37,50 @@ def add_user(
...
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1.value,
+) -> UserType:
+    provider = CentralUserProvider(
+        cmd=cmd, app_id=app_id, api_version=api_version, token=token
+    )
+
+    if email:
+        return provider.update_email_user(
```
vilit1, line 72: Would there be a case where you would want to update the email and the service principal (ex: the update_user arguments include a new email and object_id)

azext_iot/central/models/__init__.py
```diff
@@ -3,3 +3,92 @@
...
+
+        except Exception:
+            raise CLIError("Could not parse iot central device template.")
+
+    def _get_schema_name(self, schema) -> str:
+        return "name" if "name" in schema else "@id"
+
+    def _extract_schemas(self, entity: dict) -> dict:
+        schema = entity.get("schema")
+        if schema is not None:
```
vilit1, line 32: this can be simplified to an if, elif, else

azext_iot/central/providers/device_template_provider.py
```diff
@@ -72,9 +73,25 @@ def list_device_templates(
             api_version=self._api_version,
         )
 
-        self._device_templates.update(
-            {template.id: template.raw_template for template in templates}
-        )
+        if compact:  # if asked for compact, just keep reduced info
+            for template in templates:
```
vilit1, line 77: This assumes that self._device_template won't have a key that templates does not have, correct? Would there be an edge case in which this assumption is invalid (through deletions and additions of templates)?

vilit1: Is the self._device_template used anywhere? I see that the code tries to keep it up to date but do not see any direct usage of it.

azext_iot/central/providers/user_provider.py
```diff
@@ -104,18 +137,46 @@ def add_email(
...
+    ) -> central_services.user.User:
+        if not email:
+            raise CLIError("Must specify --email when adding a user by email")
+
+        if org_id and self._api_version == ApiVersion.v1_1_preview.value:
+            roles = rf"{org_id}\{role}"
+        else:
+            roles = role
+
+        return central_services.user.addorupdate_email_user(
```
vilit1, line 152: create_or_update has the _ but addorupdate does not?

azext_iot/tests/central/test_iot_central_devices_int.py
```diff
@@ -183,6 +183,19 @@ def test_central_device_methods_CRD(self):
             )
             is not None
         )
+
+        # UPDATE
+        new_device_name = f"{device_name}_new"
+        command = "iot central device update --app-id {} -d {} --device-name {}".format(
```
vilit1, line 189: Make the update tests here change all of the allowed updates (ex: enabling/disabling a device) and make sure the same is true for the other update tests.

vilit1 (approved):
Some comments on functionality but otherwise LGTM

azext_iot/central/commands_organization.py
lucadruda: all unsupported api versions for commands are handled by outer scope so will never reach implementation.
Here's the output when api-version is forced to unsupported version
image

azext_iot/central/commands_user.py
lucadruda: "email user" and "service principal user" are two different concept in central. if email is specified, other arguments are simply ignored.

azext_iot/central/models/__init__.py
lucadruda: +1

azext_iot/central/providers/device_template_provider.py
lucadruda: the edge case is only possible if self._device_templates persists across commands. right now we removed all those cases and we just kept cache to be removed in future changes as you correctly pointed out.

azext_iot/central/providers/user_provider.py
lucadruda: +1

vilit1:
Also can you post another run of the IT tests?

azext_iot/central/commands_user.py
vilit1: Can you add to the email param description which arguments would be ignored if the email is specified?

azext_iot/tests/central/test_iot_central_devices_int.py
lucadruda: device PATCH has an ongoing bug on backend API. keep as is for now.
strengthen other tests

azext_iot/central/commands_user.py
lucadruda: added

lucadruda:
> Also can you post another run of the IT tests?
Tests updated

PR #481, opened 2022-01-11
Add commands for Edge module image terms

avagraw:
---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -1081,6 +1081,57 @@ helps[
 """
 
 helps[
+    "iot edge module image terms show"
+] = """
+    type: command
+    short-summary: Get the details of IoT Edge module image terms.
+    examples:
+    - name: Get details of IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms show --urn {publisher:offer:sku:version}
+    - name: Get details of another Azure subscription's IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms show --urn {publisher:offer:sku:version} -s {subscription_id_or_name}
+    - name: Get details of Azure Marketplace image terms using Offer ID, Plan ID and Publisher ID
+      text: >
+        az iot edge module image terms show --offer {offer_id} --plan {plan_id} --publisher {publisher_id}
+"""
+
+helps[
+    "iot edge module image terms accept"
+] = """
+    type: command
+    short-summary: Get the details of IoT Edge module image terms.
+    examples:
+    - name: Accept IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms accept --urn {publisher:offer:sku:version}
+    - name: Accept another Azure subscription's IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms accept --urn {publisher:offer:sku:version} -s {subscription_id_or_name}
+    - name: Accept Azure Marketplace image terms using Offer ID, Plan ID and Publisher ID
+      text: >
+        az iot edge module image terms accept --offer {offer_id} --plan {plan_id} --publisher {publisher_id}
+"""
+
+helps[
+    "iot edge module image terms cancel"
+] = """
+    type: command
+    short-summary: Get the details of IoT Edge module image terms.
+    examples:
+    - name: Cancel IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms cancel --urn {publisher:offer:sku:version}
+    - name: Cancel another Azure subscription's IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms cancel --urn {publisher:offer:sku:version} -s {subscription_id_or_name}
+    - name: Cancel Azure Marketplace image terms using Offer ID, Plan ID and Publisher ID
+      text: >
+        az iot edge module image terms cancel --offer {offer_id} --plan {plan_id} --publisher {publisher_id}
+"""
+
+helps[
     "iot dps"
 ] = """
     type: group
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -1081,6 +1081,27 @@ helps[
 """
 
 helps[
+    "iot edge module"
+] = """
+    type: group
+    short-summary: Manage IoT Edge modules.
+"""
+
+helps[
+    "iot edge module image"
+] = """
+    type: group
+    short-summary: Manage IoT Edge module images.
+"""
+
+helps[
+    "iot edge module image terms"
+] = """
+    type: group
+    short-summary: Manage IoT Edge module image terms.
+"""
+
+helps[
     "iot edge module image terms show"
 ] = """
     type: command
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -1084,29 +1084,29 @@ helps[
     "iot edge image"
 ] = """
     type: group
-    short-summary: Manage IoT Edge images.
+    short-summary: Manage IoT Edge module images.
 """
 
 helps[
     "iot edge image terms"
 ] = """
     type: group
-    short-summary: Manage IoT Edge image terms.
+    short-summary: Manage IoT Edge module image terms.
 """
 
 helps[
     "iot edge image terms show"
 ] = """
     type: command
-    short-summary: Get the details of IoT Edge image terms.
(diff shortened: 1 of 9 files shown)


avagraw:
The **pytest skip if** logic is not working because the role assignment doesn't happen before initialization. Hence, switching to conditional execution of the integration test.

digimaun:
> The '**pytest skip if**' logic is not working because the role assignment doesn't happen before initialization. Hence, switching to conditional execution of the integration test.

There is more than one way to skip a test. You can skip a test imperatively during test execution by using `pytest.skip()`

Example:
```
def test_function():
    if not valid_config():
        pytest.skip("unsupported configuration")
```

avagraw:
I modified the test to check for failure when service principal is used. I think that also makes the intent clear.

azext_iot/_params.py
```diff
@@ -815,6 +815,33 @@ def load_arguments(self, _):
...
+            "planId",
+            options_list=["--plan"],
+            help="IoT Edge module image billing plan.",
+        )
+        context.argument(
+            "publisherId",
+            options_list=["--publisher", "-p"],
+            help="IoT Edge module image publisher.",
+        )
+        context.argument(
```
digimaun, line 834: urn should be removed unless a provider API for lookups is discovered or created.

azext_iot/operations/hub.py
```diff
@@ -1726,6 +1727,69 @@ def iot_hub_configuration_metric_show(
...
+    urn: str = None,
+    subscription: str = None
+):
+    from azext_iot.common.embedded_cli import EmbeddedCLI
+    from azext_iot.tests.settings import UserTypes
+    from azext_iot.common.utility import get_current_user
+
+    current_user = get_current_user()
+    if current_user["type"] != UserTypes.user.value:
+        raise CLIError('Edge module image terms operations are supported only for real user AAD tokens (not service principals)')
```
digimaun, line 1744: nit: "real user" -> "user principal"

azext_iot/_help.py
```diff
@@ -1080,6 +1080,78 @@
...
+
+helps[
+    "iot edge module image terms cancel"
+] = """
+    type: command
+    short-summary: Get the details of IoT Edge module image terms.
+    examples:
+    - name: Cancel IoT Edge module image terms using URN
+      text: >
+        az iot edge module image terms cancel --urn {publisher:offer:sku:version}
```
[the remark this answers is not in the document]
digimaun: Reminder to update doc formatting.

azext_iot/tests/iothub/configurations/test_iot_config_int.py
```diff
@@ -800,3 +804,96 @@ def test_device_configurations(self):
...
+        else:
+            # Error - Command run by a service principal user
+            self.cmd(
+                "iot edge image terms show --offer {} --plan {} --publisher {}".format(
+                    terms_offerId, terms_planId, terms_publisherId
+                ),
+                expect_failure=True,
+            )
+
+    @pytest.fixture(scope='class', autouse=True)
```
digimaun, line 890: I don't see a reason the fixture needs to be autouse, by applying it explicitly its easier to understand what fixtures are at play for what tests.

azext_iot/tests/iothub/configurations/test_iot_config_int.py
```diff
@@ -800,3 +804,96 @@ def test_device_configurations(self):
...
+    def setup_edge_image_terms_tests(self):
+        # Ensure Edge image offer terms are not accepted before the test starts
+        if self.current_user["type"] == UserTypes.user.value:
+            self.cmd(
+                "iot edge image terms cancel --offer {} --plan {} --publisher {}".format(
+                    terms_offerId, terms_planId, terms_publisherId
+                )
+            )
+        yield
```
digimaun, line 899: Why yield? There is nothing happening afterwards.

avagraw: Yield signifies that the fixture is the first thing to run when the test class initializes. It's not mandatory but makes the intent clear. The same practice is being followed in central tests.

azext_iot/tests/iothub/configurations/test_iot_config_int.py
avagraw: Makes sense

azext_iot/tests/iothub/configurations/test_iot_config_int.py
digimaun: Can you provide some documentation for what you are saying?

In pytest fixtures, upon a yield, the pytest runtime moves on to other fixtures (if any) then executes the test. After the test executes, the pytest runtime will run through fixtures (in reverse order) and execute code after the yield. In other words an effective pattern with yield is a setup before a yield, and a clean up after. https://docs.pytest.org/en/6.2.x/fixture.html#yield-fixtures-recommended

avagraw: Here is an example:

image

digimaun: Sorry let me clarify, I am asking for evidence on why you think this is a correct usage of yield in a pytest fixture or how a yield of None, followed by no code makes the intent clear.

This example you have provided shows an unneeded usage of yield which we should fix.

My prior comment has a link to the official docs showing usage of yield. Here is a link to the official docs on fixture instantiation order. https://docs.pytest.org/en/6.2.x/fixture.html#fixture-instantiation-order

digimaun: I removed yield from the fixture you have shown and it doesn't make a difference for the Central tests 

image

avagraw: Yeah the 'yield' is not required and the result is same weather or not we put it. I feel it just makes the intent clear to someone reading the code that the statements above yield run as a part of the test setup (before the tests are executed). However, I can remove it if you feel strongly against it?

digimaun: Usage of fixtures in tests imply setups or teardowns for tests to function properly - they provide a fixed baseline for the test to execute reliably.

When I saw the yield followed by nothing, the intent or purpose of it was not clear to me hence this feedback. I shared documentation from pytest explaining proper usage of yield in fixtures, which is not being done here.

yield could be replaced with a return or nothing and the net effect is the same (actually slighter improved perf because pytest does not need to attempt to run code after yield).

I don't understand the benefit.

avagraw: I see your point, will remove it.

azext_iot/_help.py
```diff
@@ -1080,6 +1080,71 @@
         az iot edge deployment show-metric -m myCustomMetric -d {deployment_name} -n {iothub_name}
 """
 
+helps[
+    "iot edge image"
+] = """
+    type: group
+    short-summary: Manage IoT Edge images.
```
digimaun, line 1087: nit: the word module is fine to include in the help descriptions, just the command invocation was getting too long.

digimaun (approved):

azext_iot/operations/hub.py
```diff
@@ -1726,6 +1727,77 @@ def iot_hub_configuration_metric_show(
...
+    if urn:
+        if (publisherId or offerId or planId):
+            raise CLIError('Publisher Id or Offer Id or Plan Id should not be specified when providing URN')
+        urn_parts = urn.split(":")
+        if urn_parts and len(urn_parts) == 3:
+            command += " --publisher '{}' --offer '{}' --plan '{}'".format(urn_parts[0], urn_parts[1], urn_parts[2])
+        else:
+            raise CLIError('Incorrect URN format. The correct format is publisherId:offerId:planId')
+    else:
+        command += " --publisher '{}'".format(publisherId) if publisherId else ""
```
digimaun, line 1756: Assuming the underlying command is handling validation of empty input str's.

avagraw:
Test evidence:
image

Pipeline test evidence: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=4628&view=logs&j=181dabdb-4055-58ff-c7bd-8011e0b92875

PR #488, opened 2022-01-28
Dt data history

vilit1:
---

### General Guidelines

Intent for Production

Basic expectations


vilit1:
with ADX cluster already pre-provisioned
image

linter_exclusions.yml
```diff
@@ -13,6 +13,19 @@ dt endpoint create eventhub:
...
+  parameters:
+    adx_resource_group:
+      rule_exclusions:
+      - parameter_should_not_end_in_resource_group
+dt data-history create adx:
+  parameters:
+    eh_resource_group:
+      rule_exclusions:
+      - parameter_should_not_end_in_resource_group
+dt data-history:
```
digimaun, line 26: Placeholder to track our need for the data-history wait command.

vilit1:
Added the wait command
image

azext_iot/constants.py
```diff
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.13.1"
+VERSION = "255.255.14"
```
digimaun, line 10: Once this is ready for merge we'll have to conform to our standard release version semantics.

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
             az dt reset -n {instance_name}
     """
 
+    helps["dt data-history"] = """
+        type: group
+        short-summary: Manage and configure data history connections.
```
digimaun, line 145: The root dt data-history command space is documented as managing connections. Why not specify a sub-command group with the target entity in the command signature so it is self describing & more organized?

As in `az dt data-history connection/conn create/list/show/**/*`.  This matches most of our other command sig patterns where the last segment of the command indicates the entity type the operations apply to. We also have another command group around managing private endpoint connections for ADT https://docs.microsoft.com/en-us/cli/azure/dt/network/private-endpoint/connection?view=azure-cli-latest

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
...
+    """
+
+    helps["dt data-history create"] = """
+        type: group
+        short-summary: Creates a data history connection between a Digital Twins instance and another resource.
+    """
+
+    helps["dt data-history create adx"] = """
+        type: command
+        short-summary: Creates a data-history connection between a Digital Twins instance and an Azure Database Explorer.
```
digimaun, line 155: "Creates a data-history connection between a Digital Twins instance **and an Azure Data Explorer cluster.**" ?

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
...
+            Requires pre-created Azure Database Explorer and Eventhub resources.
```
digimaun, line 156: "Requires pre-created Azure Data Explorer and EventHub resources."

azext_iot/digitaltwins/_help.py
digimaun, line 155: We should be consistent with `data-history` verbiage in descriptions, in some case its 'data-history' and others 'data history'. I think the latter is best.

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
...
+            az dt data-history create adx -n {instance_name}
+            --cn {time_series_database_connection_name}
+            --adx-cluster-name {adx_cluster_name}
+            --adx-database-name {adx_database_name}
+            --adx-table-name {adx_table_name}
+            --eventhub {event_hub}
+            --eventhub-namespace {event_hub_namespace}
+            --eventhub-consumer-group {event_hub_consumer_group}
+        - name: Adds a data history connection to a target instance with an Event Hub and Azure
+            Database Explorer instances in different resource groups and subscriptions from the Digital
```
digimaun, line 178: Another instance of 'Azure Database Explorer', should be 'Azure Data Explorer' -- also the 'Azure' part is prob not necessary every time.

"Adds a data history connection to a target Digital Twins instance integrating with Event Hub and Data Explorer resources contained in different resource groups or subscriptions to the target Digital Twin**s** instance."?

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
...
+        examples:
+        - name: Adds a data history connection to a target instance with the $Default event hub consumer group.
+          text: >
+            az dt data-history create adx -n {instance_name}
+            --cn {time_series_database_connection_name}
+            --adx-cluster-name {adx_cluster_name}
+            --adx-database-name {adx_database_name}
+            --eventhub {event_hub}
+            --eventhub-namespace {event_hub_namespace}
+        - name: Adds a data history connection to a target instance with a custom Azure Database Explorer table name
```
digimaun, line 166: Another Database Explorer verbiage to fix. I won't point out further ones if they exist.

azext_iot/digitaltwins/_help.py
digimaun, line 155: It would be helpful if the command description specified the role assignments the CLI will apply in order for the service to successfully create the connection. The long-summary could be used for this.

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,112 @@ def load_digitaltwins_help():
...
     """
 
+    helps["dt data-history"] = """
+        type: group
+        short-summary: Manage and configure data history connections.
+    """
+
+    helps["dt data-history create"] = """
+        type: group
+        short-summary: Creates a data history connection between a Digital Twins instance and another resource.
```
digimaun, line 150: "Create data history connections between a Digital Twins instance and supported resources." 

The 'create adx' command description has a specific resource.

azext_iot/digitaltwins/commands_resource.py
```diff
@@ -250,3 +250,59 @@ def wait_private_endpoint_conn(cmd, name, conn_name, resource_group_name=None):
...
+
+
+def show_adx_data_connection(cmd, name, conn_name, resource_group_name=None):
+    rp = ResourceProvider(cmd)
+    return rp.get_adx_data_connection(name=name, conn_name=conn_name, resource_group_name=resource_group_name)
+
+
+def wait_adx_data_connection(cmd, name, conn_name, resource_group_name=None):
+    rp = ResourceProvider(cmd)
+    return rp.get_adx_data_connection(name=name, conn_name=conn_name, resource_group_name=resource_group_name, wait=True)
```
digimaun, line 298: Calling the wrong rp method?

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
     with self.argument_context("dt network private-endpoint connection wait") as context:
         context.ignore("created")
         context.ignore("exists")
+
+    with self.argument_context("dt data-history") as context:
+        context.argument(
+            "conn_name",
+            options_list=["--conn-name", "--cn"],
+            help="Name of time series database connection.",
```
digimaun, line 441: Name of the data history connection?

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+            arg_group="TimeSeries Connection",
```
digimaun, line 442: This arg_group should be removed, conn_name is a required fundamental arg for the command and its implied at the instance level. Separating into a different arg group muddies the presentation IMO.

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+            options_list=["--adx-cluster-name", "--adxc"],
+            help="Name of Azure Data Explorer cluster to integrate with.",
+            arg_group="Azure Database Explorer Arugments",
+        )
+
+        context.argument(
+            "adx_database_name",
+            options_list=["--adx-database-name", "--adxd"],
+            help="Name of Azure Data Explorer database to integrate with.",
+            arg_group="Azure Database Explorer Arugments",
```
digimaun, line 457: Arguments spelling issue, but more importantly that word should be removed as its already added by CLI core.

image

digimaun: `arg_group="Azure Data Explorer",`

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+            help="Name of time series database connection.",
+            arg_group="TimeSeries Connection",
+        )
+
+    with self.argument_context("dt data-history create adx") as context:
+        context.argument(
+            "adx_cluster_name",
+            options_list=["--adx-cluster-name", "--adxc"],
+            help="Name of Azure Data Explorer cluster to integrate with.",
+            arg_group="Azure Database Explorer Arugments",
```
digimaun, line 450: `arg_group="Azure Data Explorer",`

azext_iot/digitaltwins/_help.py
vilit1: rip me and adx

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+        )
+
+        context.argument(
+            "eh_consumer_group",
+            options_list=["--eventhub-consumer-group", "--ehc"],
+            help="The EventHub consumer group to use when ADX reads from EventHub.",
+            arg_group="Event Hub Arguments",
+        )
+
+        context.argument(
```
digimaun, line 504: It may make sense to re-use the dt endpoint eventhub arg definitions rather than having to [re-]define in multiple areas.

azext_iot/digitaltwins/commands_resource.py
vilit1: nope - it is the right one; wait needs the get (with a slight workaround for the error)

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+            options_list=["--eventhub-resource-group", "--ehg"],
+            help="Name of EventHub resource group. If not provided, will use the Digital Twin's resource group.",
+            arg_group="Event Hub Arguments",
+        )
+
+        context.argument(
+            "eh_subscription",
+            options_list=["--eventhub-subscription", "--ehs"],
+            help="Name or ID of subscription where the EventHub exists. If not provided, will use the subscription that contains"
+                 " the Digital Twin Instance..",
```
digimaun, line 515: Double period at end `..`

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,91 @@ def load_digitaltwins_arguments(self, _):
...
+                 "that contains the Digital Twin Instance.",
+            arg_group="Azure Database Explorer Arugments",
+        )
+
+        context.argument(
+            "adx_table_name",
+            options_list=["--adx-table-name", "--adxt"],
+            help="Name of Azure Data Explorer table to be created. If not provided, will use the format "
+                 "adt_dh_{dt_name}_{dt_location}.",
+            arg_group="Azure Database Explorer Arugments",
```
digimaun, line 480: remove Arugments from end of arg_group, and else where in this module

azext_iot/digitaltwins/providers/resource.py
```diff
@@ -495,3 +496,139 @@ def delete_private_endpoint_conn(self, name, conn_name, resource_group_name=None
...
+            eh_namespace=eh_namespace,
+            eh_entity_path=eh_entity_path,
+            eh_consumer_group=eh_consumer_group,
+            eh_resource_group=eh_resource_group,
+            eh_subscription=eh_subscription,
+            yes=yes,
+        )
+
+        try:
+            def check_state(lro):
```
digimaun, line 551: Revisit of DT RP issues with msrest? We should combine as much of the duplicate code as possible from our workaround in check_state used in the instance creation method.

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,290 @@
...
+            "/subscriptions/{}/resourceGroups/{}/providers/Microsoft.EventHub/namespaces/{}".format(
+                eh_subscription,
+                eh_resource_group,
+                eh_namespace,
+            )
+        )
+        logger.debug("Finished checking the EventHub resource.")
+
+        self.add_dt_role_assignment(
+            role="Azure Event Hubs Data Owner",
```
digimaun, line 101: Let's extract rbac role strings in a common area to be referenced....like the dt common.py module or elsewhere with similar intent.

azext_iot/sdk/digitaltwins/controlplane/models/connection_properties_private_link_service_connection_state_py3.py
```diff
@@ -13,14 +13,14 @@
...
-    """ConnectionPropertiesPrivateLinkServiceConnectionState.
+    """The connection state.
 
     All required parameters must be populated in order to send to Azure.
 
     :param status: Required. The status of a private endpoint connection.
      Possible values include: 'Pending', 'Approved', 'Rejected', 'Disconnected'
     :type status: str or
-     ~controlplane.models.PrivateLinkServiceConnectionStatus
+     ~azure.mgmt.digitaltwins.models.PrivateLinkServiceConnectionStatus
```
digimaun, line 23: Why describe a different root namespace? It's technically no longer accurate.

azext_iot/sdk/digitaltwins/controlplane/models/digital_twins_endpoint_resource.py
```diff
@@ -26,23 +26,28 @@ class DigitalTwinsEndpointResource(ExternalResource):
...
         'type': {'readonly': True},
+        'system_data': {'readonly': True},
         'properties': {'required': True},
     }
 
     _attribute_map = {
         'id': {'key': 'id', 'type': 'str'},
         'name': {'key': 'name', 'type': 'str'},
         'type': {'key': 'type', 'type': 'str'},
+        'system_data': {'key': 'systemData', 'type': 'SystemData'},
```
digimaun, line 50: Ideally we cover new property assertions as we integrate with them.

azext_iot/digitaltwins/params.py
vilit1: For endpoints, I would need to change eventhub-resource-group to be optional (and use the digital twin resource group if not provided)

azext_iot/sdk/digitaltwins/controlplane/models/connection_properties_private_link_service_connection_state_py3.py
vilit1: got generated like that - can manually fix

azext_iot/digitaltwins/params.py
digimaun: Ack. That would be a nice quality of life change, and is not breaking. But we should cover the other endpoint resource types as well.

azext_iot/sdk/digitaltwins/controlplane/models/connection_properties_private_link_service_connection_state_py3.py
digimaun: Ok. At least in prior autorest configurations you could specify the namespace.

azext_iot/digitaltwins/_help.py
vilit1: Personally, I prefer shortness but I can see where having the longer commands would be helpful. This will not go into the bug bash but I will ask if there is a preference for a different command naming convention that would align with the private endpoints.

azext_iot/digitaltwins/providers/resource.py
vilit1: Will write a generalized ms_rest check_state - depending on the resource, the provisioning state can be in the main level or nested in properties.

azext_iot/digitaltwins/_help.py
digimaun: It's not just about aligning with private endpoints, its about self-describing semantics for the command, and adhering to existing precedence and best practices for the IoT CLI.

Just take a look at other commands (scoped to DT commands for now) where the resource being operated on is explicit in the command signature

- az dt * // Root operations with respect to DT instances
- az dt endpoint * // Operations for DT instance endpoints
- az dt twin *     // Operations for twins on the instance
- az dt twin component * // Operations for twin components on the instance
- az dt network private-endpoint connection * // Operations for instance private-endpoint connections

Now we have az dt data-history * -- implying the DT instance support creating a data history? How about listing all data histories? What is an ADT data-history entity anyway?

_After_ reading the help its more clear, in all cases the 'az dt data-history *' commands are operating on ADT data history _connections_ integrating the instance with other resources through a 'connection' managed on the instance.

Yes you can make the command shorter, but shortness should be weighed against optimal semantics.

azext_iot/digitaltwins/providers/resource.py
digimaun: AFAIK, thus far we've only run into polling issues with ADT resources so it can be scoped to DT.

vilit1: There are two additional issues:
the error message will have to be more generic (something like "Please monitor the status of the created instance with the show command.")
the amount of polling for the dt instance creation vs the data history creation is very different (5 times for the DT create vs 20 times for data history)

azext_iot/digitaltwins/_help.py
```diff
@@ -140,6 +140,118 @@ def load_digitaltwins_help():
...
+
+    helps["dt data-history create"] = """
+        type: group
+        short-summary: Creates a data history connection between a Digital Twins instance and supported resources.
+    """
+
+    helps["dt data-history create adx"] = """
+        type: command
+        short-summary: Creates a data history connection between a Digital Twins instance and an Azure
+            Data Explorer Cluster. Requires pre-created Azure Data Explorer and Event Hub resources.
```
digimaun, line 156: nit, think Cluster should be lower case ....similar treatment here https://docs.microsoft.com/en-us/azure/data-explorer/create-cluster-database-portal?tabs=one-click-create-database

azext_iot/digitaltwins/common.py
```diff
@@ -5,16 +5,36 @@
...
+ERROR_PREFIX = "Could not create ADT instance connection. Unable to"
+FINISHED_CHECK_RESOURCE_LOG_MSG = "Finished checking the {0} resource."
+ADX_ROLE_MSG = "'Database Admin' permission on the Digital Twins instance for the Azure Data Explorer database '{0}'"
+RBAC_ROLE_MSG = "'{0}' role on the Digital Twins instance for the scope '{1}'"
+TRY_ADD_ROLE_LOG_MSG = "Trying to add the {0}."
+PRESENT_ADD_ROLE_LOG_MSG = "The {0} is already present."
+FINISHED_ADD_ROLE_LOG_MSG = "Finished adding the {0}."
+SKIP_MSG = "Skipping addition of the {0}. This may prevent creation of the data history connection."
+FAIL_GENERIC_MSG = ERROR_PREFIX + " assign {0}. Please assign this role manually."
+FAIL_RBAC_MSG = ERROR_PREFIX + " assign {0}. Please assign this role manuallywith the command `az {1}`."
```
digimaun, line 31: need space  "manuallywith"

azext_iot/digitaltwins/common.py
```diff
@@ -5,16 +5,36 @@
...
+ABORT_MSG = "Command was aborted."
+ADD_INPUT_MSG = "Add the {0}?"
+CONT_INPUT_MSG = "Continue with Data History Connection create anyway?"
```
digimaun, line 34: what about "Continue with data history connection creation anyway?"

azext_iot/digitaltwins/params.py
```diff
@@ -433,3 +433,90 @@ def load_digitaltwins_arguments(self, _):
...
+        context.argument(
+            "eh_subscription",
+            options_list=["--eventhub-subscription", "--ehs"],
+            help="Name or ID of subscription where the EventHub exists. If not provided, will use the subscription that contains"
+                 " the Digital Twin Instance.",
+            arg_group="Event Hub",
+        )
+
+        context.argument(
+            'yes',
```
digimaun, line 519: nit, can these single quote strings be dbl quote strings so we are consistent in this module

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,285 @@
...
+
+    def validate_eventhub(
+        self,
+        eh_namespace: str,
+        eh_entity_path: str,
+        eh_resource_group: str,
+        eh_subscription: str,
+        eh_consumer_group: str
+    ):
+        from azext_iot.digitaltwins.providers.endpoint.builders import EventHubEndpointBuilder
```
digimaun, line 81: nice reuse of endpoint builders module

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,285 @@
...
+                eh_namespace,
+                eh_resource_group,
+            ),
+            subscription=eh_subscription,
+        )
+        if not eh_consumer_group_op.success():
+            raise CLIError("{} retrieve Event Hub Consumer Group.".format(ERROR_PREFIX))
+
+        self.eh_namespace_resource_id = (
+            "/subscriptions/{}/resourceGroups/{}/providers/Microsoft.EventHub/namespaces/{}".format(
```
digimaun, line 105: Ideally the endpoint builder supports providing the fully qualified Id as it already has all the information.

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,285 @@
...
+
+        current_roles_op = self.cli.invoke(
+            "role assignment list --role '{}' --assignee {} --scope {}".format(
+                role,
+                self.dt.identity.principal_id,
+                resource_id
+            )
+        )
+
+        if current_roles_op.success() and len(current_roles_op.as_json()) > 1:
```
digimaun, line 180: For the second condition is it supposed to be `>= 1`?

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,285 @@
...
+
+            database_admin_list_op = self.cli.invoke(
+                "rest --method POST --url {}/databases/{}/listPrincipals?{}".format(
+                    self.adx_resource_id,
+                    adx_database_name,
+                    api_version,
+                )
+            )
+            if not database_admin_list_op.success():
+                raise database_admin_list_op.az_cli.result.error
```
digimaun, line 216: Is this raising SystemExit or are there more KPIs than that?

azext_iot/digitaltwins/providers/connection/builders.py
vilit1: would require code changes in the endpoint builder but the dt endpoint api has no need for the resource id. Only the data history requires the resource ids

azext_iot/digitaltwins/providers/connection/builders.py
vilit1: assuming that the rbac api isnt wonky, there should be only 0 items or 1 item returned from the call since it is looking for the specific resource, scope and role. There isn't any include inherited so I am not sure where a duplicate can come from.

But I can change it so we would be on the safe side (as do not trust the rbac api)

azext_iot/digitaltwins/providers/connection/builders.py
vilit1: This was a fun exploration I did to figure out where the error was in an embedded cli call since it wraps up the error.

azext_iot/constants.py
vilit1: Will need to rerun tests on 2/8

azext_iot/digitaltwins/providers/connection/builders.py
digimaun: Discussed offline.

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -0,0 +1,285 @@
...
+    FAIL_GENERIC_MSG,
+    ABORT_MSG,
+    ADD_INPUT_MSG,
+    CONT_INPUT_MSG
+)
+
+logger = get_logger(__name__)
+
+
+class AdxConnectionValidator(ABC):
```
digimaun, line 34: This is a full concrete class implementation, what is the purpose of deriving from ABC? This can be a vanilla class.

azext_iot/digitaltwins/providers/connection/builders.py
vilit1: made a task for the future

azext_iot/digitaltwins/params.py
vilit1: made a task for the future

azext_iot/digitaltwins/_help.py
vilit1: done

azext_iot/tests/digitaltwins/__init__.py
```diff
@@ -172,6 +183,92 @@ def get_available_region(self, capacity: int = 1, skip_regions: list = None) ->
...
+                {
+                    "location": "eastus",
+                    "kind": "ReadWrite",
+                    "properties": {
+                        "softDeletePeriod": "P1D"
+                    }
+                }
+            )
+            database_creation_state = self.embedded_cli.invoke(
+                "rest --method PUT --url 'https://management.azure.com/subscriptions/{}/resourceGroups/"
```
c-ryan-k, line 246: Just a note - these URLs are hard-coded for the default ARM endpoint `management.azure.com` but we might (possibly at some point) want to run tests against different control-plane endpoints.

azext_iot/digitaltwins/command_map.py
```diff
@@ -45,6 +45,31 @@ def load_digitaltwins_commands(self, _):
...
+        "dt data-history connection",
+        command_type=digitaltwins_resource_ops,
+        is_preview=True
+    ) as cmd_group:
+        cmd_group.show_command("show", "show_adx_data_connection")
+        cmd_group.wait_command("wait", "wait_adx_data_connection")
+        cmd_group.command("list", "list_adx_data_connection")
+        cmd_group.command(
+            "delete", "delete_adx_data_connection", confirmation=True, supports_no_wait=True
+        )
```
c-ryan-k, line 65: small nit - if other types of data connections are added, we may want to move these into a `show_data_connection` method, and with `list` as an example - maybe add a `--type adx` param as a possible filter

azext_iot/digitaltwins/common.py
```diff
@@ -5,16 +5,36 @@
...
+
+# Data History strings
+DT_IDENTITY_ERROR = "Digital Twins instance does not have System-Assigned Identity enabled. Please enable and try again."
+ERROR_PREFIX = "Unable to"
+FINISHED_CHECK_RESOURCE_LOG_MSG = "Finished checking the {0} resource."
+ADX_ROLE_MSG = "'Database Admin' permission on the Digital Twins instance for the Azure Data Explorer database '{0}'"
+RBAC_ROLE_MSG = "'{0}' role on the Digital Twins instance for the scope '{1}'"
+TRY_ADD_ROLE_LOG_MSG = "Trying to add the {0}."
+PRESENT_ADD_ROLE_LOG_MSG = "The {0} is already present."
+FINISHED_ADD_ROLE_LOG_MSG = "Finished adding the {0}."
```
c-ryan-k, line 28: I like adding output format strings here - this will likely save us some headache finding and fixing messages in the future.

Using `Trying to add the {0}.` as an example - what is the expected format of the input? Maybe we could have a comment for the less obvious messages explaining how it should be used.

azext_iot/digitaltwins/command_map.py
digimaun: Everything but 'create [connection type]', is meant to generically support data history connections, so agreed the method callback names should not be specific for adx.

Filter wise the built in jmespath --query support _should_ work out of the box (reducing criticality of an explicit filter param/arg) but the net benefit is low because today one kind of connection (adx) can be made.

azext_iot/digitaltwins/command_map.py
```diff
@@ -45,6 +45,31 @@ def load_digitaltwins_commands(self, _):
...
+        "dt data-history",
+        command_type=digitaltwins_resource_ops,
+        is_preview=True
+    ) as cmd_group:
+        pass
+
+    with self.command_group(
+        "dt data-history connection",
+        command_type=digitaltwins_resource_ops,
+        is_preview=True
```
digimaun, line 58: Since the parent `dt data-history`  has `is_preview=True`, we shouldn't need to re-declare it here. We should test/ensure this statement, then could remove it here.

azext_iot/tests/digitaltwins/__init__.py
vilit1: Will remove the ADX resource generation until the az kusto extension is no longer preview. This should solve the issue of these hardcoded urls

digimaun: Or you could drop the host, just like how the az rest commands in the implementation are (or should be) formatted -- in this case the CLI assumes the url as an Azure resource and prefixes the value with the ARM endpoint of the current cloud .

azext_iot/digitaltwins/command_map.py
vilit1: image

azext_iot/digitaltwins/command_map.py
vilit1: Also there is a general theme of one connection per digital twin so far. I have list in case this changes in the future.

As paymaun said, there is little need for adding a type filter right now.

azext_iot/digitaltwins/common.py
vilit1: Added some instructions and an example since these strings are supposed to be built up using earlier message strings.

vilit1:
running all the unit and int tests:
image

vilit1:
image
first is if adx cluster exists, second is if adx cluster is created

vilit1:
image

avagraw:
As discussed, lets add the pipeline variables to support the data history tests.

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -0,0 +1,481 @@
...
+                ADX_CLUSTER,
+                EP_EVENTHUB_NAMESPACE,
+                EP_EVENTHUB_TOPIC,
+                EP_RG,
+                "testresource",
+            ),
+            expect_failure=True
+        )
+
+    def test_dt_data_history_adx_wait(self):
```
avagraw, line 331: Separating out this test into another test script, will allow the test to be executed in parallel with the test_dt_data_history_adx test. This should significantly reduce the total time to complete the tests.

avagraw (changes requested):

.azure-devops/create-release.yml
```diff
@@ -6,7 +6,7 @@ trigger: none
 parameters:
 - name: variableGroup
   type: string
-  default: 'aziotcli_test_primary'
+  default: 'vilit-test-variables'
```
avagraw, line 9: Placeholder to revert this back when you are done with testing.

avagraw (changes requested):

.azure-devops/templates/set-testenv-sentinel.yml
```diff
@@ -13,7 +13,7 @@ steps:
             "AZURE_TEST_RUN_LIVE":True,
             "azext_iot_testrg": os.environ.get("AZEXT_IOT_TESTRG", sentinel_value),
             "azext_iot_testhub": os.environ.get("AZEXT_IOT_TESTHUB", sentinel_value),
```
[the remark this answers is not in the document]
vilit1: Added unless there is another yaml file to edit

.azure-devops/create-release.yml
digimaun: Also the parameter is meant to be overridable, you don't need a code change to use a different value. Then re-running ( "Run New") from a prior run retains parameter value changes.

azext_iot/sdk/digitaltwins/dataplane/operations/digital_twins_operations.py
```diff
@@ -947,7 +953,7 @@ def internal_paging(next_link=None, raw=False):
     list_incoming_relationships.metadata = {'url': '/digitaltwins/{id}/incomingrelationships'}
 
     def send_telemetry(
-            self, telemetry, id, message_id, telemetry_source_time=None, digital_twins_send_telemetry_options=None, custom_headers=None, raw=False, **operation_config):
+            self, id, telemetry, message_id, telemetry_source_time=None, digital_twins_send_telemetry_options=None, custom_headers=None, raw=False, **operation_config):
```
digimaun, line 956: telemetry_source_time and digital_twins_send_telemetry_options need to be integrated for the corresponding command here and for send_component_telemetry (and dt_timestamp removed). Looks like it was missed in a prior SDK refresh (the original parameter was dt_timestamp).

HISTORY.rst
```diff
@@ -3,6 +3,17 @@
 Release History
 ===============
 
+
+0.13.1
```
digimaun, line 7: These changes should be under `0.14.0`. Also please add line item(s) for the API/sdk refresh (API version being pointed at is a user facing change) and the to be done addition of telemetry options for `az dt twin telemetry send`.

azext_iot/sdk/digitaltwins/dataplane/operations/digital_twins_operations.py
digimaun: The user facing param could be something like:

`--telemetry-source-time`: `An RFC 3339 timestamp that identifies the time the telemetry was measured. If not provided the client will generate a UTC timestamp of when the request was processed.`

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
vilit1: Will keep in the same script for now

HISTORY.rst
vilit1: done

.azure-devops/create-release.yml
vilit1: done

azext_iot/constants.py
vilit1: changed to final values

azext_iot/sdk/digitaltwins/dataplane/operations/digital_twins_operations.py
vilit1: done - also made it so that the client will not generate the timestamp. That way the service can decide what to do if there is no telemetry source time (like how it was doing that before)

HISTORY.rst
```diff
@@ -3,6 +3,22 @@
 Release History
 ===============
 
+
+0.14.0
++++++++++++++++
+
+**Digital Twin updates**
+
+* Added `az dt data-history` command group, which will allow users to configure a data history connection
```
digimaun, line 12: We'll have to adjust some things to get these changes merged for a near-term release.  Like not exposing the `az dt data-history` to the user.

azext_iot/digitaltwins/command_map.py
```diff
@@ -45,6 +45,30 @@ def load_digitaltwins_commands(self, _):
         cmd_group.wait_command("wait", "wait_instance")
         cmd_group.command("reset", "reset_instance", confirmation=True, is_preview=True)
 
+    with self.command_group(
```
digimaun, line 48: For near-term release lets comment out or disable `az dt data-history`

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -0,0 +1,484 @@
...
+    generate_generic_id
+)
+
+logger = get_logger(__name__)
+
+
+ADH_REGION = "northeurope"
+
+
+class TestDTConnections(DTLiveScenarioTest):
```
digimaun, line 29: For near-term release lets skip the data history connection ITs.

azext_iot/digitaltwins/providers/generic.py
```diff
@@ -47,3 +52,35 @@ def remove_prefix(text, prefix):
     if text.startswith(prefix):
         return text[len(prefix) :]
     return text
+
+
+def generic_check_state(lro, show_cmd, max_retires):
```
digimaun, line 57: sp - max_retries?

azext_iot/digitaltwins/command_map.py
digimaun: We should think about a feature flag pattern in the future (using an env var(s) for configuration).

azext_iot/digitaltwins/command_map.py
```diff
@@ -30,6 +30,9 @@
 )
 
 
+enabled_data_history = False
```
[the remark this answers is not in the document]
digimaun: The point of the flag is we enable all elements (commands, help, tests) if the flag is set

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=4981&view=results

vilit1:
image

azext_iot/digitaltwins/command_map.py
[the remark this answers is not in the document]
vilit1: done

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -0,0 +1,485 @@
...
+    EP_EVENTHUB_TOPIC,
+    generate_generic_id
+)
+
+logger = get_logger(__name__)
+ADH_REGION = "northeurope"
+enabled_data_history = os.environ.get("ADT_DH_PREVIEW_ENABLE")
+
+
+@pytest.mark.skipif(not enabled_data_history, "Wait until Data History Public Preview")
```
digimaun, line 29: `@pytest.mark.skipif(not enabled_data_history, reason="Wait until Data History Public Preview")`

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=4988&view=results

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -0,0 +1,485 @@
...
+        )
+
+        list_result = self.cmd(
+            "dt data-history connection list -n {} -g {}".format(
+                instance_name, self.rg
+            )
+        ).get_output_in_json()
+        assert len(list_result) == 0
+
+    def test_dt_data_history_adx_create_incorrect_resource(self):
```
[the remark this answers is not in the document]
digimaun: Chatted more offline.

digimaun (approved):

Issue #475, opened 2021-12-13
[bug] monitor-events sys.excepthook issue when using keyboard interrupt (ctrl-c) on a Python 3.10 runtime


digimaun:
**Describe the bug**
At least on a Python 3.10 runtime using `az iot hub monitor-events`, sys.excepthook errors are shown on keyboard interrupt. This was observed on a homebrew based install.

**To Reproduce**

- Install the azure-iot extension (`0.12.0`, `0.12.1`) on a Py 3.10.x based Azure CLI install. Run monitor-events and input keyboard interrupt. Observe errors.

**Expected behavior**
- Gracefully closing and deallocating resources. 

**Screenshots**

**Environment (please complete the following information):**

- OS: `MacOS`
- Shell: `bash`
- Az CLI version: `2.31.0`
- IoT extension version: `0.12.0`, `0.12.1`
- Python version (if pip installed):  `Python 3.10.x`

lucadruda:
confirming for 3.9 as well on WSL2

vilit1:
Closing since this was resolved.

PR #498, opened 2022-03-07
Generate resources for Central tests

vilit1:
Automate resource generation for central tests. If the resource values are provided in the configuration file, they will be used. Otherwise, new resources will be created to run the tests. Within a given testing resource group, 
 - a storage account will be created to be used for all the central tests
 - a separate central app will be created for each test and retrieving all possible central app values
 - the resources will be deleted once the tests are done.

Specifically, these configuration variables can be safely removed:
 -  azext_iot_central_app_id
 -  azext_iot_central_primarykey
 -  azext_iot_central_scope_id
 -  azext_iot_central_token
 -  azext_iot_central_dns_suffix
 -  azext_iot_central_storage_cstring
 -  azext_iot_central_storage_container

### Generated variables
If the central app id is not provided, the resource group to create the central app must be provided with:
 - azext_iot_testrg

If a central app is provided, the resource group value will be retrieved in case it is needed for storage account creation.
Then, the following variables will be populated:
 - azext_iot_central_app_id during central app creation
 - azext_iot_central_scope_id by creating a throwaway device (in get_app_scope_id)
 - azext_iot_central_storage_cstring and azext_iot_central_storage_container during storage account creation.

### Limitations and non-generated variables
Note that with the current commands, two parameters will still be needed to run all tests:
 - azext_iot_central_primarykey
 - azext_iot_central_app_id

If one of these parameters is not provided, then this test will be skipped:
- test_device_connect

If only azext_iot_central_primarykey is provided, then the primary key value will be nulled out because the newly created central app would not have the same primary key as the one that is defined in the configuration file.

For internal testing, these parameters will only be populated (and used) if present in the pytest.ini configuration file:
 - azext_iot_central_dns_suffix
 - azext_iot_central_token

#### Delete device templates
Changed the _delete_device_template code to invoke the correct command each time. There was an issue in which the list command defined after was invoked instead of the delete command.

#### sys.excepthook error
In addition, fixing sys.excepthook error by specifying which loop is to be closed.

The issue is here: https://github.com/Azure/azure-iot-cli-extension/issues/475

---

### General Guidelines

Intent for Production

Basic expectations

~- [x] Have you made an entry in HISTORY.rst which concisely explains your user-facing feature or change?~


vilit1:
image

azext_iot/tests/central/__init__.py
```diff
@@ -33,18 +47,75 @@
...
+            self.cmd(
+                "iot central app create -n {} -g {} -s {} --mi-system-assigned -l {}".format(
+                    self.app_id,
+                    self.app_rg,
+                    self.app_id,
+                    "westus"
+                ),
+                include_opt_args=False
+            )
+            self.app_primary_key = None
```
yang-hai-feng, line 105: Once app was created, the self.dns_suffix should be the fixed with "azureiotcentral.com", self.token should not be read from env as well.

yang-hai-feng: clean self._scope_id as well.

yang-hai-feng (commented):
Looks good to me, put some comments on creating new app flow, btw have you run the integration tests successfully with auto created apps?

azext_iot/tests/central/__init__.py
lucadruda: fixing to "azureiotcentral.com" prevents internal development against dev clusters. This is useful when working on changes that require backend features not yet available in Central PROD

lucadruda: unless we assume internal dev will always have an existing app and can go with variables in pytest

lucadruda (commented):
Looks good to me

azext_iot/tests/central/__init__.py
yang-hai-feng: not so sure if ```az iot central app create``` can create an PPE app which had "azureiotcentral-ppe.com", in create app flow, I feel this should be picked up from the app, because we had mooncake app as well.

vilit1: dns_suffix defaults to azureiotcentral.com already for commands so there is no need to set it 
image

Also will fix the help for this param...

How do you currently create a ppe app or mooncake app? Is it possible to map a location/subscription to a ppe environment (ex: creating an app in X location in Y subscription will always create a ppe app)?
If so, is it possible to set default tokens/scope_id/other variables? For example, all apps created in ppe environment share the scope id? I want to make sure that token, dns_suffix, and scope_id can be safely nulled out if a new app is created.

vilit1: Also regarding the test runs, see the screenshots above - they include time differences between generating apps and not generating apps.

vilit1: updated the tests to null out scope_id too so the scope_id could be taken from the newly created app.  WIll not null out dns_suffix or token for now.

yang-hai-feng: Sounds good to me.

yang-hai-feng (approved):

azext_iot/tests/central/test_iot_central_int.py
```diff
@@ -40,6 +34,17 @@ def fixture_api_version(self, request):
             print("Testing 1.1-preview")
         yield
 
+    @pytest.fixture(scope='class', autouse=True)
+    def setUpSuite(self):
+        self._create_storage_account()
+        yield None
```
digimaun, line 40: Why `yield None` here? Can be removed.

digimaun (approved):

azext_iot/tests/central/__init__.py
vilit1: added token tests

vilit1:
I removed the functionality in the integration tests that would run the tests twice (once with a token and once without). This meant that tests took twice as long.

Instead, the integration tests will rely on the configuration to determine if they are going to be run with a token or not; if there is a token defined then the integration tests will use that token and if there is no token, the integration tests will run without a token. 

In a future pr, I will add unit tests to ensure that the tokens are generated properly. Specifically the unit tests will cover the functions in the central utility.

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5098&view=results

digimaun (approved):

yang-hai-feng (approved):

PR #505, opened 2022-04-04
Test updates

vilit1:
Some improvements for testing:
1. Test documentation updates (please read carefully)
2. lower max for DT instances in a region to 9 to avoid max resource conflicts
3. Change the central tests to use the same storage account as the iot hub ones. Now, the storage account name can be used instead of the connection string. By default the containers will be different for both: devices for Iot Hub and central for Central.
4. Add tagging to Iot Hub, DPS, and central resources. Now these resources will have a test tag to indicate number of runs for a test.
image

5. Name changes to generated Iot Hubs + extra tagging to help differentiate the hubs generated for iot hub tests vs dps tests
image

6. Fix storage deletion post-test.

Note: tagging has not been added to dt instances since there are integration tests that rely on set tags for dt instances. 

Example of generated test resources:
image

---

### General Guidelines

Intent for Production

Basic expectations


CONTRIBUTING.md
```diff
@@ -87,50 +87,139 @@ You may need to install the dev_requirements for this
...
+`azext_dt_ep_eventhub_topic_consumer_group` is the Service Bus Namespace to use. Optional variable, used for Digital Twin data history tests.
+
+`azext_dt_ep_rg` is the resource group that contains the endpoint (Event Hub, Event Grid, Service Bus) variables. Optional variable, if not provided, `az_iot_testrg` will be used as the resource group.
+
 ##### IoT Hub
 
 Execute the following command to run the IoT Hub integration tests:
 
-`pytest azext_iot/tests/iothub/ -k "_int"`
+`pytest azext_iot/tests/iothub/ -k "_int.py"`
```
digimaun, line 222: The `.py` suffix shouldn't be strictly necessary to run the IT's.

vilit1: without the ".py", some unit tests get caught because they have "_int" in the test name (although this only happens for central tests, I got used to adding the py):
image

If we say that the keyword "_int" must only match unit tests, it will restrict naming unit tests so I prefer keeping the extension.

digimaun: Ok - the line I commented on has a path constraint for iothub, which doesn't have this issue (like you pointed out its currently an issue with central tests).

Then why include the extension path _only_ for a service-test area that doesn't have the issue? Further down there is still usage of `_int`

image

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
-_DPS:_  
-`pytest azext_iot/tests/dps/test_iot_dps_unit.py`
+_Hub:_
+`pytest azext_iot/tests/iothub/core/test_iot_ext_unit.py`
+
+_DPS:_
+`pytest azext_iot/tests/dps/core/test_dps_discovery_unit.py`
+
+Unit tests end in "_unit" so execute the following command to run all unit tests,
+`pytest -k "_unit"`
```
digimaun, line 99: Should the unit test examples not have the file extension based on the concern raised on the integration test examples? I think they should be consistent.

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
 
-1. Create IoT Hub
-
-    > IMPORTANT: Your IoT Hub must be created specifically for integration tests and must not contain any devices when the tests are run.
+The following resources will be needed for the integration tests.
 
-2. Create Files Storage - In IoT Hub, click Files, create a new Storage Account and link to an empty Container.
+- IoT Hub
+- IoT Device Provisioning Service
+- IoT Digital Twin
```
digimaun, line 115:
```suggestion
- Azure Digital Twins instance
```
The official marketing for the Azure Digital Twins service usually doesn't have an IoT prefix https://docs.microsoft.com/en-us/azure/digital-twins/overview

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
+- Event Grid Topic
+- Event Hub Namespace with an Event Hub
+- Service Bus Namespace with a Topic
+- Azure Data Explorer Cluster with a Database
 
-3. Create IoT Hub Device Provisioning Service (DPS)
+If specified in the pytest.ini configuration file, those resources will be used. Please ensure that the resources are in a clean, new state (ex: Iot Hub should not have any devices). Otherwise, new resources will be generated during the test startup and deleted during the test teardown.
 
-4. Link IoT Hub to DPS - From DPS, click "Linked IoT Hub" and link the IoT Hub you just created.
+Note that if you stop the code with ctrl + C, the resources may not be deleted properly.
```
digimaun, line 125:
```suggestion
> Note: If you interrupt test execution (for example via ctrl + C), the teardown or clean up processes may not run leaving resources in an indeterminant state.
```

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
+
+To run all tests, `azext_iot_testrg`, `azext_iot_central_app_id`, and `azext_iot_central_primarykey` are needed because the IoT Central Primary Key cannot be currently retrieved through the CLI.
+
+For all resources, if the environmental variable is not provided, a new instance will be created for the test run and deleted at the end of the test run.
+
+##### General Variables
+
+`az_iot_testrg` is the resource group that contains the IoT Hub and DPS instances or where all test resources are created. This is required, as it will be the default resource group if any other resource group variables are not provided.
+
+`azext_iot_testhub` is the test IoT Hub name. Optional variable, used for IoT Hub and DPS tests. If not provided, a new IoT Hub will be created for the test run (and deleted at the end of the test).
```
digimaun, line 168: `azext_iot_testhub` should probably be in an 'IoT Hub test variables' category

`azext_iot_testdps` should probably be in an 'IoT Device Provisioning Service' category

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
+
+`azext_iot_testdps` is the test DPS name. Optional variable used for IoT DPS tests. If not provided, a new DPS instance will be created for the test run (and deleted at the end of the test).
+
+`azext_iot_teststorageaccount` is the storage account used for running IoT Hub and Central storage tests. Optional variable, specify only when you want to run storage tests. During these tests, your hub will be assigned a System-Assigned AAD identity, and will be granted the role of "Storage Blob Data Contributor" on the storage account you provide. Both the hub's identity and the RBAC role will be removed once the test completes. No role assignments are made for the IoT Central App.
 
 `azext_iot_teststoragecontainer` is the name of blob container belonging to the above mentioned storage account. Optional environment variable, defaults to 'devices' when not specified.
 
+##### IoT Central Test variables
```
digimaun, line 176: What about a markdown table to organize the vars for a sharper presentation? Columns could be  corresponding to the var  name, required and description.

CONTRIBUTING.md
```diff
@@ -87,69 +87,158 @@ You may need to install the dev_requirements for this
...
 
-`pytest azext_iot/tests/iothub/ -k "_int"`
+`pytest azext_iot/tests/iothub/ -k "_int.py"`
 
 ##### Device Provisioning Service
 
 Execute the following command to run the IoT Hub DPS integration tests:
 
-`pytest azext_iot/tests/dps/ -k "_int"`
+`pytest azext_iot/tests/dps/ -k "_int.py"`
```
digimaun, line 228: Thoughts about combining the 'executing the service tests' sections with the 'setting the service specific env vars to run tests' sections?

vilit1: I reformatted how the tests are presented. Please look over the changes

vilit1:
Have been running many tests so far on the pipeline: https://dev.azure.com/azureiotdevxp/aziotcli/_build?definitionId=81&_a=summary

There seem to still be service hiccups:

https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5376&view=logs&j=1e6b4794-1f9a-54f4-f473-41f4c9a18170

vilit1:
Note that Data History Tests are temporarily disabled as per request by the service team

vilit1:
Current test runs:
Passing hub tests, dt tests (and almost all dps tests):
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5379&view=results

azext_iot/tests/iothub/__init__.py
```diff
@@ -186,6 +182,92 @@ def generate_job_names(self, count=1):
...
+                ),
+            )
+
+    def _add_test_tag(self, test_tag):
+        tags = self.cmd(
+            "iot hub show -n {} -g {}".format(self.entity_name, self.entity_rg)
+        ).get_output_in_json()["tags"]
+
+        if tags.get(test_tag):
+            tags[test_tag] = int(tags[test_tag]) + 1
```
digimaun, line 258: What are the scenarios when the same test case is run twice where this increment happens?

azext_iot/tests/iothub/__init__.py
```diff
@@ -186,6 +182,92 @@ def generate_job_names(self, count=1):
...
+            )
+
+        elif not settings.env.azext_iot_teststorageaccount:
+            self.cmd(
+                "storage container delete -n {} --connection-string '{}'".format(
+                    self.storage_account_name, self.storage_cstring
+                ),
+            )
+
+    def _add_test_tag(self, test_tag):
```
digimaun, line 252: I'm seeing multiple similar implementations for this, can we refactor for more re-use? `az resource tag` can be used to apply tags generically to ARM resources.

azext_iot/tests/iothub/__init__.py
vilit1: Two major scenarios:
1. test reruns (in the pipeline)
2. providing your own resource

Example of central resource
image

digimaun: Isn't this the same single scenario? When a test is re-run against an existing resource? If the pipeline is re-run when no prior created resource is used it should provision new ones.

Can you please explain how adding tags incrementing the number of times a test case is run against an resource helps debugging issues? Is there anything being asserted? Is there a strong correlation between an increment and a pipeline run?  The pipeline logs already show you what tests were run & when....a stronger correlation between the pipeline and the resource may be tagging the pipeline run Id.

azext_iot/tests/digitaltwins/__init__.py
```diff
@@ -269,7 +285,8 @@ def ensure_adx_resource(self):
                 retries += 1
 
             if retries == MAX_ADX_RETRIES:
-                logger.error("Waited 2.5 minutes for database creation and database has not been created yet.")
+                database_msg = f"Waited 2.5 minutes for database creation and database {ADX_DATABASE} has not been created yet."
```
digimaun, line 288: The time waited can be a variable/dynamically calcualted so it doesn't have to be manually updated if the retries/sleep intervals change.

Section higher up with the same suggestion.

azext_iot/tests/iothub/__init__.py
```diff
@@ -186,6 +182,92 @@ def generate_job_names(self, count=1):
...
+        Delete the storage account if it was created.
+        """
+        if not settings.env.azext_iot_teststorageaccount:
+            self.cmd(
+                "storage account delete -n {} -g {} -y".format(
+                    self.storage_account_name, self.entity_rg
+                ),
+            )
+
+        elif not settings.env.azext_iot_teststorageaccount:
```
digimaun, line 245: Same condition as the if on line 238

azext_iot/tests/iothub/__init__.py
```diff
@@ -186,6 +182,92 @@ def generate_job_names(self, count=1):
...
+            self._populate_storage_cstring()
+
+        # This won't overwrite or error if container is already created.
+        self.cmd(
+            "storage container create -n {} --connection-string '{}'".format(
+                self.storage_container, self.storage_cstring
+            ),
+        )
+
+    def _populate_storage_cstring(self):
```
digimaun, line 223: Rather than similar/duplicate implementations for provisioning storage accounts can we refactor for re-usability?

azext_iot/tests/test_constants.py
```diff
@@ -4,6 +4,17 @@
...
 # --------------------------------------------------------------------------------------------
 
+from enum import Enum
+
+
+class ResourceTypes(Enum):
+    """
+    Resource types to use with az resource
+    """
+    central = "Microsoft.IoTCentral/IoTApps"
```
digimaun, line 14: You can also use the fully qualified resource Id which already has the resource type.

azext_iot/tests/iothub/__init__.py
vilit1: The tagging should be an extra way to map resource to test in a pipeline. The current set up does not spit out information if a test run is successful and even when it does, it can take a while to find the generated resource name (and sometimes it's not possible to find all generated names). Specifically, this helped with dps test runs where the hubs were in activatedfailed states and the tests did not fail but the hubs were not deleted.

It helps with identifying testing bugs - like when a resource is not deleted correctly (b/c service issue like the activatedfailed hubs or test got cancelled/stopped or some other funky issue). The tests do check that if the autogenerated name is already used then that resource will be used instead. In a perfect world all the test runs should be 1 but I have seen some tests go to two for some reason in a single generated resource- give me a bit to see if I can still recreate.

Having the pipeline log as a tag is a great idea - it would require an extra optional variable. Going to do that. Another idea is to also add the python version in (since the pipelines run 4 times) - but I am not sure on how to do that yet.

azext_iot/tests/iothub/__init__.py
vilit1: done

azext_iot/tests/test_constants.py
vilit1: I was going to use that but then I needed an extra call to get the resource id of existing resources (you can get the id from any create call if the resource is being generated but if the resource exists, we do not do those calls/check if the resource exists...) 

I figured since all test classes have the name + rg and the types _should_ not change, this would be a suitable alternative. Unless you think we should have a show call when the resource is provided.

azext_iot/tests/iothub/__init__.py
digimaun: I also thought about the python version, but applying specifically the pipeline job Id as a tag would already link the resource to a Python version since the IT jobs are organized by Python version.

azext_iot/tests/test_constants.py
digimaun: Ok, the reasoning makes sense, I'm fine with keeping this.

avagraw (approved):

CONTRIBUTING.md
```diff
@@ -87,69 +95,149 @@ You may need to install the dev_requirements for this
...
-1. Create IoT Hub
+_Hub:_
+`pytest azext_iot/tests/iothub/core/test_iothub_storage_int.py`
 
-    > IMPORTANT: Your IoT Hub must be created specifically for integration tests and must not contain any devices when the tests are run.
+_DPS:_
+`pytest azext_iot/tests/dps/core/test_dps_discovery_int.py`
+
+Integration tests end in "_int_.py" so execute the following command to run all integration tests,
+`pytest -k "_int_.py"`
```
digimaun, line 126: \_int_.py?

CONTRIBUTING.md
```diff
@@ -87,69 +95,149 @@ You may need to install the dev_requirements for this
...
 
-Execute the following command to run the IoT Hub DPS integration tests:
+If the "--api-version" argument is not specified, all runs act against default api version for each tested command.
 
-`pytest azext_iot/tests/dps/ -k "_int"`
+##### Test Resource Tagging
 
-##### IoT Central
-Execute the following command to run the IoT Central integration tests:
+There are 4 more test variables used for tagging the test resources:
```
digimaun, line 227: Not including these test variables in the IT variables table or list above?

azext_iot/tests/central/__init__.py
```diff
@@ -605,16 +632,21 @@ def _populate_storage_cstring(self):
...
         """
-        if self.storage_account_name:
+        if not settings.env.azext_iot_teststorageaccount:
             self.cmd(
                 "storage account delete -n {} -g {} -y".format(
                     self.storage_account_name, self.app_rg
                 ),
                 include_opt_args=False,
             )
+        elif not settings.env.azext_iot_teststorageaccount:
```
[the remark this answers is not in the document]
digimaun: In this updated state line 605 & 612 have the same condition.

azext_iot/tests/central/__init__.py
```diff
@@ -58,9 +55,13 @@ class CentralLiveScenarioTest(CaptureOutputLiveScenarioTest):
     def __init__(self, test_scenario):
         super(CentralLiveScenarioTest, self).__init__(test_scenario)
         self._create_app()
```
digimaun, line 57: self._create_app() may set variables like the token, dns_suffix and the app primary key , however later in initialization (like 65, 66) self.token and self.dns_suffix get set to `None`, preventing the respective test configuration.

CC: @gtrifonov

CONTRIBUTING.md
vilit1: I am trying to show that these are separate and would be used by internal developers (as in us)

azext_iot/tests/central/__init__.py
vilit1: fixe - going to fix the description for the functions too

azext_iot/tests/central/__init__.py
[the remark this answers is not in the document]
vilit1: I do not know how this keeps slipping through

CONTRIBUTING.md
vilit1: updated

vilit1:
Current test runs:
DPS + IoT Hub
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5430&view=results

DT:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5432&view=results

Central:

Smoke tests:

CONTRIBUTING.md
```diff
@@ -87,69 +95,153 @@ You may need to install the dev_requirements for this
...
+| `job_display_name`	| IoT Hub, DPS, and Central Tagging	| Job Display Name populated by an internal pipeline run. Can be manually set to customize the pipeline name tag. Please see "Test Resource Tagging" for more details.	|
+| `job_id`	| IoT Hub, DPS, and Central Tagging	| Job Id populated by an internal pipeline run. Can be manually set to customize the pipeline name tag. Please see "Test Resource Tagging" for more details.	|
 
-`azext_iot_teststoragecontainer` is the name of blob container belonging to the above mentioned storage account. Optional environment variable, defaults to 'devices' when not specified.
 
-##### IoT Hub
+##### IoT Central
 
-Execute the following command to run the IoT Hub integration tests:
+IoT Central integration tests can be run against a specific available api versions using command line argument "--api-version"
```
digimaun, line 221:
```suggestion
IoT Central integration tests can be run against a specific available api version using command line argument "--api-version"
```

digimaun, line 221:
```suggestion
IoT Central integration tests can be run against a specific available api version using command line argument "--api-version"
```

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -23,6 +23,7 @@
 logger = get_logger(__name__)
 
 
+@pytest.mark.skip(reason="Waiting on service team")
```
digimaun, line 26: Let's make sure we don't lose track of this

vilit1:
passing data history (with a new kusto cluster) https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5444&view=results

digimaun (approved):

PR #507, opened 2022-04-15
Certops rewrite to use Cryptography

vilit1:
Rewrite certops.py to use the cryptography library rather than the pyopenssl library. This solves the issue of having to uninstall and reinstall both libraries because they cannot be installed correctly the first time.

Before:
image

After:
image

Pipeline run
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=5449&view=results

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/common/certops.py
+++ b/azext_iot/common/certops.py
@@ -60,7 +60,7 @@ def create_self_signed_certificate(subject, valid_days, cert_output_dir, cert_on
     ).decode('utf-8')
     key_dump = cert.public_bytes(serialization.Encoding.PEM).decode('utf-8')
     thumbprint_bytes = cert.fingerprint(hashes.SHA1())
-    thumbprint = ':'.join(f'{b:02X}' for b in thumbprint_bytes)
+    thumbprint = ''.join(f'{b:02X}' for b in thumbprint_bytes)
 
     if cert_output_dir and exists(cert_output_dir):
         cert_file = subject + '-cert.pem'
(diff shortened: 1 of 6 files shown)


azext_iot/common/certops.py
```diff
@@ -76,11 +96,13 @@ def open_certificate(certificate_path):
         certificate (str): returns utf-8 encoded value from certificate file.
     """
     certificate = ""
-    if certificate_path.endswith('.pem') or certificate_path.endswith('.cer'):
+    if certificate_path.endswith(".pem") or certificate_path.endswith(".cer"):
```
c-ryan-k, line 99: Something we should probably patch up here while we're at it - if `certificate_path` is not a `.pem` or `.cer` file - we end up returning `""` for certificate here - should we instead throw an exception?

azext_iot/common/certops.py
```diff
@@ -8,12 +8,18 @@
 certops: Functions for working with certificates.
 """
 
+import datetime
 from os.path import exists, join
 import base64
-from OpenSSL import crypto
```
c-ryan-k, line 13: I believe this is our only import of `OpenSSL` - If so we can probably remove it from the `.pylintrc` ignored modules:
https://github.com/Azure/azure-iot-cli-extension/blob/58de288788d3d3bbe978f8311f47e9783aa0ca75/.pylintrc#L693

azext_iot/common/certops.py
digimaun: .cert and .crt (as well as .cer) are common synonyms for .pem

c-ryan-k (approved):

PR #509, opened 2022-04-20
Fix tests on IoT Edge devices for Central

lucadruda:
### General Guidelines

Intent for Production

Basic expectations

- Race condition when deleting child devices with ApiVersion 1.0 got removed.
INT test:
image
image

### DIFF
--- a/azext_iot/tests/central/test_iot_central_devices_int.py
+++ b/azext_iot/tests/central/test_iot_central_devices_int.py
@@ -297,7 +297,7 @@ class TestIotCentralDevices(CentralLiveScenarioTest):
         )
 
         # wait about a few seconds for simulator to kick in so that provisioning completes
-        time.sleep(30)
+        time.sleep(60)
 
         if api_version == ApiVersion.v1_1_preview.value:
             # check if device appears as edge
--- a/azext_iot/tests/central/test_iot_central_devices_int.py
+++ b/azext_iot/tests/central/test_iot_central_devices_int.py
@@ -324,7 +324,7 @@ class TestIotCentralDevices(CentralLiveScenarioTest):
         command = "iot central device edge children list --app-id {} -d {}".format(
             self.app_id, device_id
         )
-        
+
         children = self.cmd(
             command, api_version=ApiVersion.v1_1_preview.value
         ).get_output_in_json()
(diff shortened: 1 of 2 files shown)


azext_iot/tests/central/test_iot_central_devices_int.py
```diff
@@ -318,25 +318,29 @@ def test_central_edge_children_methods(self):
...
         self.cmd(command, api_version=ApiVersion.v1_1_preview.value)
+        # wait for relationship to be established
+        time.sleep(10)
 
         command = "iot central device edge children list --app-id {} -d {}".format(
             self.app_id, device_id
         )
         # Use api-version 1.0 to also test interoperability
         children = self.cmd(
-            command, api_version=ApiVersion.v1.value
```
digimaun, line 327: Will we have a gap in key test coverage by setting the API version here from v1 to v1.1 preview?

digimaun, line 327: Will we have a gap in key test coverage by setting the API version here from v1 to v1.1 preview?

digimaun (approved):

azext_iot/tests/central/test_iot_central_devices_int.py
vilit1: Especially concerning since the comment says use api-vresion 1.0

vilit1 (approved):

azext_iot/tests/central/test_iot_central_devices_int.py
```diff
@@ -297,7 +297,7 @@ def test_central_edge_children_methods(self):
         )
 
         # wait about a few seconds for simulator to kick in so that provisioning completes
-        time.sleep(60)
+        time.sleep(30)
```
vilit1, line 300: Other simulator sleeps are still 60

azext_iot/tests/central/test_iot_central_devices_int.py
lucadruda: right now v1.0 supports edge children by using device twin info. We have alerts on the command to push people using 1.1-preview. Since using device twins causes delays in updating configurations, tests may fail for race conditions as we've seen. So reducing the test to 1.1-preview which is the real supported one.

lucadruda: same as above

azext_iot/tests/central/test_iot_central_devices_int.py
lucadruda: good catch, leftover

PR #515, opened 2022-05-23
Dps device registration using device sdk

vilit1:
Support DPS device registration using the Device sdk (https://github.com/Azure/azure-iot-sdk-python)

Note that this will just register a device using the DPS and it should end up in IoT Hub. Since the HTTP protocol for device registration is not as "stable" as the mqtt pipeline in the device sdk, the device sdk is used.

Key usability improvements:
1. Can retrieve the id scope
2. Can retrieve the symmetric key or generate device key
3. If the id scope and symmetric key/device key is provided, can avoid any az login attempts.

Note that there are a lot of "fixes" and "add-ons" for this:
1. The return type from the device sdk registration is manually created. The device sdk registration state's string format does not adhere to our purposes.
2. Error handling involves error matching and changing the error message to be more understandable.

Some x509 limitations discovered:
- the registration id for individual enrollments must be the same as the subject: https://docs.microsoft.com/en-us/azure/iot-dps/concepts-service#individual-enrollment
- group id is not needed for group enrollments - the registration id is only used.

Help Page in question:
```
Command
    az iot device registration create : Register an IoT Device with IoT Device Provisioning Service.
        The following attestation mechanisms are supported:
        - Symmetric key
        - x509 certificate
        If using x509 authentication methods, the certificate and key files must be provided.
        Command group 'iot device registration' is in preview and under development. Reference
        and support levels: https://aka.ms/CLI_refstatus

Arguments
    --registration-id --rid     [Required] : Device registration ID or individual enrollment ID.
    --enrollment-group-id --gid --group-id : Enrollment group ID.
    --payload                              : Custom allocation payload as JSON. Specifically for use
                                             with custom allocation policies using Azure Functions.
    --resource-group -g                    : Name of resource group. You can configure the default
                                             group using `az configure --defaults group=<name>`.

Access Control Arguments
    --auth-type                            : Indicates whether the operation should auto-derive a
                                             policy key or use the current Azure AD session. You can
                                             configure the default using `az configure --defaults
                                             iothub-data-auth-type=<auth-type-value>`.  Allowed
                                             values: key, login.  Config: key.

Device Provisioning Service Identifier Arguments
    --dps-name -n                          : Name of the Azure IoT Hub Device Provisioning Service.
                                             Required if --login is not provided or authenticaton
                                             arguments and --id-scope are not provided.
    --id-scope --scope                     : Id Scope of the Azure IoT Hub Device Provisioning
                                             Service. If provided with authentication arguments,
                                             will avoid session login.
[...]

PR #525, opened 2022-06-09
Add Iot Central device group CRUD support

HangyiWang:
---

### General Guidelines

Intent for Production

Basic expectations

Screenshot 2022-06-09 010602
Screenshot 2022-06-09 010519

### DIFF
--- a/azext_iot/central/command_map.py
+++ b/azext_iot/central/command_map.py
@@ -183,6 +183,10 @@ def load_central_commands(self, _):
         is_preview=True,
     ) as cmd_group:
         cmd_group.command("list", "list_device_groups")
+        cmd_group.show_command("show", "get_device_group")
+        cmd_group.command("create", "create_device_group")
+        cmd_group.command("update", "update_device_group")
+        cmd_group.command("delete", "delete_device_group")
 
     with self.command_group(
         "iot central role", command_type=central_roles_ops, is_preview=True
--- a/azext_iot/central/commands_device_group.py
+++ b/azext_iot/central/commands_device_group.py
@@ -35,7 +35,7 @@ def get_device_group(
     token=None,
     central_dns_suffix=CENTRAL_ENDPOINT,
     api_version=ApiVersion.v1_1_preview.value,
-) -> List[Union[DeviceGroupPreview, DeviceGroupV1_1_preview]]:
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
     provider = CentralDeviceGroupProvider(
         cmd=cmd, app_id=app_id, token=token, api_version=api_version
     )
--- a/azext_iot/central/commands_device_group.py
+++ b/azext_iot/central/commands_device_group.py
@@ -6,6 +6,8 @@
 # Dev note - think of this as a controller
 
 from typing import List, Union
+from azure.cli.core.azclierror import InvalidArgumentValueError
+from azext_iot.common import utility
 from azext_iot.constants import CENTRAL_ENDPOINT
 from azext_iot.central.models.preview import DeviceGroupPreview
 from azext_iot.central.models.v1_1_preview import DeviceGroupV1_1_preview
@@ -24,3 +26,87 @@ def list_device_groups(
         cmd=cmd, app_id=app_id, token=token, api_version=api_version
     )
     return provider.list_device_groups(central_dns_suffix=central_dns_suffix)
+
+
+def get_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+) -> List[Union[DeviceGroupPreview, DeviceGroupV1_1_preview]]:
+    provider = CentralDeviceGroupProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+    return provider.get_device_group(device_group_id=device_group_id, central_dns_suffix=central_dns_suffix)
+
+
+def create_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    content: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+):
+    if not isinstance(content, str):
+        raise InvalidArgumentValueError("content must be a string: {}".format(content))
+
+    payload = utility.process_json_arg(content, argument_name="content")
+
+    provider = CentralDeviceGroupProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+
+    device_group = provider.create_device_group(
+        device_group_id=device_group_id,
+        payload=payload,
+        central_dns_suffix=central_dns_suffix,
+    )
+    return device_group
+
+
+def update_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    content: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+):
+    if not isinstance(content, str):
+        raise InvalidArgumentValueError("content must be a string: {}".format(content))
+
+    payload = utility.process_json_arg(content, argument_name="content")
+
+    provider = CentralDeviceGroupProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+
+    device_group = provider.update_device_group(
+        device_group_id=device_group_id,
+        payload=payload,
+        central_dns_suffix=central_dns_suffix,
+    )
+    return device_group
+
+
+def delete_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1.value,
+):
+    provider = CentralDeviceGroupProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+
+    return provider.delete_device_group(
+        device_group_id=device_group_id,
+        central_dns_suffix=central_dns_suffix,
+    )
--- a/azext_iot/central/providers/device_group_provider.py
+++ b/azext_iot/central/providers/device_group_provider.py
@@ -6,6 +6,10 @@
 
 
 from typing import List, Union
+from azure.cli.core.azclierror import (
+    RequiredArgumentMissingError,
+    ResourceNotFoundError
+)
 from knack.log import get_logger
 from azext_iot.constants import CENTRAL_ENDPOINT
 from azext_iot.central import services as central_services
@@ -52,3 +56,92 @@ class CentralDeviceGroupProvider:
         )
 
         return device_groups
+
+    def get_device_group(
+        self, device_group_id, central_dns_suffix=CENTRAL_ENDPOINT
+    ) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
+        # get or add to cache
+        device_group = self._device_groups.get(device_group_id)
+        if not device_group:
+            device_group = central_services.device_group.get_device_group(
+                cmd=self._cmd,
+                app_id=self._app_id,
+                device_group_id=device_group_id,
+                token=self._token,
+                central_dns_suffix=central_dns_suffix,
+                api_version=self._api_version,
+            )
+            # add to cache
+            self._device_groups[device_group_id] = device_group
+
+        if not device_group:
+            raise ResourceNotFoundError(
+                "No device group for device group with id: '{}'.".format(
+                    device_group_id
+                )
+            )
+
+        return device_group
+
+    def create_device_group(
+        self,
+        device_group_id,
+        payload: str,
+        central_dns_suffix=CENTRAL_ENDPOINT
+    ):
+        device_group = central_services.device_group.create_device_group(
+            cmd=self._cmd,
+            app_id=self._app_id,
+            device_group_id=device_group_id,
+            payload=payload,
+            token=self._token,
+            central_dns_suffix=central_dns_suffix,
+            api_version=self._api_version,
+        )
+
+        self._device_groups[device_group.id] = device_group
+
+        return device_group
+
+    def update_device_group(
+        self,
+        device_group_id,
+        payload: str,
+        central_dns_suffix=CENTRAL_ENDPOINT
+    ):
+        device_group = central_services.device_group.update_device_group(
+            cmd=self._cmd,
+            app_id=self._app_id,
+            device_group_id=device_group_id,
+            payload=payload,
+            token=self._token,
+            central_dns_suffix=central_dns_suffix,
+            api_version=self._api_version,
+        )
+
+        self._device_groups[device_group.id] = device_group
+
+        return device_group
+
+    def delete_device_group(
+        self,
+        device_group_id,
+        central_dns_suffix=CENTRAL_ENDPOINT,
+    ):
+        if not device_group_id:
+            raise RequiredArgumentMissingError("Device group id must be specified.")
+
+        result = central_services.device_group.delete_device_group(
+            cmd=self._cmd,
+            token=self._token,
+            app_id=self._app_id,
+            device_group_id=device_group_id,
+            central_dns_suffix=central_dns_suffix,
+            api_version=self._api_version,
+        )
+
+        # remove from cache
+        # pop "miss" raises a KeyError if None is not provided
+        self._device_groups.pop(device_group_id, None)
+
+        return result
--- a/azext_iot/central/services/device_group.py
+++ b/azext_iot/central/services/device_group.py
@@ -84,7 +84,7 @@ def get_device_group(
     token: str,
     api_version: str,
     central_dns_suffix=CENTRAL_ENDPOINT,
-) -> List[Union[DeviceGroupPreview, DeviceGroupV1_1_preview]]:
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
     """
     Get a specific device group from IoTC
 
--- a/azext_iot/central/services/device_group.py
+++ b/azext_iot/central/services/device_group.py
@@ -20,6 +20,7 @@ from azext_iot.central.models.enum import ApiVersion
 logger = get_logger(__name__)
 
 BASE_PATH = "api/deviceGroups"
+MODEL = "deviceGroup"
 
 
 def list_device_groups(
@@ -74,3 +75,161 @@ def list_device_groups(
         pages_processed = pages_processed + 1
 
     return device_groups
+
+
+def get_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token: str,
+    api_version: str,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+) -> List[Union[DeviceGroupPreview, DeviceGroupV1_1_preview]]:
+    """
+    Get a specific device group from IoTC
+
+    Args:
+        cmd: command passed into az
+        device_group_id: case sensitive device group id,
+        app_id: name of app (used for forming request URL)
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        device_group: dict
+    """
+
+    url = "https://{}.{}/{}/{}".format(
+        app_id, central_dns_suffix, BASE_PATH, device_group_id
+    )
+    headers = _utility.get_headers(token, cmd)
+
+    # Construct parameters
+    query_parameters = {}
+    query_parameters["api-version"] = api_version
+
+    response = requests.get(url, headers=headers, params=query_parameters)
+    result = _utility.try_extract_result(response)
+    return _utility.get_object(result, model=MODEL, api_version=api_version)
+
+
+def create_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    payload: dict,
+    token: str,
+    api_version: str,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
+    """
+    Create a device group in IoTC
+
+    Args:
+        cmd: command passed into az
+        app_id: name of app (used for forming request URL)
+        device_group_id: case sensitive device group id,
+        payload: see example payload available in
+            <repo-root>/azext_iot/tests/central/json/device_group_int_test.json
+            or check here for more information
+            https://docs.microsoft.com/en-us/rest/api/iotcentral/devicegroups
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        device_group: dict
+    """
+
+    url = "https://{}.{}/{}/{}".format(
+        app_id, central_dns_suffix, BASE_PATH, device_group_id
+    )
+    headers = _utility.get_headers(token, cmd, has_json_payload=True)
+
+    # Construct parameters
+    query_parameters = {}
+    query_parameters["api-version"] = api_version
+
+    response = requests.put(url, headers=headers, json=payload, params=query_parameters)
+    result = _utility.try_extract_result(response)
+    return _utility.get_object(result, model=MODEL, api_version=api_version)
+
+
+def update_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    payload: dict,
+    token: str,
+    api_version: str,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
+    """
+    Updates a device group in IoTC
+
+    Args:
+        cmd: command passed into az
+        app_id: name of app (used for forming request URL)
+        device_group_id: case sensitive device group id,
+        payload: see example payload available in
+            <repo-root>/azext_iot/tests/central/json/device_group_int_test.json
+            or check here for more information
+            https://docs.microsoft.com/en-us/rest/api/iotcentral/devicegroups
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        device_group: dict
+    """
+
+    url = "https://{}.{}/{}/{}".format(
+        app_id, central_dns_suffix, BASE_PATH, device_group_id
+    )
+    headers = _utility.get_headers(token, cmd, has_json_payload=True)
+
+    # Construct parameters
+    query_parameters = {}
+    query_parameters["api-version"] = api_version
+
+    response = requests.patch(
+        url, headers=headers, json=payload, params=query_parameters
+    )
+    result = _utility.try_extract_result(response)
+    return _utility.get_object(result, MODEL, api_version)
+
+
+def delete_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token: str,
+    api_version: str,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+) -> dict:
+    """
+    Delete a device group from IoTC
+
+    Args:
+        cmd: command passed into az
+        app_id: name of app (used for forming request URL)
+        device_group_id: case sensitive device group id,
+        token: (OPTIONAL) authorization token to fetch device details from IoTC.
+            MUST INCLUDE type (e.g. 'SharedAccessToken ...', 'Bearer ...')
+        central_dns_suffix: {centralDnsSuffixInPath} as found in docs
+
+    Returns:
+        device_group: dict
+    """
+    url = "https://{}.{}/{}/{}".format(
+        app_id, central_dns_suffix, BASE_PATH, device_group_id
+    )
+    headers = _utility.get_headers(token, cmd)
+
+    # Construct parameters
+    query_parameters = {}
+    query_parameters["api-version"] = api_version
+
+    response = requests.delete(url, headers=headers, params=query_parameters)
+    return _utility.try_extract_result(response)
--- a/azext_iot/central/services/device_template.py
+++ b/azext_iot/central/services/device_template.py
@@ -45,7 +45,7 @@ def get_device_template(
         central_dns_suffix: {centralDnsSuffixInPath} as found in docs
 
     Returns:
-        device: dict
+        device_template: dict
     """
     url = "https://{}.{}/{}/{}".format(
         app_id, central_dns_suffix, BASE_PATH, device_template_id
--- a/azext_iot/tests/central/test_iot_central_unit.py
+++ b/azext_iot/tests/central/test_iot_central_unit.py
@@ -365,6 +365,7 @@ class TestCentralDeviceGroupProvider:
         DeviceGroupV1_1_preview(group)
         for group in load_json(FileNames.central_device_group_file)
     ]
+    _device_group = load_json(FileNames.central_device_group_file)[0]
 
     @mock.patch("azext_iot.central.services.device_group")
     def test_should_return_device_groups(self, mock_device_group_svc):
@@ -397,6 +398,33 @@ class TestCentralDeviceGroupProvider:
         # call counts should be at most 1 since the provider has a cache
         assert mock_device_group_svc.list_device_groups.call_count == 1
         assert set(device_groups) == set(self._device_groups)
+
+    @mock.patch("azext_iot.central.services.device_group")
+    def test_should_update_device_group_name(self, mock_device_group_svc):
+        # setup
+        provider = CentralDeviceGroupProvider(
+            cmd=None, app_id=app_id, api_version=ApiVersion.v1_1_preview.value
+        )
+        existing = get_object(
+            self._device_group, "DeviceGroup", api_version=ApiVersion.v1_1_preview.value
+        )
+        display_name = "NewDeviceGroupName"
+        mock_device_group_svc.get_device_group.return_value = existing
+        updated_device_group_dict = deepcopy(self._device_group)
+        updated_device_group_dict["displayName"] = display_name
+        mock_device_group_svc.update_device_group.return_value = get_object(
+            updated_device_group_dict, "DeviceGroup", api_version=ApiVersion.v1_1_preview.value
+        )
+
+        # act
+        device_group = provider.update_device_group(
+            device_group_id=existing.id, payload=updated_device_group_dict
+        )
+
+        # verify
+        # call counts should be at most 1 since the provider has a cache
+        assert mock_device_group_svc.update_device_group.call_count == 1
+        assert device_group.display_name == display_name
 
 
 class TestCentralRoleProvider:


HangyiWang:
Not sure why some checks failed. I don't have access to these details. 😢

digimaun:
> Not sure why some checks failed. I don't have access to these details. 😢

You should have access to see the pipeline runs now.

HangyiWang:
> > Not sure why some checks failed. I don't have access to these details. 😢
> 
> You should have access to see the pipeline runs now.

There's an error for no defaults  parameter in update command. But I found some other iot central update commands with default parameters. 😕  

```
-  FAIL - HIGH severity: no_parameter_defaults_for_update_commands
    Parameter: iot central device-group update, `api_version` - Update commands should not have parameters with default values. 'api_version' in 'iot central device-group update' has a default value of '1.1-preview'
    Parameter: iot central device-group update, `central_dns_suffix` - Update commands should not have parameters with default values. 'central_dns_suffix' in 'iot central device-group update' has a default value of 'azureiotcentral.com'
```

HangyiWang:
And there're 2 other errors not from iot central, but digital twins. I didn't change those parts. 👌 

```
FAIL - MEDIUM severity: require_wait_command_if_no_wait
    Command-Group: `dt data-history connection create` - Group does not have a 'wait' command, yet 'dt data-history connection create adx' exposes '--no-wait'
    Command-Group: `dt endpoint create` - Group does not have a 'wait' command, yet 'dt endpoint create eventgrid' exposes '--no-wait'

-  FAIL - MEDIUM severity: parameter_should_not_end_in_resource_group
    Parameter: dt data-history connection create adx, `adx_resource_group` - A command should only have '--resource-group' as its resource group parameter. However options '--adx-resource-group' in command 'dt data-history connection create adx' end with 'resource-group' or similar.
    Parameter: dt data-history connection create adx, `eh_resource_group` - A command should only have '--resource-group' as its resource group parameter. However options '--eventhub-resource-group' in command 'dt data-history connection create adx' end with 'resource-group' or similar.
```

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +26,87 @@ def list_device_groups(
...
+    )
+    return device_group
+
+
+def update_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    content: str,
+    central_dns_suffix: str,
```
vilit1, line 76: why is there no default argument for this central_dns_suffix?

azext_iot/central/params.py
```diff
@@ -223,6 +223,50 @@ def load_central_arguments(self, _):
...
+            "content",
+            options_list=["--content", "-k"],
+            help="The device group definition. Provide path to JSON file or raw stringified JSON."
+            " [File Path Example: ./path/to/file.json] [Example of stringified JSON: {<Device Group JSON>}]."
+        )
+
+    with self.argument_context("iot central device-group update") as context:
+        context.argument(
+            "device_group_id",
+            options_list=["--device_group_id", "-d"],
```
vilit1, line 251: --device-group-id

Also no need for the regex pattern.

azext_iot/central/params.py
```diff
@@ -223,6 +223,50 @@ def load_central_arguments(self, _):
...
+            help="Unique ID of the device group."
+            "Regex pattern: ^[a-zA-Z0-9_-]*$",
+        )
+        context.argument(
+            "content",
+            options_list=["--content", "-k"],
+            help="The device group definition. Provide path to JSON file or raw stringified JSON."
```
vilit1, line 258: We have other commands that have similar inputs, like `az dt model create` specifically --models.

Try to follow those patterns to make the cli more unified. Also no need for the file path example and example of string json

azext_iot/central/params.py
vilit1: please remove the regex pattern for the other params too

azext_iot/central/commands_device_group.py
HangyiWang: Actually the azure pipeline check failed this part. Where should I remove the default parameter instead here? 😲 
```
-  FAIL - HIGH severity: no_parameter_defaults_for_update_commands
    Parameter: iot central device-group update, `api_version` - Update commands should not have parameters with default values. 'api_version' in 'iot central device-group update' has a default value of '1.1-preview'
    Parameter: iot central device-group update, `central_dns_suffix` - Update commands should not have parameters with default values. 'central_dns_suffix' in 'iot central device-group update' has a default value of 'azureiotcentral.com'
```

azext_iot/central/params.py
HangyiWang: Yes. Will change this.

azext_iot/central/params.py
HangyiWang: Sure.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +26,87 @@ def list_device_groups(
...
+def create_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    content: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+):
+    if not isinstance(content, str):
```
vilit1, line 54: the azure cli should technically make this argument into string, especially if you identify this as str in this file (command_device_group)

So there isnt a need for this check.

azext_iot/central/providers/device_group_provider.py
```diff
@@ -52,3 +56,92 @@ def list_device_groups(
...
 
         return device_groups
+
+    def get_device_group(
+        self, device_group_id, central_dns_suffix=CENTRAL_ENDPOINT
+    ) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
+        # get or add to cache
+        device_group = self._device_groups.get(device_group_id)
+        if not device_group:
+            device_group = central_services.device_group.get_device_group(
```
vilit1, line 66: will this really not just throw an error if you provide a bad device group (non existant)? what if you provide the wrong app id?

azext_iot/central/commands_device_group.py
HangyiWang: I see. Thanks!

vilit1:
why not make some int tests too?

azext_iot/central/providers/device_group_provider.py
HangyiWang: I followed the pattern in `get_device_template` and I think it will raise `AzureResponseError` error in service utility in `try_extract_result` function if the response is 404.

azext_iot/central/commands_device_group.py
vilit1: if you look at other commands, you will see that you want a default central_dns_suffix. To make the linter not angry like here, you will need to edit linter_exclusions.yml which is in the root directory of the this project

HangyiWang: Gotcha. Thanks! That's what I was trying to figure out.

azext_iot/central/params.py
HangyiWang: I followed the some patterns in iot central create commands here.

azext_iot/central/providers/device_group_provider.py
HangyiWang: Yes, I saw those services using try-except to handle exceptions in a generic handling function. But I think my way here should work as well.

lucadruda: Services throw out error messages as they come out from API call. Almost all central commands are wrappers around the public APIs so I would keep it as is and let throw at service level.

azext_iot/central/services/device_group.py
```diff
@@ -74,3 +75,175 @@ def list_device_groups(
...
+def get_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token: str,
+    api_version: str,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
+    """
+    Get a specific device group from IoTC
```
lucadruda, line 89: nit: "IoT Central" or remove it.

lucadruda (changes requested):
Create a v1.2-preview models folder

azext_iot/central/command_map.py
```diff
@@ -183,6 +183,10 @@ def load_central_commands(self, _):
         is_preview=True,
     ) as cmd_group:
         cmd_group.command("list", "list_device_groups")
+        cmd_group.show_command("show", "get_device_group")
```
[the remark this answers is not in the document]
lucadruda: this is one of the causes for checks failures

azext_iot/central/services/device_template.py
```diff
@@ -45,7 +45,7 @@ def get_device_template(
         central_dns_suffix: {centralDnsSuffixInPath} as found in docs
 
     Returns:
-        device: dict
+        device_template: dict
```
yang-hai-feng, line 48: nice catch!

azext_iot/central/params.py
```diff
@@ -223,6 +223,74 @@ def load_central_arguments(self, _):
             help="The API version for the requested operation.",
         )
 
+    with self.argument_context("iot central device-group show") as context:
```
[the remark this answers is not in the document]
HangyiWang: Sounds good!

azext_iot/central/command_map.py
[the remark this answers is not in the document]
HangyiWang: Gotcha. Thanks! Actually I've been looking for the root cause for this error.

azext_iot/central/services/device_group.py
HangyiWang: Okay

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
     return provider.list_device_groups(central_dns_suffix=central_dns_suffix)
+
+
+def get_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
```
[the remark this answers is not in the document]
HangyiWang: Yes, but this would be changed after I add the 2022-05-31 ga version in another PR.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
+) -> Union[DeviceGroupPreview, DeviceGroupV1_1_preview]:
```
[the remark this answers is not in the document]
HangyiWang: Yep. I will add another PR for new GA version.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
+    app_id: str,
+    device_group_id: str,
+    display_name: str,
+    filter: str,
+    description: str = None,
+    organizations: List[str] = None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
+):
```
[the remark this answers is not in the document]
HangyiWang: Sure. Will add it.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    display_name: str,
+    filter: str,
+    description: str = None,
+    organizations: List[str] = None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
```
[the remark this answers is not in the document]
HangyiWang: Okay. I didn't realized it.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    display_name: str = None,
+    filter: str = None,
+    description: str = None,
+    organizations: List[str] = None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.v1_1_preview.value,
```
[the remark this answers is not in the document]
HangyiWang: Sure

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
HangyiWang: 😢 To fix this, either I need to remove this function or I add support for new GA version. I think I will remove this first then add it later in another GA PR.

vilit1:
regarding https://github.com/Azure/azure-iot-cli-extension/pull/525#issuecomment-1151685055

I am going to update the linter exclusions for dt in my pr https://github.com/Azure/azure-iot-cli-extension/pull/515

HangyiWang:
Adding 2022-05-31 version support for Iot central. UT and IT will be in next commit.

azext_iot/central/commands_device_group.py
```diff
@@ -24,3 +24,89 @@ def list_device_groups(
...
+
+
+def create_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    display_name: str,
+    filter: str,
+    description: str = None,
+    organizations: List[str] = None,
```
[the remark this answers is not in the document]
HangyiWang: Oh I directly followed the definition in rest API official doc. It did say we only support one right now but should we use str but not List[str] here?

azext_iot/central/providers/device_group_provider.py
HangyiWang: @vilit1 Remove old preview versions shouldn't have big user impact and it'll make codebase cleaner.

azext_iot/central/models/v1_1_preview/device_group.py
```diff
@@ -10,3 +10,5 @@ def __init__(self, group: dict):
         self.display_name = group.get("displayName")
         self.id = group.get("id")
         self.organizations = group.get("organizations")
+        self.filter = group.get("filter")
```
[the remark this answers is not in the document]
HangyiWang: @lucadruda  I created ga_2022_05_31 folder for this model.

HangyiWang:
> regarding [#525 (comment)](https://github.com/Azure/azure-iot-cli-extension/pull/525#issuecomment-1151685055)
> 
> I am going to update the linter exclusions for dt in my pr #515

Thanks for updating. Somehow it didn't work in my PR so I updated the lint rules a little.

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
HangyiWang: Updated version.

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
HangyiWang: Added.

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
HangyiWang: Updated.

azext_iot/central/services/device_group.py
HangyiWang: Removed.

azext_iot/central/command_map.py
```diff
@@ -143,6 +143,19 @@ def load_central_commands(self, _):
         cmd_group.command("compute-device-key", "compute_device_key")
         cmd_group.command("manual-failover", "run_manual_failover")
         cmd_group.command("manual-failback", "run_manual_failback")
+        cmd_group.command("show-attestation", "get_attestation")
```
vilit1, line 146: You should put some of these commands into sub groups - like iot central device attestation. Helps with making the iot central device not as cluttered

azext_iot/central/command_map.py
```diff
@@ -235,6 +252,24 @@ def load_central_commands(self, _):
         cmd_group.command("list", "list_device_modules")
         cmd_group.show_command("show", "get_device_module")
         cmd_group.command("restart", "restart_device_module")
+        cmd_group.command("list-components", "list_module_components")
```
vilit1, line 255: please make sub groups

azext_iot/central/params.py
```diff
@@ -214,13 +224,29 @@ def load_central_arguments(self, _):
...
+        )
+        context.argument(
+            "description",
+            options_list=["--description"],
+            help="Short summary of device group."
+        )
+        context.argument(
+            "organizations",
+            options_list=["--organizations"],
+            help="List of organization IDs of the device group."
```
vilit1, line 249: use nargs for lists. By default everything is a string

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
vilit1: if it is going to be multiple organizations at some point, it is better to have the list. Just make sure that the command works properly

azext_iot/central/_help.py
```diff
@@ -850,6 +1051,71 @@ def _load_central_device_groups_help():
...
+            --description {description}
+            --organizations {organizations}
+    """
+
+    helps[
+        "iot central device-group update"
+    ] = """
+        type: command
+        short-summary: Update a device group via patch.
+        long-summary: Update an existing device group by ID.
```
vilit1, line 1091: for most of your one sentence long summaries, you can just replace the short summaries with the long ones. The long summaries are for multi line/specific instructions.

azext_iot/central/params.py
HangyiWang: I see.

azext_iot/central/command_map.py
HangyiWang: Okay.

azext_iot/central/command_map.py
HangyiWang: Sure. I'll create some new command groups.

HISTORY.rst
```diff
@@ -25,6 +25,13 @@ unreleased
   are supported. Once registered, the device will show up in the linked IoT Hub and can be interacted with or
   simulated using other `az iot device` commands.
 
+**IoT Central updates**
+
+* Add support for device groups CRUD.
```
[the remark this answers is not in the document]
HangyiWang: Okay. Will add more details.

azext_iot/central/commands_device.py
```diff
@@ -303,6 +355,241 @@ def get_command_history(
...
+    api_version=ApiVersion.ga_2022_05_31.value,
+) -> dict:
+    provider = CentralDeviceProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+
+    return provider.get_module_command_history(
+        device_id=device_id,
+        module_name=module_name,
+        component_name=None,
```
vilit1, line 375: how come component name is none? is this not supported yet?

azext_iot/central/commands_device.py
```diff
@@ -303,6 +355,241 @@ def get_command_history(
...
+    )
+
+    return provider.get_module_properties(
+        device_id=device_id,
+        module_name=module_name,
+        central_dns_suffix=central_dns_suffix,
+    )
+
+
+def replace_module_properties(
```
vilit1, line 425: update?

azext_iot/central/commands_device_group.py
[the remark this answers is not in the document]
HangyiWang: Okay, we'll going to support multiple orgs in the future. I'll test it.

azext_iot/central/commands_device.py
HangyiWang: This is calling `provider.replace_module_properties` for completely replacing module properties. while `update_module_properties` is only for partially patch.

azext_iot/central/providers/device_group_provider.py
HangyiWang: Actually we support all versions in this cli extension right now: preview, 1.1-preview and v1(GA version). One more preview version: 1.2-preview doesn't exist here. 😕 
I'm going to add 2022-05-31(GA) here. And we also have one new preview version and one new GA version incoming.
Since we're going to retire preview and 1.1-preview, we want to only keep one latest preview version in cli extension repo. Users still can utilize previous CLI release version if they want to use old API.
@yang-hai-feng

HISTORY.rst
```diff
@@ -25,6 +25,55 @@ unreleased
...
+
+* Add support for device/module properties CRUD.
+  - az iot central device list-components
+  - az iot central device show-telemetry-value
+  - az iot central device show-component-telemetry-value
+  - az iot central device properties
+    - az iot central device properties show
+    - az iot central device properties replace
+    - az iot central device properties update
+    - az iot central device properties show-component-properties
```
vilit1, line 53: do a separate group for component-properties:
az iot central device properties
az iot central device component-properties
the commands will feel less awk then

CONTRIBUTING.md
```diff
@@ -226,6 +226,28 @@ e.g. run tests against v 1.0
...
+    azext_iot_teststoragecontainer=
+    azext_iot_central_app_id=
+    azext_iot_central_scope_id=
+    azext_iot_central_token=
+    azext_iot_central_primarykey=
+    azext_iot_central_storage_cstring=
+    azext_iot_central_storage_container=
+```
+
+Not all of them are required. `azext_iot_teststoragecontainer` is the same as `azext_iot_central_storage_container`.
```
vilit1, line 249: why not incorporate this into the table above? We are also trying to reduce the number of pytest variables (unless you have a reason to have the storage account for central tests be different from the storage account for other iot services

azext_iot/central/_help.py
```diff
@@ -1375,6 +1650,232 @@ def _load_central_edge_help():
...
+            --device-id {deviceId}
+            --module-name {moduleName}
+            --content {json}
+    """
+
+    helps[
+        "iot central device edge module properties replace"
+    ] = """
+        type: command
+        short-summary: Replace all property values of a module.
```
vilit1, line 1772: will replace overwrite while update merges the changes with the original properties?

azext_iot/central/_help.py
```diff
@@ -1375,6 +1650,232 @@ def _load_central_edge_help():
...
+    helps[
+        "iot central device edge module command run-component-command"
+    ] = """
+            type: command
+            short-summary: Run a module component command.
+            long-summary: |
+                Note: payload should be nested under "request".
+                i.e. if your device expects the payload in a shape {"key": "value"}
+                payload should be {"request": {"key": "value"}}.
+                --content can also be pointed at a filepath like this (.../path/to/payload.json)
```
vilit1, line 1835: content note should be part of the param (inline content or path to filepath with content) and a separate example(s). Add one for file path and one for the payload example. 

will content ever have anything aside from the payload?

azext_iot/central/commands_device_group.py
```diff
@@ -18,9 +19,95 @@ def list_device_groups(
...
+        central_dns_suffix=central_dns_suffix,
+    )
+    return device_group
+
+
+def update_device_group(
+    cmd,
+    app_id: str,
+    device_group_id: str,
+    display_name: str = None,
```
vilit1, line 75: use optional[str] so linters correctly show str | None

azext_iot/central/_help.py
HangyiWang: We'll overwrite all writable properties through replace API.

CONTRIBUTING.md
HangyiWang: This is just an explanation for later contributors. That table above is for all integration tests here, not only for iot central. I didn't add anything new. I found that `azext_iot_teststoragecontainer` and `azext_iot_central_storage_container` are both working for iot central integration test so let others know.

HISTORY.rst
HangyiWang: Sure

azext_iot/central/commands_device_group.py
HangyiWang: Sure

azext_iot/central/models/ga_2022_05_31/template.py
```diff
@@ -0,0 +1,78 @@
...
+            self.id = template.get("@id")
+            self.interfaces = self._extract_interfaces(template)
+            self.schema_names = self._extract_schema_names(self.interfaces)
+            self.components = self._extract_components(template)
+            if self.components:
+                self.component_schema_names = self._extract_schema_names(
+                    self.components
+                )
+
+        except Exception:
```
vilit1, line 26: is there a more specific exception?

azext_iot/central/models/ga_2022_05_31/template.py
```diff
@@ -0,0 +1,78 @@
...
+                interfaces.append(self._extract_root_interface_contents(dcm))
+
+            if dcm.get("extends"):
+                interfaces.extend(dcm.get("extends"))
+
+            return {
+                interface["@id"]: self._extract_schemas(interface)
+                for interface in interfaces
+            }
+        except Exception:
```
vilit1, line 68: is there a more specific exception?

HISTORY.rst
```diff
@@ -25,6 +25,55 @@ unreleased
...
+  - az iot central device edge module show-component-telemetry-value
+  - az iot central device edge module properties
+    - az iot central device edge module properties show
+    - az iot central device edge module properties replace
+    - az iot central device edge module properties update
+    - az iot central device edge module properties show-component-properties
+    - az iot central device edge module properties replace-component-properties
+    - az iot central device edge module properties update-component-properties
+
+* Add support for 2022-05-31 GA version.
```
vilit1, line 67: is there any difference btw v1 and 2022-05-31 GA from the user's view? (like how devices are formatted, etc)

azext_iot/central/providers/device_group_provider.py
vilit1: make sure it is clearly documented which extension versions support which preview versions. Would be good to have this in a read me somewhere (as a table format of when we started to support the version to when we stopped)

azext_iot/central/models/v1_1_preview/device_group.py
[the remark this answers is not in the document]
lucadruda: can etag be provided by the user when run update commands? if yes we can add it to the model, otherwise this will simply be ignored in the output

azext_iot/central/providers/device_group_provider.py
HangyiWang: Sounds good. We'll do a  refactor in iot central cli extension to use SDK but not directly api call after this PR and we'll remove the old version in that one. We'll keep only latest preview and GA version in the future.

azext_iot/central/models/v1_1_preview/device_group.py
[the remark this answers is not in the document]
HangyiWang: No, it can only be specified in create-device-group command. We don't need to add this into model but we can accept etag when user create device group.

azext_iot/central/models/ga_2022_05_31/template.py
HangyiWang: No I used the same template model as old versions. They directly return this exception to users but not expose any other error messages like: schema parsing error, interface parsing error and etc. We can refactor this if more detailed message is necessary. But I won't do this in this PR only for this version.

azext_iot/central/models/ga_2022_05_31/template.py
HangyiWang: No I used the same template model as old versions. They directly return this exception to users but not expose any other error messages like: schema parsing error, interface parsing error and etc. We can refactor this if more detailed message is necessary later.

HISTORY.rst
HangyiWang: No. Azure API review board requires fully backward compatibility so every format, payload is the same from users' view. We support some new optional parameters and other new APIs in new GA version though.

azext_iot/central/params.py
```diff
@@ -613,6 +679,74 @@ def load_central_arguments(self, _):
...
+        )
+        context.argument(
+            "telemetry_name",
+            options_list=["--telemetry-name"],
+            help="The name of the device telemetry.",
+        )
+
+    with self.argument_context("iot central device edge module properties") as context:
+        context.argument(
+            "device_id",
```
lucadruda, line 700: same as above

yang-hai-feng:
> Do we want to keep deprecated versions?

@lucadruda, as API GA must ensure full backward compatibility

yang-hai-feng (approved):

azext_iot/central/params.py
```diff
@@ -212,15 +222,71 @@ def load_central_arguments(self, _):
             "by visiting https://github.com/iot-for-all/iot-central-high-availability-clients#readme",
         )
 
+    with self.argument_context("iot central device attestation") as context:
```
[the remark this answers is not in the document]
HangyiWang: I see.

HISTORY.rst
```diff
@@ -25,6 +25,55 @@ unreleased
...
+  - az iot central device show-component-telemetry-value
+  - az iot central device properties
+    - az iot central device properties show
+    - az iot central device properties replace
+    - az iot central device properties update
+    - az iot central device properties show-component-properties
+    - az iot central device properties replace-component-properties
+    - az iot central device properties update-component-properties
+  - az iot central device edge module list-components
+  - az iot central device edge module show-telemetry-value
```
vilit1, line 57: az iot central device edge module telemetry-value show

HISTORY.rst
vilit1: az iot central device properties component-properties show

HISTORY.rst
```diff
@@ -25,6 +25,55 @@ unreleased
...
+  - az iot central device edge module show-component-telemetry-value
+  - az iot central device edge module properties
+    - az iot central device edge module properties show
+    - az iot central device edge module properties replace
+    - az iot central device edge module properties update
+    - az iot central device edge module properties show-component-properties
```
vilit1, line 63: can you fix this to be less awk too?

CONTRIBUTING.md
vilit1: you can set them to the same value but by default they are different

CONTRIBUTING.md
```diff
@@ -226,6 +226,36 @@ e.g. run tests against v 1.0
...
+
+If you have the error message when you start running integration tests:
+
+```
+ERROR    cli.azure.cli.core.azclierror:azlogging.py:212 (ResourceGroupNotFound) Resource group 'xxxxx' could not be found.
+```
+
+You need to set up your azure CLI subscription using `az account set --subscription xxxxxx` so CLI could find the resource group successfully.
```
vilit1, line 257: this is more of common knowledge - as in the tests will run with your current credentials (and account - including sub). You do not need to have this in the contributing since local hand testing (as in writing out the commands) wont work till you are in the right sub.

HISTORY.rst
vilit1: when you fix this and others, make sure you fix it everywhere: the contributing, help, params, tests, commands, etc

vilit1:
can you rerun the int tests after your changes? Try to run int tests whenever you change something

HISTORY.rst
HangyiWang: Right, I'll fix this.

CONTRIBUTING.md
HangyiWang: Okay. I'll remove this.

HISTORY.rst
```diff
@@ -25,6 +25,55 @@ unreleased
   are supported. Once registered, the device will show up in the linked IoT Hub and can be interacted with or
   simulated using other `az iot device` commands.
 
+**IoT Central updates**
```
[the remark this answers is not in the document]
HangyiWang: No deprecated version in this PR. Will remove them in next PR.

azext_iot/central/command_map.py
```diff
@@ -235,6 +267,34 @@ def load_central_commands(self, _):
...
+        cmd_group.show_command("show", "get_module_component_properties")
+        cmd_group.command("replace", "replace_module_component_properties")
+        cmd_group.command("update", "update_module_component_properties")
+
+    with self.command_group(
+        "iot central device edge module command",
+        command_type=central_device_ops,
+    ) as cmd_group:
+        cmd_group.command("run", "run_module_command")
+        cmd_group.command("run-component-command", "run_module_component_command")
```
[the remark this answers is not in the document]
HangyiWang: Okay.

azext_iot/central/services/device_group.py
```diff
@@ -65,7 +67,11 @@ def list_device_groups(
             [
                 DeviceGroupPreview(device_group)
                 if api_version == ApiVersion.preview.value
-                else DeviceGroupV1_1_preview(device_group)
+                else (
+                    DeviceGroupV1_1_preview(device_group)
+                    if api_version == ApiVersion.v1_1_preview.value
+                    else DeviceGroupGa20220531(device_group)
```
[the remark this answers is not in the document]
HangyiWang: Sure. I'll make it more readable.

azext_iot/tests/central/__init__.py
```diff
@@ -31,6 +31,7 @@
     "azext_iot_central_dns_suffix",
     "azext_iot_teststorageaccount",
     "azext_iot_central_storage_container",
+    "azext_iot_teststoragecontainer",
```
vilit1, line 34: I see that you added this test variable but you do not use it anywhere aside from the delete. (like you do not check if it has a value and use that value, etc). I strongly recommend not pulling in this variable/ having delete depend on this variable so that we can keep the iot hub and central containers separate (and if the user wants to have the same container they would have to set both variables to point to the same container).

HangyiWang: Sure. Let me remove this. Just found that this variable also work for iot central int tests so I add it. It should be my misunderstanding.

HangyiWang:
> can you rerun the int tests after your changes? Try to run int tests whenever you change something

I'll attach the screenshot very soon.

CONTRIBUTING.md
HangyiWang: Thanks for the explanations. I removed all useless updates in this file.

HISTORY.rst
```diff
@@ -25,6 +25,60 @@ unreleased
...
+    - az iot central device attestation show
+    - az iot central device attestation create
+    - az iot central device attestation delete
+    - az iot central device attestation update
+
+* Add support for device/module properties/telemetry/command.
+  - az iot central device list-components
+  - az iot central device telemetry
+    - az iot central device telemetry show
+  - az iot central device component-telemetry
```
yang-hai-feng, line 49: we shouldn't create separated command for device telemetry, use ```az iot central device telemetry show --module-name xxx --component-name yyy --name temperature --device-id --app-id```

HISTORY.rst
```diff
@@ -25,6 +25,60 @@ unreleased
...
+    - az iot central device component-telemetry show
+  - az iot central device properties
```
yang-hai-feng, line 51: As we discussed, we shouldn't expose different `properties` concept from central, let's algin with `twin`, using following command to address get `twin` from device, device module, device component, device module component: ``` az iot central device twin show --app-id [appid] --device-id [deviceid] --module-name [modulename] --component-name [componentname]```

HISTORY.rst
```diff
@@ -25,6 +25,60 @@ unreleased
...
+    - az iot central device properties show
+    - az iot central device properties replace
+    - az iot central device properties update
+    - az iot central device component-properties show
+    - az iot central device component-properties replace
+    - az iot central device component-properties update
+  - az iot central device edge module list-components
```
yang-hai-feng, line 58: Shouldn't change any thing from ```az iot central device edge```, the concept `module` here is more on the agent definition within deployment manifest, we will have manifest associated, update operation for this command group soon.
all twin, command, telemetry from edge device should follow the same group as ``` az iot central device command|twin|telemetry```

yang-hai-feng (changes requested):

HISTORY.rst
yang-hai-feng: And please also include a new commands ```az iot central device list-modules``` to show module name list.

HISTORY.rst
HangyiWang: Right, working on this.

HISTORY.rst
HangyiWang: Yes. I'm updating these commands. Thanks for explanations.

HISTORY.rst
HangyiWang: Sure. Will add it.

CONTRIBUTING.md
```diff
@@ -226,6 +226,10 @@ e.g. run tests against v 1.0
 
 If the "--api-version" argument is not specified, all runs act against default api version for each tested command.
 
+To run specific test:
+
+`pytest azext_iot/tests/central/test_iot_central_int.py::TestIotCentral::test_central_query_methods_run`
```
vilit1, line 231: this applies to all tests - move this to the general Integration Tests section

azext_iot/central/_help.py
```diff
@@ -528,10 +686,10 @@ def _load_central_command_help():
...
             short-summary: Run a command on a device and view associated response. Does NOT monitor property updates that the command may perform.
             long-summary: |
-                Note: payload should be nested under "request".
-                i.e. if your device expects the payload in a shape {"key": "value"}
-                payload should be {"request": {"key": "value"}}.
-                --content can also be pointed at a filepath like this (.../path/to/payload.json)
+              Note: payload should be nested under "request".
+              i.e. if your device expects the payload in a shape {"key": "value"}
+              payload should be {"request": {"key": "value"}}.
+              --content can also be pointed at a filepath like this (.../path/to/payload.json)
```
vilit1, line 692: This help should be with the params or the help examples, not the long summary

vilit1: move the payload to the help example, content to param help (look at how we do help for `az dt model create`

azext_iot/central/_help.py
```diff
@@ -550,6 +708,25 @@ def _load_central_command_help():
                 -i {interfaceid}
                 --cn {commandname}
                 -k {payload}
+
+            - name: Run component command response
```
vilit1, line 712: what does run command response mean?

azext_iot/central/_help.py
```diff
@@ -1301,7 +1542,45 @@ def _load_central_monitors_help():
...
+            az iot central device twin update
+            --app-id {appid}
+            --device-id {deviceid}
+            -k {content}
+        - name: Replace device properties
+          text: >
+            az iot central device twin update
+            --app-id {appid}
+            --device-id {deviceid}
+            -r {replace}
```
vilit1, line 1565: can have the replace be a flag (as in just have -r without the {replace}

azext_iot/central/commands_device.py
```diff
@@ -679,3 +684,191 @@ def get_downstream_rel_name(
                 rel_name = content.get("name")
 
     return rel_name
+
+
+def get_attestation(
+    cmd,
+    app_id: str,
+    device_id,
```
vilit1, line 692: typing?

azext_iot/central/params.py
```diff
@@ -212,15 +227,40 @@ def load_central_arguments(self, _):
             "by visiting https://github.com/iot-for-all/iot-central-high-availability-clients#readme",
         )
 
+    with self.argument_context("iot central device twin") as context:
+        context.argument(
+            "replace",
+            options_list=["--replace", "-r"],
+            arg_type=get_three_state_flag(),
+            help="The boolean status for replacing device twin, True or False.",
```
vilit1, line 235: look at `az iot hub device-identity create` specifically the edge-enabled to see how we format flags.

azext_iot/central/_help.py
HangyiWang: Okay.

azext_iot/central/_help.py
HangyiWang: I used the previous command name. I'll change them since they're not very readable.

azext_iot/central/params.py
```diff
@@ -162,6 +162,21 @@ def load_central_arguments(self, _):
             help="The command name as specified in the device template. Command name could be different from the Display"
             " Name of the command.",
         )
+        context.argument(
+            "component_name",
+            options_list=["--component-name", "--co"],
```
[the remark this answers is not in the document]
HangyiWang: cn is for command_name in some commands.

azext_iot/central/params.py
HangyiWang: Okay.

azext_iot/central/_help.py
HangyiWang: Let me try

azext_iot/central/params.py
HangyiWang: Right. I was looking at the enabled flag in `az iot central export create`. That help message is not correct for the flag.

azext_iot/central/_help.py
```diff
@@ -478,7 +478,148 @@ def _load_central_devices_help():
...
+            az iot central device list-modules
+            --app-id {appid}
+            --device-id {deviceid}
+    """
+
+    helps[
+        "iot central device telemetry"
+    ] = """
+        type: group
+        short-summary: Query IoT Central devices telemetry value
```
yang-hai-feng, line 590: Nit: "Query IoT Central device latest telemetry value"

azext_iot/central/_help.py
```diff
@@ -520,6 +661,23 @@ def _load_central_command_help():
                 --device-id {deviceid}
                 --interface-id {interfaceid}
```
yang-hai-feng, line 662: Double check on --interface-id, is this a legacy meaning of combined command path? be sure it does not have conflict with what we provided from --module-name and --component-name.

azext_iot/central/commands_device.py
```diff
@@ -679,3 +684,191 @@ def get_downstream_rel_name(
...
+    module_name=None,
+    token=None,
+    central_dns_suffix=CENTRAL_ENDPOINT,
+    api_version=ApiVersion.ga_2022_05_31.value,
+) -> dict:
+    provider = CentralDeviceProvider(
+        cmd=cmd, app_id=app_id, token=token, api_version=api_version
+    )
+
+    return provider.list_device_module_components(
```
yang-hai-feng, line 788: this provide method is confuse here with list_device_module_components, I suppose it will be list_device_components, if module is not null then return from module

azext_iot/central/_help.py
```diff
@@ -520,35 +661,67 @@ def _load_central_command_help():
...
 
-            - name: Short Run command response
+            - name: Short Run command with json payload path.
               text: >
                 az iot central device command run
                 -n {appid}
                 -d {deviceid}
                 -i {interfaceid}
                 --cn {commandname}
+                -k {.../path/to/payload.json}
```
vilit1, line 706: you can just have -k {payload_path}

does the file also need to have the "request" in it?

azext_iot/central/_help.py
HangyiWang: No conflicts now, if interface is is component name, we'll treat it as component name. Otherwise we ignore it.

azext_iot/central/commands_device.py
HangyiWang: Sure. I'll update it.

azext_iot/central/_help.py
HangyiWang: Okay. payload is either under 'request' in inline json or in json file.

azext_iot/central/_help.py
HangyiWang: I need to check the interface in device module.

azext_iot/central/params.py
```diff
@@ -212,15 +227,40 @@ def load_central_arguments(self, _):
...
         context.argument(
-            "api_version",
-            options_list=["--api-version", "--av"],
-            choices=CaseInsensitiveList(
-                [ApiVersion.preview.value, ApiVersion.v1_1_preview.value]
-            ),
-            default=ApiVersion.v1_1_preview.value,
-            help="The API version for the requested operation.",
+            "device_group_id",
+            options_list=["--device-group-id", "-d"],
```
digimaun, line 241: We should be careful redefining short alias args (-d here) that are ubiquitous in other IoT CLI areas or even in central ....`-d` commonly refers to `--device-id`. On the flip side it is good to align arg definitions as much as possible because the devEx is more predictable for users.

azext_iot/central/_help.py
```diff
@@ -1301,7 +1538,45 @@ def _load_central_monitors_help():
...
+            -r
+        - name: Update device component properties
+          text: >
+            az iot central device twin update
+            --app-id {appid}
+            --device-id {deviceid}
+            --co {componentname}
+            -k {content}
+            -r
+        - name: Update device module component properties
```
digimaun, line 1571: When --replace / -r is used, the action is replace - we should note the difference in the description. Replace vs update may also use different HTTP verbs like put vs patch.

azext_iot/central/_help.py
HangyiWang: for clarification: don't have 'request' in file

azext_iot/central/params.py
HangyiWang: Correct. -d is not appropriate here since it's used by device-id. Fixed.

azext_iot/central/_help.py
HangyiWang: Yes, I was using another command: `iot central device twin replace` here for replacing action. Do we want to change it back or only add more explanations in this description?

azext_iot/central/_help.py
HangyiWang: Add one more check for interface in device module, if interface is component in device module, it should be treated as component name into api call as well.

azext_iot/central/params.py
HangyiWang: If we want to use `device twin replace` again, we might need to remove this flag

azext_iot/central/_help.py
digimaun: My comment was focused on the help docs / description to clearly illustrate the difference between replace and update operations.

Between `az iot central device twin [replace || update]` you, @yang-hai-feng and @lucadruda can make the call on it. To show past precedence for the iot hub CLI, we consider replace & update of device twin separate commands.

azext_iot/central/_help.py
```diff
@@ -1301,7 +1538,45 @@ def _load_central_monitors_help():
         "iot central device twin show"
     ] = """
         type: command
-        short-summary: Get the device twin from IoT Hub.
+        short-summary: Get the device properties from IoT Central.
```
digimaun, line 1541: Adding more content here?

HangyiWang:
Latest integration test result:

azext_iot/central/_help.py
HangyiWang: Sure.

azext_iot/central/_help.py
yang-hai-feng: Thanks @digimaun, I preferer to use separated commands for `replace` and `update`, @lucadruda, @HangyiWang  any thoughts?

azext_iot/central/models/ga_2022_05_31/organization.py
```diff
@@ -0,0 +1,10 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+class Organization:
```
yang-hai-feng, line 7: Organization should have `parent` field as well.

azext_iot/central/params.py
```diff
@@ -212,15 +227,40 @@ def load_central_arguments(self, _):
             "by visiting https://github.com/iot-for-all/iot-central-high-availability-clients#readme",
         )
 
+    with self.argument_context("iot central device twin") as context:
+        context.argument(
```
yang-hai-feng, line 231: @HangyiWang are you using --replace under `update` command or use `replace` as command?

HangyiWang: Talked with @vilit1 last week. Will use `device replace` command but not `device update -r`.

azext_iot/central/models/ga_2022_05_31/organization.py
HangyiWang: Sure. Thanks for catching.

azext_iot/central/_help.py
HangyiWang: Yes. will use `device replace` but not the `device update` with `-r` flag.

HangyiWang:
@digimaun 
Actually @vilit1 suggests me using `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module is the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.

And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`. 

Any advice here?

yang-hai-feng:
> @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> 
> And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> 
> Any advice here?

@digimaun @lucadruda @vilit1, I would prefer to using `--component-name` and `--module-name`  instead of module/component as separated group,  module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties, besides, using parameter could also simplify our backend implementations.

yang-hai-feng (approved):

digimaun:
> > @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> > And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> > Any advice here?
> 
> @digimaun @lucadruda @vilit1, I would prefer to use `--component-name` and `--module-name` instead of module/component as separated group, module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties|commands|telemetries, besides, using parameter could also simplify our backend implementations.

The reason `az iot hub digital-twin invoke-command` looks the way it does is because in IoT Hub, the invoke digitaltwins command API operation does not differentiate between devices or modules, the top level resource is just referenced as a digital twin. It is also an outlier, considering other commands specific to device or module operations are generally separated as such (az iot hub module-identity create, az iot hub device-twin replace etc).

While elevating devEx is a top factor, of course the command groups / general command surface area has to make sense and click for an IoT Central user. Ideally someone who has a 100 level understanding of Central should be able to use and reason through IoT Central commands with minimal help. We are not going to block on this particular topic if you as the Central CLI product team are happy (and feel strongly) with this updated command mapping taking into account devEx and barrier to entry as outlined earlier. The Central CLI experience should bring users joy and fullfill whatever objective (automation or otherwise) a Central user/scenario needs.

yang-hai-feng:
> > > @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> > > And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> > > Any advice here?
> > 
> > 
> > @digimaun @lucadruda @vilit1, I would prefer to use `--component-name` and `--module-name` instead of module/component as separated group, module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties|commands|telemetries, besides, using parameter could also simplify our backend implementations.
> 
> The reason `az iot hub digital-twin invoke-command` looks the way it does is because in IoT Hub, the invoke digitaltwins command API operation does not differentiate between devices or modules, the top level resource is just referenced as a digital twin. It is also an outlier, considering other commands specific to device or module operations are generally separated as such (az iot hub module-identity create, az iot hub device-twin replace etc).
> 
> While elevating devEx is a top factor, of course the command groups / general command surface area has to make sense and click for an IoT Central user. Ideally someone who has a 100 level understanding of Central should be able to use and reason through IoT Central commands with minimal help. We are not going to block on this particular topic if you as the Central CLI product team are happy (and feel strongly) with this updated command mapping taking into account devEx and barrier to entry as outlined earlier. The Central CLI experience should bring users joy and fullfill whatever objective (automation or otherwise) a Central user/scenario needs.

I see, we are using [`az iot central device edge module`](https://docs.microsoft.com/en-us/cli/azure/iot/central/device/edge/module?view=azure-cli-latest) to operate on module command, the --module-name in `az iot central device twin show --module-name`, is concept from device model, different from edge module one.

HangyiWang:
> > > @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> > > And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> > > Any advice here?
> > 
> > 
> > @digimaun @lucadruda @vilit1, I would prefer to use `--component-name` and `--module-name` instead of module/component as separated group, module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties|commands|telemetries, besides, using parameter could also simplify our backend implementations.
> 
> The reason `az iot hub digital-twin invoke-command` looks the way it does is because in IoT Hub, the invoke digitaltwins command API operation does not differentiate between devices or modules, the top level resource is just referenced as a digital twin. It is also an outlier, considering other commands specific to device or module operations are generally separated as such (az iot hub module-identity create, az iot hub device-twin replace etc).
> 
> While elevating devEx is a top factor, of course the command groups / general command surface area has to make sense and click for an IoT Central user. Ideally someone who has a 100 level understanding of Central should be able to use and reason through IoT Central commands with minimal help. We are not going to block on this particular topic if you as the Central CLI product team are happy (and feel strongly) with this updated command mapping taking into account devEx and barrier to entry as outlined earlier. The Central CLI experience should bring users joy and fullfill whatever objective (automation or otherwise) a Central user/scenario needs.

Thanks @digimaun . It won't take too much effort to add these new commands since I was mapping these APIs 1on1 to CLI commands by mistake. I just need to change few of them back.

The main point is that 'component' and 'module' is the part of the whole device twin. Each time users pass `--component-name` or `--module-name`, we'll get the whole instance twin data first and then extract the component/module data to them. If they don't specify any parameters, we'll return all twin. So the top-level resource is always the same. The difference is to return the whole data or filter part of the data based on the parameters.

digimaun:
> > > > @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> > > > And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> > > > Any advice here?
> > > 
> > > 
> > > @digimaun @lucadruda @vilit1, I would prefer to use `--component-name` and `--module-name` instead of module/component as separated group, module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties|commands|telemetries, besides, using parameter could also simplify our backend implementations.
> > 
> > 
> > The reason `az iot hub digital-twin invoke-command` looks the way it does is because in IoT Hub, the invoke digitaltwins command API operation does not differentiate between devices or modules, the top level resource is just referenced as a digital twin. It is also an outlier, considering other commands specific to device or module operations are generally separated as such (az iot hub module-identity create, az iot hub device-twin replace etc).
> > While elevating devEx is a top factor, of course the command groups / general command surface area has to make sense and click for an IoT Central user. Ideally someone who has a 100 level understanding of Central should be able to use and reason through IoT Central commands with minimal help. We are not going to block on this particular topic if you as the Central CLI product team are happy (and feel strongly) with this updated command mapping taking into account devEx and barrier to entry as outlined earlier. The Central CLI experience should bring users joy and fullfill whatever objective (automation or otherwise) a Central user/scenario needs.
> 
> Thanks @digimaun . It won't take too much effort to add these new commands since I was mapping these APIs 1on1 to CLI commands by mistake. I just need to change few of them back.
> 
> The main point is that 'component' and 'module' is the part of the whole device twin. Each time users pass `--component-name` or `--module-name`, we'll get the whole instance twin data first and then extract the component/module data to them. If they don't specify any parameters, we'll return all twin. So the top-level resource is always the same. The difference is to return the whole data or filter part of the data based on the parameters.

Understood - thanks for expanding the details here, in that case this pivot does make sense to me. Let's make sure there is enough docs / help guidance that this would be extra clear for users.

HangyiWang:
> > > > > @digimaun Actually @vilit1 suggests me use `az iot central device component twin`, `az iot central device component module twin` and etc. to return twin in different component/module for customers. But @yang-hai-feng and I feel like it's unnecessary since we'll have 4 set of different twin command with 3 different operations in this way: get, update, replace. These 12 new commands could be simplified using `az iot central device twin` only. `--component-name` and `--module-name` is enough since component and module are the part of the device twin. Users won't have any confusion if they don't want to use `--component-name` or `--module-name` in command.
> > > > > And I checked a command in iothub: `iot hub digital-twin invoke-command`. They are using `--component-path` to invoke command in device component but not using sth like `iot hub digital-twin component invoke-command`.
> > > > > Any advice here?
> > > > 
> > > > 
> > > > @digimaun @lucadruda @vilit1, I would prefer to use `--component-name` and `--module-name` instead of module/component as separated group, module/component are more preferred within device model concept, that means from device template command you can show/update detailed definition for component or module, within device, more like a scope of properties|commands|telemetries, besides, using parameter could also simplify our backend implementations.
> > > 
> > > 
> > > The reason `az iot hub digital-twin invoke-command` looks the way it does is because in IoT Hub, the invoke digitaltwins command API operation does not differentiate between devices or modules, the top level resource is just referenced as a digital twin. It is also an outlier, considering other commands specific to device or module operations are generally separated as such (az iot hub module-identity create, az iot hub device-twin replace etc).
> > > While elevating devEx is a top factor, of course the command groups / general command surface area has to make sense and click for an IoT Central user. Ideally someone who has a 100 level understanding of Central should be able to use and reason through IoT Central commands with minimal help. We are not going to block on this particular topic if you as the Central CLI product team are happy (and feel strongly) with this updated command mapping taking into account devEx and barrier to entry as outlined earlier. The Central CLI experience should bring users joy and fullfill whatever objective (automation or otherwise) a Central user/scenario needs.
> > 
> > 
> > Thanks @digimaun . It won't take too much effort to add these new commands since I was mapping these APIs 1on1 to CLI commands by mistake. I just need to change few of them back.
> > The main point is that 'component' and 'module' is the part of the whole device twin. Each time users pass `--component-name` or `--module-name`, we'll get the whole instance twin data first and then extract the component/module data to them. If they don't specify any parameters, we'll return all twin. So the top-level resource is always the same. The difference is to return the whole data or filter part of the data based on the parameters.
> 
> Understood - thanks for expanding the details here, in that case this pivot does make sense to me. Let's make sure there is enough docs / help guidance that this would be extra clear for users.

No problem. Will update this very soon.

vilit1:
also did a test run https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6033&view=results

HangyiWang:
This branch has been synced with base 'dev' branch.

vilit1:
newest test run https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6041&view=results

HISTORY.rst
```diff
@@ -36,6 +36,43 @@ unreleased
   are supported. Once registered, the device will show up in the linked IoT Hub and can be interacted with or
   simulated using other `az iot device` commands.
 
+**IoT Central updates**
```
digimaun, line 39: Since we published last week, these release notes should go under the `unreleased` section

azext_iot/central/params.py
```diff
@@ -38,7 +38,7 @@
 api_version = CLIArgumentType(
     options_list=["--api-version", "--av"],
```
digimaun, line 39: Wasn't a goal to deprecate and remove `--api-version` ? You could add the deprecation attribute in this change set, warning users of the impending removal.

linter_exclusions.yml
```diff
@@ -19,6 +16,9 @@ dt endpoint create eventhub:
     eventhub_resource_group:
       rule_exclusions:
       - parameter_should_not_end_in_resource_group
+dt data-history connection create:
+  rule_exclusions:
+      - require_wait_command_if_no_wait
```
digimaun, line 21: We should be consistent with indentation (compare indentation between other rule_exclusions entities).

HISTORY.rst
HangyiWang: Sure

digimaun (approved):

HISTORY.rst
```diff
@@ -11,6 +11,48 @@ unreleased
...
+  - az iot central api-token
+  - az iot central device-template
+  - az iot central device-group
+  - az iot central device
+  - az iot central file-upload-config
+  - az iot central organization
+  - az iot central role
+  - az iot central user
+
+* Added support for simulating device identities that use x509 thumbprint or CA authentication, impacting the
```
digimaun, line 51: This change is from 0.15.0

HangyiWang: Right.

HISTORY.rst
```diff
@@ -49,6 +81,13 @@ unreleased
 
 * Data History commands, under the `az dt data-history` command group, are now GA.
 
+**IoT Central updates**
+
+* Added support for simulating device identities that use x509 thumbprint or CA authentication, impacting the
```
digimaun, line 86: These updates go under the IoT device updates section, you could basically undo the existing removal

image

azext_iot/central/_help.py
```diff
@@ -1027,21 +1027,29 @@ def _load_central_device_groups_help():
...
+            --organizations {organization}
     """
 
     helps[
         "iot central device-group create"
     ] = """
         type: command
         short-summary: Create or update a device group.
+        long-summary: |
+                Note:  Only one organization support today, multiple organizations will be supported soon.
```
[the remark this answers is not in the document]
HangyiWang: Nice catch. Should be fixed. But I directly copied this from https://docs.microsoft.com/en-us/rest/api/iotcentral/2022-05-31dataplane/device-groups/create .

PR #539, opened 2022-06-30
export, import and migrate commands for device identities and twins on a hub

msmoot0:
---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/_params.py
```diff
@@ -761,6 +761,87 @@ def load_arguments(self, _):
...
+        context.argument(
+            "overwrite",
+            options_list=["--overwrite"],
+            help="If this flag is set, then the import will overwrite all configurations and devices in the destination hub."
+        )
+
+    with self.argument_context("iot hub state migrate") as context:
+        context.argument(
+            "dest_hub",
+            options_list=["--dest-hub", "-n"],
```
vilit1, line 817: for longer names, like destination hub, spell them out. Then the short hands usually are just the first letters of the words: --dh
for destination hub resource group -> --dhg (cause -g == resource group)

digimaun:
We should use `tqdm` to visualize progress.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,273 @@
...
+                        module2 = iot_device_module_show(cmd=self.cmd, device_id=id["deviceId"], module_id=module["module_id"], hub_name=self.orig_hub, \
+                            resource_group_name=self.orig_rg, auth_type_dataplane=self.auth_type)
+                        module["symmetricKey"] = module2["authentication"]["symmetricKey"]
+
+                    if (module["authenticationType"] == "certificateAuthority"):
+                        module2 = iot_device_module_show(cmd=self.cmd, device_id=id["deviceId"], module_id=module["module_id"], hub_name=self.orig_hub, \
+                            resource_group_name=self.orig_rg, auth_type_dataplane=self.auth_type)
+                        module["x509Thumbprint"] = module2["authentication"]["x509Thumbprint"]
+
+                    print("uploading module '{0}' in device '{1}'.".format(module["module_id"], id["deviceId"]))
```
digimaun, line 266: In most cases you do not want to use print. Instead use the logger methods like logger.info(), logger.warning() etc

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,273 @@
...
+    CLIInternalError,
+    InvalidArgumentValueError,
+    RequiredArgumentMissingError,
+)
+from azext_iot.common.shared import SdkType, JobStatusType, JobType, JobVersionType
+from azext_iot.common.utility import handle_service_exception, process_json_arg
+from azext_iot.operations.generic import _execute_query, _process_top
+from azext_iot.iothub.providers.base import IoTHubProvider, CloudError, SerializationError
+from azext_iot.operations.hub import iot_device_list, iot_device_show, iot_device_create, iot_device_delete, \
+    iot_device_twin_show, iot_device_twin_replace, iot_hub_configuration_create, iot_hub_configuration_list, \
```
digimaun, line 21: The problem with many top level functions defined in operations.hub is they are optimized for one-off command process usage rather than general re-use, making them inefficient to use without a discovery.get_target passthrough (as this method will try and look up the target IoT Hub) via login connection string (limited to key based auth), or other mechanism. In the past, as a shortcut to efficiency we split up certain operations defined in the module in two parts ...one that includes IoT Hub discovery and one that just needs the returned target dictionary. Here is an example for that https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/operations/hub.py#L108. This is not the most optimal solution, and we've iterated on vNext patterns a few times since then, but it shows the gist of a way to not look up the IoT Hub every time you call an IoT Hub service operation.

azext_iot/iothub/command_map.py
```diff
@@ -13,9 +13,7 @@
     operations_tmpl="azext_iot.iothub.commands_pnp_runtime#{}"
 )
 iothub_job_ops = CliCommandType(operations_tmpl="azext_iot.iothub.commands_job#{}")
-device_messaging_ops = CliCommandType(
```
vilit1, line 16: you need this to have the device messaging `az iot device` commands run

azext_iot/_params.py
```diff
@@ -761,6 +761,87 @@ def load_arguments(self, _):
...
+
+    with self.argument_context("iot hub state migrate") as context:
+        context.argument(
+            "dest_hub",
+            options_list=["--destination-hub", "-dh"],
+            help="Name of IoT Hub to which the origin hub will be copied."
+        )
+        context.argument(
+            "dest_rg",
+            options_list=["--destination-resource-group", "-dg"],
```
digimaun, line 822: 2 or more characters for [short] alias options should start with double dash '--'

azext_iot/iothub/commands_state.py
```diff
@@ -0,0 +1,26 @@
...
+from azext_iot.iothub.providers.job import JobProvider
+from azext_iot.iothub.providers.state import StateProvider
+
+import time
+
+logger = get_logger(__name__)
+
+
+def state_export(cmd, hub, rg=None, filename='/project/files/deviceFile.json', auth_type_dataplane=None):
+    sp = StateProvider(cmd, hub, rg, auth_type_dataplane=None)
```
avagraw, line 17: Apart from auth_type_dataplane, should we also support connection strings for all the commands?

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,277 @@
...
+
+            json.dump(configs, f)
+            f.write('\n')
+
+            for id in identities: 
+
+                module_objs = iot_device_module_list(cmd=self.cmd, device_id=id["deviceId"], hub_name=self.hub_name, resource_group_name=self.rg, auth_type_dataplane=self.auth_type)
+
+                if(id["capabilities"]["iotEdge"]):
+                    id["numModules"] = len(module_objs) - 2
```
avagraw, line 67: Why are we doing a -2 here? It will be good to add some comments across the code.

avagraw (changes requested):

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,277 @@
...
+from azext_iot.operations.hub import _iot_device_show, _iot_device_twin_show, _iot_device_module_twin_show, _iot_device_module_show
+from azext_iot.operations.hub import iot_device_list, iot_device_create, iot_device_delete, iot_device_twin_replace, iot_hub_configuration_create, iot_hub_configuration_list, \
+    iot_hub_configuration_delete, iot_device_module_list, iot_device_module_create, iot_device_module_twin_replace
+
+import json
+
+logger = get_logger(__name__)
+
+
+class StateProvider(IoTHubProvider):
```
avagraw, line 29: Please add some unit tests and integration tests for the newly added functions

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,277 @@
...
+            status_reason = None
+
+        if(auth_type == "sas"):
+            pk = identity["symmetricKey"]["primaryKey"]
+            sk = identity["symmetricKey"]["secondaryKey"]
+
+            iot_device_create(self.cmd, device_id, self.hub_name, edge, primary_key=pk, secondary_key=sk, status=status, \
+                status_reason=status_reason, resource_group_name=self.rg, auth_type_dataplane=self.auth_type)
+
+        elif(auth_type == "selfSigned"):
```
avagraw, line 130: Please use the **DeviceAuthType** Enum defined in **azext_iot\common\shared.py** for the value of _auth_method_ parameter.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,277 @@
...
+        status = identity["status"]
+        ptp = identity["x509Thumbprint"]["primaryThumbprint"]
+        stp = identity["x509Thumbprint"]["secondaryThumbprint"]
+
+        if "status_reason" in identity.keys():
+            status_reason = identity["statusReason"]
+        else:
+            status_reason = None
+
+        if(auth_type == "sas"):
```
avagraw, line 123: Please refer to **azext_iot\common\shared.py** file for auth_type enums

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,277 @@
...
+
+            iot_device_module_create(self.cmd, device_id, module_id, self.hub_name, auth_method='x509_thumbprint', primary_thumbprint=ptp, secondary_thumbprint=stp, \
+                resource_group_name=self.rg, auth_type_dataplane=self.auth_type)
+
+        elif(auth_type == "certificateAuthority"):
+
+            iot_device_module_create(self.cmd, device_id, module_id, self.hub_name, auth_method='x509_ca', resource_group_name=self.rg, auth_type_dataplane=self.auth_type)
+
+        else:
+            logger.warning("Authorization type for module '{0}' in device '{1}' not recognized.".format(module_id, device_id))
```
avagraw, line 166: Should these be logger.error, since the operation will fail for invalid auth_type

avagraw (dismissed):

azext_iot/iothub/commands_state.py
```diff
@@ -0,0 +1,26 @@
...
+from knack.log import get_logger
+from azext_iot.iothub.providers.job import JobProvider
+from azext_iot.iothub.providers.state import StateProvider
+
+import time
+
+logger = get_logger(__name__)
+
+
+def state_export(cmd, hub, rg=None, filename='/project/files/deviceFile.json', auth_type_dataplane=None):
```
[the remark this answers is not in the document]
vilit1: what if we want to be lazy (and do not care about the file name?)

azext_iot/_params.py
```diff
@@ -761,6 +761,72 @@ def load_arguments(self, _):
...
+    
+    with self.argument_context("iot hub state export") as context:
+        context.argument(
+            "filename",
+            options_list=["--filename", "-f"],
+            help="The path to the file where the hub and device information will be stored."
+        )
+
+    with self.argument_context("iot hub state import") as context:
+        context.argument(
```
vilit1, line 773: move filename to "iot hub state"

azext_iot/_params.py
```diff
@@ -761,6 +761,72 @@ def load_arguments(self, _):
...
+    with self.argument_context("iot hub state migrate") as context:
+        context.argument(
+            "hub_name",
+            options_list=["--destination-hub", "--dh"],
+            help="Name of IoT Hub to which the origin hub will be copied."
+        )
+        context.argument(
+            "rg",
+            options_list=["--destination-resource-group", "--dg"],
+            help="Name of resource group of the IoT Hub to which the origin hub will be copied."
```
vilit1, line 793: Name of resource group of the IoT Hub which the origin hub will be copied to.

^ this may be more grammatically correct. same with from sentence

azext_iot/_params.py
```diff
@@ -761,6 +761,72 @@ def load_arguments(self, _):
...
+        )
+        context.argument(
+            "login",
+            options_list=["--destination-hub-login", "--dl"],
+            validator=mode2_iot_login_handler,
+            help="This command supports an entity connection string with rights to perform action. "
```
vilit1, line 799: we will have to work on this wording - since if only one of the connection string is provided then only one session login will be avoided.

Maybe say Use to avoid session login via "az login" for this IoT Hub instance.

azext_iot/_params.py
```diff
@@ -761,6 +761,72 @@ def load_arguments(self, _):
...
+            'Use to avoid session login via "az login". '
+            "If both an entity connection string and name are provided the connection string takes priority. "
+            "Required if --dps-name is not provided.",
+            arg_group="Device Provisioning Service Identifier"
+        )
+        context.argument(
+            "orig_hub",
+            options_list=["--origin-hub", "--oh"],
+            help="Name of IoT Hub from which the origin hub will be copied."
```
vilit1, line 808: name of the IoT Hub which will be copied from

vilit1: if you want to use origin/original,
name of the original IoT Hub which will be copied from

azext_iot/_params.py
```diff
@@ -761,6 +761,72 @@ def load_arguments(self, _):
...
+            options_list=["--origin-hub-login", "--ol"],
+            validator=mode2_iot_login_handler,
+            help="This command supports an entity connection string with rights to perform action. "
+            'Use to avoid session login via "az login". '
+            "If both an entity connection string and name are provided the connection string takes priority. "
+            "Required if --dps-name is not provided.",
+            arg_group="Device Provisioning Service Identifier"
+        )
+        context.argument(
+            "overwrite",
```
vilit1, line 826: move to iot hub state

lets try to reduce duplicates in params

azext_iot/iothub/commands_state.py
```diff
@@ -0,0 +1,33 @@
...
+from azext_iot.iothub.providers.job import JobProvider
+from azext_iot.iothub.providers.state import StateProvider
+
+import time
+
+logger = get_logger(__name__)
+
+
+def state_export(cmd, filename, hub_name=None, resource_group_name=None, login=None, auth_type_dataplane=None):
+    b = time.perf_counter()
```
vilit1, line 17: make sure to remove this in the final thing - but having the times will be interesting (and good data points for your presentation)

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
+import pytest
```
vilit1, line 1: tbh this file shouldn't be in devices since it touches more than just devices, move it to core or make a folder named state. Core folder is probably best option for now.

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+    DEVICE_TYPES,
+)
+
+settings = DynamoSettings(req_env_set=ENV_SET_TEST_IOTHUB_REQUIRED, opt_env_set=ENV_SET_TEST_IOTHUB_OPTIONAL)
+
+class TestHubExportImport(IoTLiveScenarioTest):
+    def __init__(self, test_case):
+        super(TestHubExportImport, self).__init__(test_case)
+        self.dest_hub = settings.env.azext_iot_desthub or "test-hub-" + generate_generic_id()
+        self.dest_hub_rg = settings.env.azext_iot_destrg
```
vilit1, line 28: what if settings.env.azext_iot_destrg is None? What should it be then?

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+        self.dest_hub_cstring = self.cmd(
```
vilit1, line 29: you can just go into the iothub's __init__ and refactor the base class so there is a function to create an iot hub. See if it having a function do all of this for you is better (it may be messy with arguments, and if it is too messy then leave this as is)

azext_iot/iothub/commands_state.py
```diff
@@ -0,0 +1,33 @@
...
+from knack.log import get_logger
+from azext_iot.iothub.providers.job import JobProvider
+from azext_iot.iothub.providers.state import StateProvider
+
+import time
+
+logger = get_logger(__name__)
+
+
+def state_export(cmd, filename, hub_name=None, resource_group_name=None, login=None, auth_type_dataplane=None):
```
vilit1, line 16: add typing!

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+        self.kwargs["config_content"] = config_content
+        self.kwargs["labels"] = labels
+        self.kwargs["metrics"] = metrics
+
+        self.cmd("iot hub configuration create --config-id {} -n {} -g {} --content '{}' --labels '{}' --priority {} --metrics '{}' --target-condition {}" \
+            .format("hubConfig", self.entity_name, self.entity_rg, "{config_content}", "{labels}", priority, "{metrics}", "tags.bar=12"))
+
+        # populate hub with devices
+        
+        for auth_phase in DATAPLANE_AUTH_TYPES:
```
vilit1, line 71: no need for this for loop. A device created with `az iot hub device-identity create -l` and `az iot hub device-identity create -n` will be the same. Instead, focus on the different device authentication methods (sas, cert thumbprint, cert ca), like you do below

azext_iot/_params.py
msmoot0: I've had "don't end a sentence with a proposition" drilled into my head all my life but I think either is grammatically correct.

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+
+            orig_modules = self.cmd(
+                "iot hub module-identity list -d {} -n {} -g {}".format(device['deviceId'], self.entity_name, self.entity_rg)
+            ).get_output_in_json()
+            dest_modules = self.cmd(
+                "iot hub module-identity list -d {} -n {} -g {}".format(device['deviceId'], self.dest_hub, self.dest_hub_rg)
+            ).get_output_in_json()
+
+            if(device["capabilities"]["iotEdge"]):
+                assert (len(orig_modules) == 3)
```
vilit1, line 232: in python, you can do `1 == 1 == 1`

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+
+        # self.cmd(
+        #     f"iot hub state export -n {self.entity_name} -f {self.filename} -g {self.entity_rg}"
+        # )
+
+        # self.cmd(
+        #     f"iot hub state import -n {self.dest_hub} -f {self.filename} -g {self.dest_hub_rg} --overwrite"
+        # )
+
+        self.cmd(
```
vilit1, line 301: here, you would want to do the 
`for auth_phase in DATAPLANE_AUTH_TYPES:`
so you can check if your command works with different forms of authenticating to the hub

msmoot0: Okay, that was my latest question! Is it better stylistically to make those separate tests?

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,328 @@
...
+        for device in orig_hub_identities:
+            target = None
+            for id in dest_hub_identities:
+                if(device["deviceId"] == id["deviceId"]):
+                    target = id
+                    break
+
+            assert target
+
+            assert (device["authenticationType"] == target["authenticationType"])
```
vilit1, line 197: would be helpful to change some of these checks to helper functions. See if device tests already have helper functions and if not, then see if you can use the helper functions you made here for those too!

azext_iot/_params.py
vilit1: lets consult word for grammar

azext_iot/tests/iothub/devices/test_hub_export_import.py
vilit1: stylistically, you would want to have them separate. However, since the export generates the file that you will use in import, it is fine to test both in one test. I would rename the test to reflect that - test_import_export_lifecycle

msmoot0: Yeah, but would running it with different auth_phases go in different tests?

azext_iot/_help.py
```diff
@@ -576,6 +576,42 @@
...
+] = """
+    type: command
+    short-summary: Uploads the configuration and device information from a file to an IoT Hub.  
+    examples:
+    - name: Upload the devices from the specified file to the specified hub. 
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename}
+    - name: Upload the devices from the default file to the specified hub, overwriting the configurations and devices already in the hub. 
+      text: >
+        az iot hub state import -n {iothub_name} --overwrite
```
vilit1, line 601: lets make an example file within docs/samples (search for Avin's adt-bulk-import-data-sample.ndjson)

Also if you have other sample files you want to help add, feel free to do so!

linter_exclusions.yml
```diff
@@ -37,6 +37,16 @@ iot dps enrollment-group update:
     auth_type_dataplane:
       rule_exclusions:
       - no_parameter_defaults_for_update_commands
+iot hub state migrate:
+  parameters:
+    orig_resource_group_name:
+      rule exclusions:
+      - parameter_should_not_end_in_resource_group
+iot hub state migrate:
```
digimaun, line 45: `iot hub state migrate` is defined twice, you should put both parameters and their exclusions in the same bucket.

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
         -m {module_name} -j ../mymodtwin.json
 """
 
+helps[
+    "iot hub state"
+] = """
+    type: group
+    short-summary: Download, upload or migrate the state of an IoT Hub.
```
vilit1, line 583: Manage the state of an IoT Hub

we want to make sure the wording would allow extra features if we decide to add them (ex: reset an iot hub)

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
...
+"""
+
+helps[
+    "iot hub state export"
+] = """
+    type: command
+    short-summary: Saves the configuration and device information of an IoT Hub to a file.
```
vilit1, line 590: Save

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
...
+    - name: Export the devices in the specified hub to the specified file.
+      text: >
+        az iot hub state export -n {iothub_name} -f {filename}
+"""
+
+helps[
+    "iot hub state import"
+] = """
+    type: command
+    short-summary: Uploads the configuration and device information from a file to an IoT Hub.
```
vilit1, line 601: Upload

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
...
+    - name: Upload the devices from the default file to the specified hub, overwriting the configurations and devices already in the hub.
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename} -r
+"""
+
+helps[
+    "iot hub state migrate"
+] = """
+    type: command
+    short-summary: Copies the configuration and device information from one hub to another hub without saving to a file.
```
vilit1, line 615: Copy

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
...
+    examples:
+    - name: Migrate the state of the original hub to the new hub, overwriting the configurations and devices already in the hub.
```
vilit1, line 617: add an example without the -r flag

vilit1: also change this example (the one with -r) to show all the params - so you can show that you can do this with hubs in different rgs

azext_iot/_help.py
vilit1: also say device information and device twin - they are technically different terminology

azext_iot/_params.py
```diff
@@ -772,6 +772,60 @@ def load_arguments(self, _):
...
+    with self.argument_context("iot hub state") as context:
+        context.argument(
+            "filename",
+            options_list=["--filename", "-f"],
+            help="The path to the file where the hub and device information will be stored."
+        )
+        context.argument(
+            "replace",
+            options_list=["--replace", "-r"],
+            help="If this flag is set, then the import will overwrite all configurations and devices in the destination hub."
```
vilit1, line 784: the import will? you can just say:
If this flag is set, then all configurations and devices in the destination hub will be overwritten.

Although this is passive tense. We many need to consult word. If you want to make it active - then use command instead of import

azext_iot/_params.py
```diff
@@ -772,6 +772,60 @@ def load_arguments(self, _):
...
+        )
+        context.argument(
+            "login",
+            options_list=["--destination-hub-login", "--dl"],
+            validator=mode2_iot_login_handler,
+            help="This command supports an entity connection string with rights to perform action on the destination hub. "
+            'Use to avoid session login via "az login" for this IoT Hub instance. '
+            "If both an entity connection string and name are provided the connection string takes priority. "
+            "Required if --dps-name is not provided.",
+            arg_group="Device Provisioning Service Identifier"
```
vilit1, line 806: I know this is copy paste but this is iot hub not dps.

azext_iot/assets/user_messages.py
```diff
@@ -11,10 +11,13 @@
...
-        "Please provide an {0} entity name (via the '{1}' or '-n' parameter)"
-        " or {0} connection string via --login..."
-    ).format(entity_type, PARAM_MAPPING[entity_type])
+    if entity_type:
+        return (
+            "Please provide an {0} entity name (via the '{1}' or '-n' parameter)"
+            " or {0} connection string via --login..."
+        ).format(entity_type, PARAM_MAPPING[entity_type])
+    else:
+        return ("Required entity type missing, please check _validators.py for required entities.")
```
digimaun, line 20: This is a CLI tool not an sdk, it doesn't make sense to require a user to figure out how a CLI extension is installed, then inspect a python module to understand how validation works. If this condition is hit from a use case, then it should be re-tooled or refactored such that the user will have a clear understanding of what is required.

azext_iot/iothub/command_map.py
```diff
@@ -13,6 +13,8 @@
     operations_tmpl="azext_iot.iothub.commands_pnp_runtime#{}"
 )
 iothub_job_ops = CliCommandType(operations_tmpl="azext_iot.iothub.commands_job#{}")
+iothub_state_ops = CliCommandType(operations_tmpl="azext_iot.iothub.commands_state#{}")
+
```
vilit1, line 17: you can remove the newline

azext_iot/iothub/commands_state.py
```diff
@@ -0,0 +1,47 @@
...
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azext_iot.iothub.providers.state import StateProvider
+
+
+def state_export(
+    cmd,
+    filename,
```
vilit1, line 12: typing

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+                "mymetric": "SELECT deviceId from devices where properties.reported.lastDesiredStatus.code = 200"
+            }
+        }}
+        metrics = json.dumps(metrics)
+
+        self.kwargs["config_content"] = config_content
+        self.kwargs["labels"] = labels
+        self.kwargs["metrics"] = metrics
+
+        self.cmd("iot hub configuration create --config-id {} -l {} --content '{}' --labels '{}' --priority {} --metrics '{}' \
```
vilit1, line 65: ah this tabbing is weird - stay consistent (like the tabbing/spacing for metrics)

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+                self.kwargs["patch_tags"] = json.dumps(patch_tags)
+
+                self.cmd(
+                    f"iot hub device-twin update -d {device} -l {self.connection_string}"
+                    " --tags '{patch_tags}'"
+                )
+
+                # make a module for each device, same authentication method as device
+
+                if(i == 0):
```
vilit1, line 120: rather than putting this in the for loop (and awkwardly trying to force it to only occur on specific devices), why not put it right after you create the devices?

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+        assert(device1["authenticationType"] == device2["authenticationType"])
+        assert(device1["capabilities"]["iotEdge"] == device2["capabilities"]["iotEdge"])
+        assert(device1["connectionState"] == device2["connectionState"])
+        assert(device1["status"] == device2["status"])
+
+        if("tags" in device1.keys()):
+            assert(device1["tags"] == device2["tags"])
+
+        if(device1["authenticationType"] == DeviceAuthApiType.sas.value):
+            id1 = self.cmd("iot hub device-identity show -l {} -d {}".format(self.connection_string, device1['deviceId'])) \
```
vilit1, line 173: try to keep formatting consistent (like line 149)

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+            module_twin = self.cmd(
+                f"iot hub module-twin show -m {module['moduleId']} -d {device1['deviceId']} -l {self.connection_string}"
+            ).get_output_in_json()
+            target_module_twin = self.cmd(
+                f"iot hub module-twin show -m {module['moduleId']} -d {device1['deviceId']} -l {self.dest_hub_cstring}"
+            ).get_output_in_json()
+
+            assert(len(module_twin["properties"]["desired"]) == len(target_module_twin["properties"]["desired"]))
+            for prop in module_twin["properties"]["desired"]:
+                if(prop != "$metadata" and prop != "version"):
```
vilit1, line 226: prop not in ["$metadata", "version"] would be easier - especially if the service decides to make another default prop

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+                    assert(prop in target_module_twin["properties"]["desired"])
+                    assert(module_twin["properties"]["desired"][prop] == target_module_twin["properties"]["desired"][prop])
+
+            if("tags" in module_twin.keys()):
```
vilit1, line 230: "tags" in module_twin

when working with dicts, this will check the keys.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+        if login or auth_type_dataplane == "key":
+            self.login = self.target["cs"]
+        else:
+            self.login = None
+
+    def save_state(self, filename: str):
+        '''
+        Writes all hub configurations, device identities and device twins from the origin hub to a json file
+        '''
+
```
digimaun, line 52: Edge deployments are missing. `iot_hub_configuration_list` gets you only adm configurations.

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+            self.cmd(
+                "iot hub device-identity delete -d {} -l {}".format(id["deviceId"], self.dest_hub_cstring)
+            )
+
+        # gives the api enough time to update
+        time.sleep(1)
+
+    def test_import_export(self):
+
+        # auth-type = key (default)
```
vilit1, line 295: you can use the `for auth_phase in DATAPLANE_AUTH_TYPES:` and `self.set_cmd_auth_type` to do the whole auth type mess for you

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,364 @@
...
+
+        self.dest_hub_cstring = self.cmd(
+            "iot hub connection-string show -n {} -g {}".format(self.dest_hub, self.dest_hub_rg)
+        ).get_output_in_json()["connectionString"]
+
+        self.filename = generate_generic_id() + ".json"
+
+        self.create_hub_state()
+
+    def create_hub_state(self):
```
digimaun, line 40: Tests need to exercise edge deployments.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+        configs = iot_hub_configuration_list(cmd=self.cmd, hub_name=self.hub_name, resource_group_name=self.rg,
```
vilit1, line 53: you have gotten the target in the __init__ (specifically, the IoTHubProvider init) - why not use it? If a helper function that just takes in target doesnt exist, make it. 

same for identities

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+
+        identities = iot_device_list(cmd=self.cmd, hub_name=self.hub_name, top=-1, resource_group_name=self.rg,
+                                     login=self.login, auth_type_dataplane=self.auth_type)
+
+        with open(filename, 'w', encoding='utf-8') as f:
+
+            json.dump(configs, f)
+            f.write('\n')
+
+            for i in tqdm(range(len(identities)), desc="Exporting devices"):
```
vilit1, line 64: include tqdm for config and identities - you can just have different instances (so it would be one tqdm for all the configs, followed by another for identities, and so on)

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+        Writes all hub configurations, device identities and device twins from the origin hub to a json file
+        '''
+
+        configs = iot_hub_configuration_list(cmd=self.cmd, hub_name=self.hub_name, resource_group_name=self.rg,
+                                             login=self.login, auth_type_dataplane=self.auth_type)
+
+        identities = iot_device_list(cmd=self.cmd, hub_name=self.hub_name, top=-1, resource_group_name=self.rg,
+                                     login=self.login, auth_type_dataplane=self.auth_type)
+
+        with open(filename, 'w', encoding='utf-8') as f:
```
vilit1, line 59: write to file at the end, once you have everything

docs/samples/hub-state-import-data-sample.json
```diff
@@ -0,0 +1,13 @@
+[{"id": "my_config", "schemaVersion": "2.0", "labels": null, "content": {"deviceContent": {"properties.desired.awesome": 3}}, "targetCondition": "tags.environment='test'", "createdTimeUtc": "2022-06-28T22:56:32.937713Z", "lastUpdatedTimeUtc": "2022-06-28T22:56:32.937713Z", "priority": 10, "systemMetrics": {"results": {"targetedCount": 0, "appliedCount": 0}, "queries": {"targetedCount": "select deviceId from devices where tags.environment='test'", "appliedCount": "select deviceId from devices where configurations.[[my_config]].status = 'Applied'"}}, "metrics": {"results": {}, "queries": {}}, "etag": "MQ=="}, {"id": "my_config2", "schemaVersion": "2.0", "labels": null, "content": {"deviceContent": {"properties.desired.prop2": 5}}, "targetCondition": "*", "createdTimeUtc": "2022-06-28T23:03:31.8831286Z", "lastUpdatedTimeUtc": "2022-06-28T23:03:31.8831286Z", "priority": 7, "systemMetrics": {"results": {"targetedCount": 3, "appliedCount": 3}, "queries": {"targetedCount": "select deviceId from devices", "appliedCount": "select deviceId from devices where configurations.[[my_config2]].status = 'Applied'"}}, "metrics": {"results": {}, "queries": {}}, "etag": "MQ=="}]
```
digimaun, line 1: I'm not a fan of this data structure for holding hub state. Ideally it should be an object with aptly named keys holding the various aspects of state.

i.e.
```
{
  "devices": [{"identity": {}, "twin": {}}],
  "modules": [{"identity": {}, "twin": {}}],
  "configurations": [{}, {}]
  ...
  ...
}
```

That is just an example, but you can see with hub state starting with a root level object, it is easier to reason through its attributes and actually use/iterate through it.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+
+                # this may be redundant, since it seems like the same fields are included in the identity
+                twin = _iot_device_twin_show(self.target, id["deviceId"])
+                json.dump(twin, f)
+                f.write('\n')
+
+                for module in module_objs:
+                    module = vars(module)
+
+                    if(module["module_id"] not in ["$edgeAgent", "$edgeHub"]):
```
vilit1, line 93: rework your if statements to be more pythonic:
if module["module_id"] not in ["$edgeAgent", "$edgeHub"]:

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+
+                        module.pop('connection_state_updated_time')
```
vilit1, line 95: add a comment as to why you are popping/removing these

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+        if "status_reason" in identity.keys():
+            status_reason = identity["statusReason"]
+        else:
+            status_reason = None
+
+        if(auth_type == DeviceAuthApiType.sas.value):
+            pk = identity["symmetricKey"]["primaryKey"]
+            sk = identity["symmetricKey"]["secondaryKey"]
+
+            iot_device_create(self.cmd, device_id, self.hub_name, edge, primary_key=pk, secondary_key=sk, status=status,
```
vilit1, line 134: use target rather than having the function find the hub again

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,342 @@
...
+                                               hub_name=self.hub_name, login=self.login, auth_type_dataplane=self.auth_type)
+
+                i += 2
+                pbar.update(2)
+
+        pbar.close()
+
+        logger.info("Uploaded state from '{}' to IoT Hub '{}'".format(filename, self.hub_name))
+
+    def migrate_devices(
```
vilit1, line 264: if you rework how you export the state (specifically get everything you need before you write to the file), you can turn that into a helper function that would help you out here

azext_iot/assets/user_messages.py
c-ryan-k, line 20: Might be a nit - but I think I recall the CLI core linter using the pylint [`no_else_return`](https://pylint.pycqa.org/en/latest/user_guide/messages/refactor/no-else-return.html) rule - basically we don't need the `else` here, since the code is either going to return inside of the `if`, or on the next line anyway.

This didn't seem to flag in our PR pipeline, so it might be a non-issue, but just a heads up in case we see it later.

azext_iot/_params.py
msmoot0: Right - I copied this over and forgot to make it command-neutral!

azext_iot/_help.py
digimaun: Just curious any issue with using 'Export(s) ...' in the description matching the command verb? Export is a fine word.

In the short-summary saying something like

"Exports the state of an IoT Hub to a file."

Then optionally the long summary could spell out what the state entails.

"The exported state will include device and module identities as well as their twin tags and desired properties, ADM configurations, edge deployments, .... ' etc etc.

azext_iot/_help.py
digimaun, line 601: In this import command and migrate command descriptions (as well as export) the command verb isn't being used in the summary description. If 'upload' is a more clear action compared to import then the command action can be renamed to upload so verbiage is consistent....but I think import/export/migrate are totally fine.

digimaun: Overall point being we should be consistent in terminology with descriptions.

azext_iot/iothub/providers/state.py
digimaun: `_iot_hub_configuration_list` will get you both adm and edge.

azext_iot/iothub/providers/state.py
msmoot0: even though the configs are all just being dumped at once? there's no iteration happening there

azext_iot/assets/user_messages.py
vilit1: This is an error message that only a developer should see (especially if they are working with offline commands and forget to update the validator). Would rewording it to "Required entity type missing." be enough for now?

azext_iot/_help.py
vilit1: Agree with paymaun - make the short summary just say that it will save the state and the long summary detail what the state is

azext_iot/_help.py
```diff
@@ -576,6 +576,49 @@
...
+    short-summary: Download, upload or migrate the state of an IoT Hub.
+"""
+
+helps[
+    "iot hub state export"
+] = """
+    type: command
+    short-summary: Saves the configuration and device information of an IoT Hub to a file.
+    examples:
+    - name: Export the devices in the specified hub to the specified file.
```
vilit1: make sure you fix the wording here too cause you aren't just exporting the devices

azext_iot/_help.py
```diff
@@ -576,6 +576,57 @@
...
+    - name: Export the configurations and devices in the specified hub to the specified file.
+      text: >
+        az iot hub state export -n {iothub_name} -f {filename}
+"""
+
+helps[
+    "iot hub state import"
+] = """
+    type: command
+    short-summary: Import configurations, device information, and device twins from a file to an IoT Hub.
```
vilit1, line 603: make sure you have short and long summary here too

azext_iot/_help.py
```diff
@@ -576,6 +576,57 @@
...
+    examples:
+    - name: Import the devices from the specified file to the specified hub.
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename}
+    - name: Import the devices from the default file to the specified hub, overwriting the configurations and devices already in
```
vilit1, line 608: you arent importing just the devices, no?

azext_iot/_help.py
```diff
@@ -576,6 +576,57 @@
...
+
+helps[
+    "iot hub state migrate"
+] = """
+    type: command
+    short-summary: Copy the state of one hub to another hub without saving to a file.
+    long-summary: The migrated state will include hub configurations, device identities and twins, and module identities and
+                  twins.
+    examples:
+    - name: Migrate the state of the original hub to the new hub.
```
vilit1, line 622: change new to destination in here and the next example

azext_iot/_params.py
```diff
@@ -295,6 +295,16 @@ def load_arguments(self, _):
             options_list=["--auth-type"],
             arg_type=hub_auth_type_dataplane_param_type,
         )
+        context.argument(
+            "filename",
+            options_list=["--filename", "-f"],
+            help="The path to the file where the hub and device information will be stored."
```
vilit1, line 301: lets just use state information

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,310 @@
...
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from knack.log import get_logger
+
+from azext_iot.common.shared import DeviceAuthApiType, DeviceAuthType, ConfigType
+from azext_iot.iothub.providers.base import IoTHubProvider
+from azext_iot.operations.hub import _iot_device_show, _iot_device_module_twin_show, _iot_device_module_create, \
```
vilit1, line 11: you can use () here too, look at other files that have many imports from a single module

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,310 @@
...
+            cmd=cmd,
+            hub_name=hub,
+            rg=rg,
+            login=login,
+            auth_type_dataplane=auth_type_dataplane
+        )
+        self.auth_type = auth_type_dataplane
+
+        # the login argument doesn't work later if auth type = "login" and a connection string isn't given initially
+        if login or auth_type_dataplane == "key":
```
vilit1, line 42: is this still needed if you use the target?

azext_iot/operations/hub.py
```diff
@@ -148,6 +154,20 @@ def iot_device_list(
     return result
 
 
+def _iot_device_list(hub_name, target, edge_enabled=False, top=1000):
```
vilit1, line 157: target should have hub name already - fix this and the other helper functions you added

azext_iot/operations/hub.py
```diff
@@ -148,6 +154,20 @@ def iot_device_list(
...
+    top = _process_top(top)
+    query = (
+        "select * from devices where capabilities.iotEdge = true"
+        if edge_enabled
+        else "select * from devices"
+    )
+    result = _iot_query(target, query, top)
```
vilit1, line 164: make sure you include the argument name (like target=target, etc) so that if someone screws up the order of the params, it wouldnt break everything.

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,486 @@
...
+            --target-condition {}".format(self.connection_string, "{edge_content1}", "{labels}", random.randint(1, 10),
+                                          "{metrics}", "tags.bar=12")
+        )
+
+        # make a layered edge deployment
+        deployment2_path = os.path.dirname(__file__) + "\\..\\configurations\\test_edge_deployment_layered.json"
+        self.kwargs["edge_content2"] = read_file_content(deployment2_path)
+        self.cmd(
+            "iot edge deployment create -d deployment2 -l {} --content '{}' --labels '{}' --priority {} --metrics '{}' \
+            --target-condition {} --layered".format(self.connection_string, "{edge_content1}", "{labels}", random.randint(1, 10),
```
vilit1, line 75: you can put another enter after the ( for more pythonic (or at least consistent) spacing

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,486 @@
...
+            self.cmd(
+                f"iot hub device-identity create -d {device_ids[1]} -l {self.connection_string} --am x509_ca {edge_enabled}"
+            )
+            self.cmd(
+                f"iot hub module-identity create -m deviceModule -d {device_ids[1]} -l {self.connection_string} --am x509_ca"
+            )
+
+            # create x509_thumbprint device and module
+            self.cmd(
+                f"iot hub device-identity create -d {device_ids[2]} -l {self.connection_string} --am x509_thumbprint \
```
vilit1, line 109: just close the string and start a new string with f

if you use \ within strings, it looks nice till you realize that there is extra whitespace (and getting rid of the whitespace makes things ugly)

vilit1: make sure you fix the other ones too

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,310 @@
...
+
+    def migrate_devices(
+        self,
+        orig_hub: Optional[str] = None,
+        orig_rg: Optional[str] = None,
+        orig_hub_login: Optional[str] = None,
+        replace: Optional[bool] = False,
+    ):
+
+        if replace:
```
vilit1, line 296: do the replace after you get the state for the original hub - think of what happens if the two hubs are the same

azext_iot/iothub/providers/state.py
msmoot0: is it better practice to then put each import on a new line? I see that's what most of the other files are doing, my only reservation is that... that would be a whole lot of lines

azext_iot/operations/hub.py
msmoot0: target does not have the hub name directly, but it does have entity name, which includes hub name. for my hub, entity name is "ms-hub.azure-devices.net". Is entity name guaranteed to always be of the form "hubname.azure-devices.net"?

msmoot0: oh, seems like it is

azext_iot/iothub/providers/state.py
vilit1: yeah but it is better practice - plus most editors will let you collapse things within a block/grouped in parentheses

azext_iot/operations/hub.py
vilit1: you are guaranteed that the stuff before the first . is the hub name - but the hostname (azure-devices.net) is not guaranteed for preview environments and features

azext_iot/tests/iothub/devices/test_hub_export_import.py
```diff
@@ -0,0 +1,490 @@
...
+    SECONDARY_THUMBPRINT,
+    DATAPLANE_AUTH_TYPES,
+    DEVICE_TYPES,
+)
+
+settings = DynamoSettings(req_env_set=ENV_SET_TEST_IOTHUB_REQUIRED, opt_env_set=ENV_SET_TEST_IOTHUB_OPTIONAL)
+
+
+class TestHubExportImport(IoTLiveScenarioTest):
+    def __init__(self, test_case):
```
avagraw, line 23: Since, these are integration tests, hence the file name should end with **_int.py**. Also we should run the tests on the pipeline to have evidence that the tests run as expected. Happy to help you setup the pipeline run.

azext_iot/tests/iothub/__init__.py
```diff
@@ -60,61 +60,7 @@ def __init__(self, test_scenario):
...
-                        )
-                    )
-                    sleep(10)
-
-                if tries == MAX_RBAC_ASSIGNMENT_TRIES:
-                    raise Exception(
-                        "Reached max ({}) number of tries to assign RBAC role. Please re-run the test later "
-                        "or with more max number of tries.".format(MAX_RBAC_ASSIGNMENT_TRIES)
-                    )
+            self.create_hub(self.entity_name, self.entity_rg, self.storage_container, self.storage_cstring)
```
avagraw, line 63: The **self.storage_container** and  **self.storage_cstring** attributes exist only for storage tests. Hence, the other tests will fail. So, we need a conditional logic here.

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -45,8 +47,13 @@ def create_hub_state(self):
         labels = {generate_generic_id() : generate_generic_id(), generate_generic_id() : generate_generic_id()}
         labels = json.dumps(labels)
 
-        metrics_path = os.path.dirname(__file__) + "\\..\\configurations\\test_config_generic_metrics.json"
-        content_path = os.path.dirname(__file__) + "\\..\\configurations\\test_adm_device_content.json"
+        metrics_path = os.path.join(Path(CWD), "../configurations/test_config_generic_metrics.json")
```
avagraw, line 50: Let us make the path OS agnostic by doing the join like:
 _**os.path.join(Path(CWD).parent, "configurations", "test_config_generic_metrics.json")**_

azext_iot/_params.py
```diff
@@ -772,6 +782,48 @@ def load_arguments(self, _):
...
+            help="Name of resource group of the IoT Hub to which the origin hub will be copied."
+        )
+        context.argument(
+            "login",
+            options_list=["--destination-hub-login", "--dl"],
+            validator=mode2_iot_login_handler,
+            help="This command supports an entity connection string with rights to perform action on the destination hub. "
+            'Use to avoid session login via "az login" for this IoT Hub instance. '
+            "If both an entity connection string and name are provided the connection string takes priority. "
+            "Required if --hub-name is not provided.",
```
vilit1, line 803: required if what is not provided?

azext_iot/_params.py
```diff
@@ -772,6 +782,48 @@ def load_arguments(self, _):
...
+            help="Name of resource group of the IoT Hub which will be copied."
+        )
+        context.argument(
+            "orig_hub_login",
+            options_list=["--origin-hub-login", "--ol"],
+            validator=mode2_iot_login_handler,
+            help="This command supports an entity connection string with rights to perform action on the origin hub. "
+            'Use to avoid session login via "az login" for this IoT Hub instance. '
+            "If both an entity connection string and name are provided the connection string takes priority. "
+            "Required if --hub-name is not provided.",
```
vilit1, line 823: required if what is not provided?

Also not sure if this should be underneath a grouping since the origin hub isnt under it

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -0,0 +1,498 @@
...
+
+        # make a configuration for the hub (applies to 0 devices, this is just to test the configuration settings)
+
+        labels = {generate_generic_id() : generate_generic_id(), generate_generic_id() : generate_generic_id()}
+        labels = json.dumps(labels)
+
+        metrics_path = os.path.join(Path(CWD), "..", "configurations", "test_config_generic_metrics.json")
+        content_path = os.path.join(Path(CWD), "..", "configurations", "test_adm_device_content.json")
+
+        print()
```
vilit1, line 53: prints

vilit1 (approved):
Looks good to me - just fix some param comments, remove prints, and make sure you run the pipeline tests - post the link to them as a comment

azext_iot/iothub/command_map.py
```diff
@@ -47,3 +48,8 @@ def load_iothub_commands(self, _):
         cmd_group.command("receive", "iot_c2d_message_receive")
         cmd_group.command("send", "iot_c2d_message_send")
         cmd_group.command("purge", "iot_c2d_message_purge")
+
+    with self.command_group("iot hub state", command_type=iothub_state_ops, is_preview=True) as cmd_group:
```
vilit1, line 52: comment this out to hide it - make sure you try `az iot hub -h`

msmoot0:
Pipeline: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6232&view=results

avagraw:
Testing evidence with the latest changes: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6237&view=logs&j=972448a5-d3b3-5e41-ccf5-40735d33ce04&t=c4c762d4-7c09-545b-b56d-1258f342cf0c

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,319 @@
...
+        orig_hub_target = self.discovery.get_target(
+            resource_name=orig_hub,
+            resource_group_name=orig_rg,
+            login=orig_hub_login,
+            auth_type=self.auth_type
+        )
+
+        if replace:
+            self.delete_all_configs()
+            self.delete_all_devices()
```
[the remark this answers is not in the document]
c-ryan-k: Ignore the last part - `-r` **is** the confirmation from the user that this command is destructive.

Still wondering about the first part though - whether or not it makes sense to log a warning that the "hub they're importing state to" is not empty.

azext_iot/iothub/providers/state.py
```diff
@@ -0,0 +1,319 @@
...
+        '''
+        Uses device info from file to recreate the devices
+        '''
+
+        if replace:
+            self.delete_all_configs()
+            self.delete_all_devices()
+
+        with open(filename, 'r', encoding='utf-8') as f:
+            hub_state = json.load(f)
```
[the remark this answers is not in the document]
c-ryan-k: One way we could do this is similar to how the import command has the `-r` switch to delete before writing, basically an `--overwrite` switch when exporting to a file that will overwrite the target file if it exists.

If the user doesn't provide the `--overwrite` switch, maybe throw an error if the file already exists?

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -0,0 +1,496 @@
...
+class TestHubExportImport(IoTLiveScenarioTest):
+    def __init__(self, test_case):
+        super(TestHubExportImport, self).__init__(test_case)
+        self.dest_hub = settings.env.azext_iot_desthub or "test-hub-" + generate_generic_id()
+        self.dest_hub_rg = settings.env.azext_iot_destrg or settings.env.azext_iot_testrg
+
+        # create destination hub
+
+        if not settings.env.azext_iot_desthub:
+            self.create_hub(self.dest_hub, self.entity_rg)
```
vilit1, line 34: Is this hub being torn down at the end of tests? Make sure you add this to teardown.

Also would help if you added a test tag (look how the __init__ adds a test tag at the end - it will just add a tag to tell us that this test was run x amount of times, very helpful if something breaks and resources are still there or if you are using your own hub and you want to keep track of what you do to it)

msmoot0: Yes, if the hub is created here then it is torn down at the end.

avagraw:
The tests are adding ~25 mins to the pipeline run, even though they run in 2-3 minutes locally, this will have to be investigated before we can merge the commands.

msmoot0:
Latest pipeline run: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6252&view=logs&s=859b8d9a-8fd6-5a5c-6f5e-f84f1990894e

docs/samples/hub-state-import-data-sample.json
```diff
@@ -0,0 +1,925 @@
...
+          "reportedFailedCount": 0,
+          "reportedSuccessfulCount": 0,
+          "targetedCount": 0
+        }
+      },
+      "targetCondition": ""
+    }
+  ],
+  "modules": {
+    "device1-child": [],
```
digimaun, line 645: I don't see the benefit of modules being key'd by device Id. The device Id is already part of a module identity. This adds significant bloat - imagine a hub with 1 million devices and no modules. We'd have 2 million device Id entries representing the 1 million devices.

docs/samples/hub-state-import-data-sample.json
```diff
@@ -0,0 +1,925 @@
+{
+  "children": {
```
digimaun, line 2: Shouldn't this be encapsulated under a device centric element rather than at the top level?

docs/samples/hub-state-import-data-sample.json
```diff
@@ -0,0 +1,925 @@
+{
```
digimaun, line 1: In general the payload representing the state of an IoT Hub should only include what is necessary to recreate it. This sample of the payload should include the required elements and no extraneous ones.

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -0,0 +1,506 @@
...
+
+        self.dest_hub_cstring = self.cmd(
+            f"iot hub connection-string show -n {self.dest_hub} -g {self.dest_hub_rg}"
+        ).get_output_in_json()["connectionString"]
+
+        self.filename = generate_generic_id() + ".json"
+
+        self.create_hub_state()
+
+    def create_hub_state(self):
```
digimaun, line 54: We need to test more states than just one loaded state.  Pytest fixtures and parameterized testing can help iterate through and testing states faster.

azext_iot/_params.py
```diff
@@ -295,6 +295,21 @@ def load_arguments(self, _):
...
+            help="The path to the file where the state information will be stored."
+        )
+        context.argument(
+            "replace",
+            options_list=["--replace", "-r"],
+            help="If this flag is set, then the command will overwrite the state of the destination hub."
+        )
+        context.argument(
+            "overwrite_file",
+            options_list=["--overwrite-file", "--of"],
```
digimaun, line 310: `--force` is common for this, and not all arguments need short options.

azext_iot/assets/user_messages.py
```diff
@@ -11,10 +11,12 @@
...
-    return (
-        "Please provide an {0} entity name (via the '{1}' or '-n' parameter)"
-        " or {0} connection string via --login..."
-    ).format(entity_type, PARAM_MAPPING[entity_type])
+    if entity_type:
+        return (
+            "Please provide an {0} entity name (via the '{1}' or '-n' parameter)"
+            " or {0} connection string via --login..."
+        ).format(entity_type, PARAM_MAPPING[entity_type])
+    return ("Please provide either a hub name or a DPS name.")
```
digimaun, line 19: Entity type is not supposed to be optional / accepting of `None`. If `None` is provided than a `ValueError` can be thrown.

azext_iot/_params.py
```diff
@@ -295,6 +295,21 @@ def load_arguments(self, _):
             options_list=["--auth-type"],
             arg_type=hub_auth_type_dataplane_param_type,
         )
+        context.argument(
+            "filename",
```
digimaun, line 299: What if we changed this to `--state-file` to be a little less generic?

azext_iot/_help.py
```diff
@@ -576,6 +576,63 @@
...
+    short-summary: Import a Hub state from a file to an IoT Hub.
+    long-summary: The imported state will include hub configurations, edge deployments, device identities and twins, and module
+                  identities and twins.
+    examples:
+    - name: Import the state from the specified file to the specified hub.
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename}
+    - name: Import the state from the default file to the specified hub, overwriting the previous state of the hub.
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename} -r
```
digimaun, line 616: I think we should spell out `--replace` in this example (and remember to align on the action description,  overwrite -> replace).

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -0,0 +1,506 @@
...
+
+        # make a configuration for the hub (applies to 0 devices, this is just to test the configuration settings)
+
+        labels = {generate_generic_id() : generate_generic_id(), generate_generic_id() : generate_generic_id()}
+        labels = json.dumps(labels)
+
+        metrics_path = os.path.join(Path(CWD), "..", "configurations", "test_config_generic_metrics.json")
+        content_path = os.path.join(Path(CWD), "..", "configurations", "test_adm_device_content.json")
+
+        self.kwargs["config_content"] = read_file_content(content_path)
```
digimaun, line 64: We should test with adm module configuration as well.

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
msmoot0: This might have to be taken care of after I leave. I don't think I'll have time to add in several more testing states.

PR #556, opened 2022-07-29
Hub endpoint control plane

vilit1:
~Waiting for merge - need to have the newer SDK with the Cosmos DB feature released due to non-backwards compatible API.~

Edit: since 10/25, it seems like this is no longer the case and the api is now backwards compatible. Testing shows that it should be more or less satisfactory aside from being able to create and test routes with the cosmos db endpoint but not show/list/update/delete them. Service is dealing with that.

Edit: 11/1, we will go back to waiting on the Azure CLI core to have the new sdk so we don't have the extension fight with the core for sdk versions.

Addition of Cosmos DB Endpoint creation! 

Essentially moving over endpoint and route management right now. The new command groups are:
- az iot hub message-endpoint
- az iot hub message-route

Sadly the sdk used still has models (endpoints's UserAssignedIdentities property does not work well with modelless)

Endpoints creation is broken up, like so:
- az iot hub messaging-endpoint create cosmosdb-collection
- az iot hub messaging-endpoint create eventhub
- az iot hub messaging-endpoint create servicebus-topic
- az iot hub messaging-endpoint create servicebus-queue
- az iot hub messaging-endpoint create storage-container

This helps isolate which params are needed for which specific endpoints. 

Other quality of life fixes:
- Improved tests for endpoints! Now all types of endpoints are created (including service types and auth types)!
- Fixed `route test` app-properties and system-properties to correctly try to change the format of those arguments to json before sending it to the service *
- Changed how endpoint delete works (especially when specifying both the name and type)
- Added connection string lookup for endpoint creation
- Added test fixtures

Notes: * these will be fixed in the azure cli core commands too.

Also added a small fix in the dynamosettings, if the value is "sentinel" then it is nulled out. This value gets set if there is misconfigured pipeline variable (especially if adding new pipeline variables).

---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,281 @@
...
+                endpoints.service_bus_queues = remaining_endpoints
+            elif any(e.name.lower() == endpoint_name.lower() for e in endpoints.service_bus_topics):
+                remaining_endpoints = [e for e in endpoints.service_bus_topics if e.name.lower() != endpoint_name.lower()]
+                endpoints.service_bus_topics = remaining_endpoints
+            elif any(e.name.lower() == endpoint_name.lower() for e in endpoints.cosmos_db_sql_collections):
+                remaining_endpoints = [e for e in endpoints.cosmos_db_sql_collections if e.name.lower() != endpoint_name.lower()]
+                endpoints.cosmos_db_sql_collections = remaining_endpoints
+            elif any(e.name.lower() == endpoint_name.lower() for e in endpoints.storage_containers):
+                remaining_endpoints = [e for e in endpoints.storage_containers if e.name.lower() != endpoint_name.lower()]
+                endpoints.storage_containers = remaining_endpoints
```
c-ryan-k, line 261: I think the `any` checks combined with the `elif`s here are to make sure we don't delete multiple endpoints that happen to have the same name - but I don't believe IoT Hub allows you to have multiple endpoints with the same name:

image

image


In that case, we might be able to save a bunch of logic here by just changing it to:
```suggestion
            endpoints.event_hubs = [e for e in endpoints.event_hubs if e.name.lower() != endpoint_name.lower()]
            endpoints.service_bus_queues =  [e for e in endpoints.service_bus_queues if e.name.lower() != endpoint_name.lower()]
            endpoints.service_bus_topics = [e for e in endpoints.service_bus_topics if e.name.lower() != endpoint_name.lower()]
            endpoints.cosmos_db_sql_collections = [e for e in endpoints.cosmos_db_sql_collections if e.name.lower() != endpoint_name.lower()]
            endpoints.storage_containers = [e for e in endpoints.storage_containers if e.name.lower() != endpoint_name.lower()]
```

azext_iot/iothub/command_map.py
```diff
@@ -33,6 +39,43 @@ def load_iothub_commands(self, _):
         cmd_group.show_command("show", "get_digital_twin")
         cmd_group.command("update", "patch_digital_twin")
 
+    with self.command_group("iot hub messaging-endpoint", command_type=iothub_message_endpoint_ops) as cmd_group:
+        cmd_group.show_command("show", "message_endpoint_show")
+        cmd_group.command("list", "message_endpoint_list")
+        cmd_group.command(
+            "delete", "message_endpoint_delete", transform=EndpointUpdateResultTransform(self.cli_ctx)
```
c-ryan-k, line 46: Should we add confirmation to this command, since providing just the hub name (and no type/name parameters) will delete every endpoint on the hub?

Or perhaps we should require an `--all` switch (exclusive of type/name) that allows the deletion of all endpoints in one go?

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,281 @@
...
+        for endpoint in endpoints.cosmos_db_sql_collections:
+            if endpoint.name.lower() == endpoint_name.lower():
+                return endpoint
+        for endpoint in endpoints.storage_containers:
+            if endpoint.name.lower() == endpoint_name.lower():
+                return endpoint
+        # for endpoint_list in self.hub_resource.properties.routing.endpoints.values():
+        #     for endpoint in endpoint_list:
+        #         if endpoint.name.lower() == endpoint_name.lower():
+        #             return endpoint
```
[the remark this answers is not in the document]
vilit1: image
Flattening seems to make the code a bit confusing (especially with the multiple for ... in ...) so I will use your logic but keep it unflattened

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,150 @@ def load_iothub_help():
...
+        examples:
+          - name: Delete endpoint "E2" from "MyIotHub" IoT Hub.
+            text: >
+                az iot hub messaging-endpoint delete -n {iothub_name} --endpoint-name E2
+          - name: Delete all the endpoints of type "EventHub" from "MyIotHub" IoT Hub.
+            text: >
+                az iot hub messaging-endpoint delete -n {iothub_name} --endpoint-type eventhub
+          - name: Delete all the endpoints from "MyIotHub" IoT Hub.
+            text: >
+                az iot hub messaging-endpoint delete -n {iothub_name}
```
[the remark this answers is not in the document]
vilit1: Ahhh the models make 1 harder to implement. If it looks too wacky, I am fine with rewriting 1 to be 3

azext_iot/iothub/command_map.py
vilit1: will just add a blanket confirmation since users can delete multiple endpoints by providing just the type too.

azext_iot/iothub/providers/message_endpoint.py
vilit1: will do so - will keep the if statements in case someone provides a name and type

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,150 @@ def load_iothub_help():
...
+            text: >
+                az iot hub messaging-endpoint create storage-container -n {iothub_name} --en {endpoint_name} -c {connection_string} --cn {container_name}
+          - name: Create a Storage Container endpoint for your IoT Hub using system assigned identity with the given batch frequency, chunk size, and file name format. The endpoint must be specified.
+            text: >
+                az iot hub messaging-endpoint create storage-container -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri} --auth-type identityBased --entity-path {entity_path} --cn {container_name} -b {batch_frequency} -w {chunk_size} --ff {file_format}
+          - name: Create a Storage Container endpoint for your IoT Hub using user assigned identity with json encoding. The endpoint must be specified.
+            text: >
+                az iot hub messaging-endpoint create storage-container -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri} --auth-type identityBased --entity-path {entity_path} --cn {container_name} --encoding json
```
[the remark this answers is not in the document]
vilit1: ah it is defined as "--container-name", "--container" right now - I will change the collection_name to have collection instead of cn.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,150 @@ def load_iothub_help():
...
+
+    helps[
+        "iot hub messaging-endpoint create cosmosdb-collection"
+    ] = """
+        type: command
+        short-summary: Add a Cosmos DB Collection endpoint for your IoT Hub.
+        examples:
+          - name: Create a Cosmos DB Collection endpoint for your IoT Hub using a connection string.
+            text: >
+                az iot hub messaging-endpoint create cosmosdb-collection -n {iothub_name} --en {endpoint_name} -c {connection_string} --cn {collection} --db {database}
```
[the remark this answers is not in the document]
vilit1: I will change the param to `db` since it makes more sense than `dn` for database. Or should this be `database` to have more consistency?

.azure-devops/templates/trigger-tests.yml
```diff
@@ -121,6 +121,21 @@ jobs:
       pythonVersion: $(python)
       serviceConnection: $(AzureServiceConnection)
 
+- job: 'testHub_job_3'
+  condition: and(succeeded(), eq('${{ parameters.testHub }}', true))
+  displayName: 'Test IoT Hub - devices, messaging and modules'
```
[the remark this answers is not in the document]
vilit1: noted - we will most likely have this as a separate task (especially as we move more stuff)

azext_iot/iothub/_help.py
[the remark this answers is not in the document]
vilit1: Going to keep this open for now - so if we decide that the code isnt too wacky, I will add a test to check that 1 properly works (both scenarios: that the endpoint name and endpoint type is correct and the endpoint should be deleted; that endpoint name and endpoint type are mismatched so the endpoint should not be deleted)

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,281 @@
...
+        if not endpoint_subscription_id:
+            endpoint_subscription_id = self.hub_resource.additional_properties['subscriptionid']
+
+        endpoint_identity = None
+        if identity and authentication_type != AuthenticationType.IdentityBased.value:
+            raise ArgumentUsageError("In order to use an identity for authentication, you must select --auth-type as 'identityBased'")
+        elif identity and identity not in [IdentityType.none.value, SYSTEM_ASSIGNED_IDENTITY]:
+            endpoint_identity = ManagedIdentity(
+                user_assigned_identity=identity
+            )
```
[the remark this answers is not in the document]
vilit1: tbh the hub's error code is pretty straightforward here:

`(400139) Please check if User Assigned Identity '/subscriptions/blah/resourceGroups/blah/providers/Microsoft.ManagedIdentity/userAssignedIdentities/blah' is configured on IotHub. If you are deleting User Assigned Identity on IotHub, make sure it is not assigned to any endpoint of IotHub and then perform delete of User Assigned Identity. If you contact a support representative please include this correlation identifier: coderoni, timestamp: 2022-08-16 00:26:04Z, errorcode: IH400139.`

vilit1:
local Test runs:
if cosmos db is not created: 
image

if cosmos db is created before tests:
image

azext_iot/sdk/iothub/controlplane/models/_models_py3.py
```diff
@@ -0,0 +1,4499 @@
...
+        super().__init__(**kwargs)
+        self.total_device_count = None
+        self.enabled_device_count = None
+        self.disabled_device_count = None
+
+
+class RootCertificateProperties(msrest.serialization.Model):
+    """This property store root certificate related information.
+
+    @vilit - changed the type of the last_updated_rot_certificate_v2 to a rfc-1123
```
digimaun, line 2931: The change was made to last_updated_time_utc/lastUpdatedTimeUtc (a datetime attribute).

azext_iot/iothub/params.py
```diff
@@ -246,3 +247,133 @@ def load_iothub_arguments(self, _):
             options_list=["--content-type", "--ct"],
             help="MIME Type of file.",
         )
+
+    with self.argument_context("iot hub messaging-endpoint") as context:
```
digimaun: I think we should go with message-endpoint over messaging-endpoint for the command group. Then for routes we can do message-route.

azext_iot/sdk/iothub/controlplane/aio/__init__.py
```diff
@@ -0,0 +1,20 @@
+# coding=utf-8
```
digimaun: We can remove the entire aio directory if its not being used.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,251 @@ def load_iothub_help():
...
+                az iot hub message-endpoint create eventhub -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri} --auth-type identityBased --entity-path {entity_path}
+    """
+
+    helps[
+        "iot hub message-endpoint create servicebus-queue"
+    ] = """
+        type: command
+        short-summary: Add a Service Bus Queue endpoint for an IoT Hub.
+        examples:
+          - name: Create a Service Bus Queue endpoint for an IoT Hub using a connection string. The endpoint uri and entity path are omitted because they are in the connection string.
```
digimaun: Is there not a variation of the endpoint commands where the CLI can fetch connection string or other details for the user ? As it is for DT endpoints.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,251 @@ def load_iothub_help():
...
+            text: >
+                az iot hub message-endpoint create servicebus-queue -n {iothub_name} --en {endpoint_name} -c {connection_string}
+          - name: Create a Service Bus Queue endpoint for an IoT Hub using system assigned identity. The endpoint and entity path must be specified.
+            text: >
+                az iot hub message-endpoint create servicebus-queue -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri} --auth-type identityBased --entity-path {entity_path}
```
digimaun: This example is the same as the next.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,251 @@ def load_iothub_help():
...
+    ] = """
+        type: command
+        short-summary: Add a Service Bus Topic endpoint for an IoT Hub.
+        examples:
+          - name: Create a Service Bus Topic endpoint for an IoT Hub using a connection string. The endpoint uri and entity path are omitted because they are in the connection string.
+            text: >
+                az iot hub message-endpoint create servicebus-topic -n {iothub_name} --en {endpoint_name} -c {connection_string}
+          - name: Create a Service Bus Topic endpoint for an IoT Hub using system assigned identity. The endpoint and entity path must be specified.
+            text: >
+                az iot hub message-endpoint create servicebus-topic -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri} --auth-type identityBased --entity-path {entity_path}
```
digimaun: Example same as next.

azext_iot/iothub/_help.py
vilit1: right now no but I can add that in

azext_iot/sdk/iothub/controlplane/__init__.py
```diff
@@ -0,0 +1,23 @@
+# coding=utf-8
```
digimaun: The iot hub control plane client needs to be removed from this PR.

vilit1: I will keep this in for development until at least the new sdk with cosmos db functionality comes out.

azext_iot/iothub/_help.py
vilit1: Note that this will require a policy name that exists from users for eventhub/servicebus endpoints

vilit1:
image
 currently tests pass locally

vilit1:
local hub test pass
image

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,298 @@ def load_iothub_help():
...
+            text: |
+                az iot hub message-endpoint show -n {iothub_name} --endpoint-name {endpoint_name}
+    """
+
+    helps[
+        "iot hub message-endpoint delete"
+    ] = """
+        type: command
+        short-summary: Delete all or mentioned endpoint for an IoT Hub.
+        long-summary: Delete an endpoint for an IoT Hub. We recommend that you delete any routes to the endpoint, before deleting the endpoint.
```
avagraw, line 482: How about automating the route deletion for users. That is, if a user deletes an endpoint, we prompt the user to delete all the associated routes. Once they confirm, we delete the routes followed by the endpoint.

vilit1: will put that as a backlog, this pr is about getting the new routes and endpoints out (with cosmos db)

azext_iot/common/base_discovery.py
```diff
@@ -179,10 +179,17 @@ def find_resource(self, resource_name: str, rg: str = None):
         resource_list = self.get_resources()
 
         if resource_list:
-            target = next(
-                (resource for resource in resource_list if resource_name.lower() == resource.name.lower()),
-                None
-            )
+            target = None
+            if isinstance(resource_list[0], dict):
```
c-ryan-k: minor nit: we could use a placeholder value like:
`name = resource.get('name', '') if isinstance(resource_list[0], dict) else resource.name`

and then reference it as `name.lower()` later - just to avoid rewriting the iteration here, but functionally there's no difference.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+                "resourceGroup": endpoint_resource_group,
+                "authenticationType": authentication_type,
+                "endpointUri": endpoint_uri,
+                "entityPath": entity_path,
+                "identity": endpoint_identity
+            })
+        elif EndpointType.CosmosDBCollection.value == endpoint_type.lower():
+            if authentication_type != AuthenticationType.IdentityBased.value and not connection_string:
+                # try to get connection string
+                get_cosmos_db_cstring(
```
c-ryan-k: Does this need to be assigned to a value?
```suggestion
                connection_string = get_cosmos_db_cstring(
```

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+                "subscriptionId": endpoint_subscription_id,
+                "resourceGroup": endpoint_resource_group,
+                "authenticationType": authentication_type,
+                "endpointUri": endpoint_uri,
+                "identity": endpoint_identity
+            })
+        elif EndpointType.AzureStorageContainer.value == endpoint_type.lower():
+            if authentication_type != AuthenticationType.IdentityBased.value and not connection_string:
+                # try to get connection string
+                get_storage_cstring(
```
c-ryan-k: Same question here - if the calls are just for validation or if we want to use the returned value:
```suggestion
                connection_string = get_storage_cstring(
```

azext_iot/iothub/providers/base.py
```diff
@@ -29,3 +33,43 @@ def __init__(self, cmd, hub_name, rg, login=None, auth_type_dataplane=None):
...
+class IoTHubResourceProvider(object):
+    """Class to use for control plane operations."""
+    def __init__(
+        self,
+        cmd,
+        hub_name: str,
+        rg: Optional[str] = None,
+    ):
+        self.cmd = cmd
+        self.api_version = "2022-04-30-preview"
```
c-ryan-k: This is more of a conversation starter than a change request - but I'm curious how this affects our attachment to specific versions of the CLI core / hub SDK.

Assuming we're using the management SDK from CLI core, a few notes:
- We likely won't have SDKs generated for preview API versions, or if they are - they will have to be specified manually (like done here) and there may be some issues with importing models for these calls as they may be different from the non-preview Hub models
- We also need to keep track of which API-versions are in which versions of core, and we'll need to wait until CLI core is released with this new API-version before we ship the extension.
- It will also affect our min-CLI versions, as we're only as backwards-compatible as the API-versions available in whichever version of the management SDK that version of CLI core uses
- We *could* try to use the latest default version from CLI core, but that seems like a recipe for being broken by core updates without us knowing :)

vilit1: tbh it would be good if we can rely on the most current api version from the core. Otherwise, if no sdk is provided, it pulls from the api list in core and that is not the preview one that we want. The sdk is whatever is installed (I added a temporary azure-mgmt-iothub==2.3.0 to the setup.py)

I was going to make a pr in core to update the min api + force the azure-mgmt-iothub to be 2.3.0 and then in the extension's setup.py force the min cli core to be whatever cli core version that has the change in (and remove our dependency on azure-mgmt-iothub and the hardcoded api version since core will handle both)

There may be a few hiccups since you could technically install and uninstall packages and somehow get into a weird state where you have this extension with an old package. Would it make sense to have a check for "do you have the min package installed?" like how we do for track 1 vs 2?

as for your notes:
1. yes this will be an issue but as long as we say "for this feature and beyond we want min cli core version x that has sdk version y" we should be fine.
2. yes. It will be annoying but it will be done
3. yes, the new min cli version will be something very new so it may annoy some customers. But this is what happens when an api is not backwards compatible.
4. are you saying we should keep this hard coded api version? Id rather not because I would like to not have to track these here... Also will there be a cli core update touching iot (in a meaningful way) without us knowing?

azext_iot/iothub/_help.py
avagraw: Sounds good

azext_iot/common/base_discovery.py
vilit1: discussed - not proceeding with this since it is more complicated than this

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=6855&view=results

azext_iot/iothub/providers/base.py
vilit1: Ok, we can throw the dependency on core out the window since the api is now backwards compatible. From my testing, I have not found anything of issue (aside from being able to create and test routes with the cosmos db endpoint and then not list/update/show/delete them)

So it seems like all 4 points don't apply anymore...

.azure-devops/templates/setup-ci-machine.yml
```diff
@@ -1,6 +1,6 @@
 steps:
   - task: PowerShell@2
-    displayName: 'Upgrade pip, install dev_requirements and cryptography'
+    displayName: 'Upgrade pip, install dev_requirements and cryptography, force update azure-mgmt-iothub'
```
c-ryan-k: Just noticed that this is a `displayName` change but doesn't have a corresponding pipeline change - is the pipeline already doing this?

vilit1: oh this was me messing with some pipeline stuff - will remove

.azure-devops/templates/trigger-tests.yml
```diff
@@ -121,6 +121,21 @@ jobs:
       pythonVersion: $(python)
       serviceConnection: $(AzureServiceConnection)
 
+- job: 'testHub_job_3'
+  condition: and(succeeded(), eq('${{ parameters.testHub }}', true))
+  displayName: 'Test IoT Hub - routing endpoints'
```
digimaun: Can we stick to updated naming `routing endpoints` -> `message endpoints`?

digimaun:
There is a stacktrace when attempting to create a message-endpoint

image
image

azext_iot/common/base_discovery.py
```diff
@@ -178,9 +178,23 @@ def find_resource(self, resource_name: str, rg: str = None):
 
         resource_list = self.get_resources()
 
+        if isinstance(resource_list[0], dict):
```
digimaun: What is the reason for these changes, in the newer mgmt sdk the resource is a dict vs a model before?

vilit1: Good question - it may be a remain of when I tried to do modelless

vilit1:
> There is a stacktrace when attempting to create a message-endpoint
> 
> image image

temporary fix is to do `pip install -U azure-mgmt-iothub`

setup.py
```diff
@@ -46,6 +46,7 @@
 DEPENDENCIES = [
     "azure-core>=1.24.0,<2.0.0",
     "azure-mgmt-core>=1.3.0,<2.0.0",
+    "azure-mgmt-iothub==2.3.0",
```
digimaun: We should remove this since we will use the version from core.

azext_iot/common/base_discovery.py
```diff
@@ -178,9 +178,18 @@ def find_resource(self, resource_name: str, rg: str = None):
 
         resource_list = self.get_resources()
 
+        for resource in resource_list:
+            if resource_name.lower() == resource.name.lower():
+                target = resource
+
         if resource_list:
+            is_dict = isinstance(resource_list[0], dict)
```
digimaun: Still supporting both model & modeless based sdk

vilit1: I swear that I thought I removed it

azext_iot/iothub/providers/base.py
```diff
@@ -29,3 +33,43 @@ def __init__(self, cmd, hub_name, rg, login=None, auth_type_dataplane=None):
...
+        # Need to get the direct resource
+        self.hub_resource = self.get_iot_hub_resource(hub_name, rg)
+
+    def get_client(self):
+        from azure.cli.core.commands.client_factory import get_mgmt_service_client
+        from azure.cli.core.profiles import ResourceType
+        client = get_mgmt_service_client(
+            self.cmd.cli_ctx,
+            ResourceType.MGMT_IOTHUB,
+            api_version=self.api_version
```
digimaun: Unclear if the optimal pattern is specifying the API version of the mgmt client here. If you don't specify one, you get the latest version correct?

digimaun: We resolved this in pair-programming /w @c-ryan-k

azext_iot/iothub/common.py
```diff
@@ -0,0 +1,68 @@
...
+    user_assigned = "UserAssigned"
+    system_assigned_user_assigned = "SystemAssigned, UserAssigned"
+    none = "None"
+
+
+class RouteSourceType(Enum):
+    """
+    Type of the route source.
+    """
+    # Invalid = 'invalid'
```
vilit1: this was in the azure cli core endpoint types but this source type just errors out. Can we get rid of it?

azext_iot/iothub/params.py
```diff
@@ -246,3 +247,210 @@ def load_iothub_arguments(self, _):
...
+        )
+        context.argument(
+            "endpoint_type",
+            arg_type=get_enum_type(EndpointType),
+            options_list=["--endpoint-type", "--type", "-t"],
+            help="Type of the Routing Endpoint."
+        )
+
+    with self.argument_context("iot hub message-endpoint create") as context:
+        context.argument(
```
vilit1, line 272: we could infer auth_type from identity (as done for the dt UAI work)

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+        endpoint_identity = None
+        if identity and authentication_type != AuthenticationType.IdentityBased.value:
+            raise ArgumentUsageError(
+                "In order to use an identity for authentication, you must select --auth-type as 'identityBased'."
+            )
+        elif identity and identity not in [IdentityType.none.value, SYSTEM_ASSIGNED_IDENTITY]:
+            endpoint_identity = ManagedIdentity(
+                user_assigned_identity=identity
+            )
+        elif authentication_type != AuthenticationType.IdentityBased.value and not connection_string:
```
vilit1: missing some logic on ensuring that values are missing before raising the error

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,298 @@ def load_iothub_help():
...
+
+    helps[
+        "iot hub message-endpoint create"
+    ] = """
+        type: group
+        short-summary: Add an endpoint to an IoT Hub.
+    """
+
+    helps[
+        "iot hub message-endpoint create cosmosdb-collection"
```
vilit1: may need to add a check using ensure_iothub_sdk_min_version('2.3.0') to hide the help for older azure cli cores

azext_iot/iothub/providers/base.py
```diff
@@ -29,3 +31,17 @@ def __init__(self, cmd, hub_name, rg, login=None, auth_type_dataplane=None):
 
     def get_sdk(self, sdk_type):
         return self.resolver.get_sdk(sdk_type)
+
+
+class IoTHubResourceProvider(object):
```
vilit1: since this no longer does weird stuff with getting the client, we could just combine this with the normal iothub provider

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+                    error_msg + " or endpoint namespace '--endpoint-namespace', endpoint "
+                    "entity path '--entity-path', and policy name '--policy-name'."
+                )
+            else:
+                raise ArgumentUsageError(
+                    error_msg + " or endpoint account '--endpoint-account'."
+                )
+
+        endpoints = self.hub_resource.properties.routing.endpoints
+        if EndpointType.EventHub.value == endpoint_type.lower():
```
vilit1, line 106: I hate the length of this command, is there a better way to refactor it or use helper functions

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+            endpoints.service_bus_queues,
+            endpoints.service_bus_topics,
+            endpoints.cosmos_db_sql_collections,
+            endpoints.storage_containers
+        ]
+        for endpoint_list in endpoint_lists:
+            for endpoint in endpoint_list:
+                if endpoint.name.lower() == endpoint_name.lower():
+                    return endpoint
+        raise ResourceNotFoundError("No endpoint found.")
```
vilit1: Error message should be "Endpoint {blah} not found in IoT Hub {blah}."

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+            elif EndpointType.ServiceBusQueue.value == endpoint_type:
+                endpoints.service_bus_queues = []
+            elif EndpointType.ServiceBusTopic.value == endpoint_type:
+                endpoints.service_bus_topics = []
+            elif EndpointType.CosmosDBCollection.value == endpoint_type:
+                endpoints.cosmos_db_sql_collections = []
+            elif EndpointType.AzureStorageContainer.value == endpoint_type:
+                endpoints.storage_containers = []
+
+        if not endpoint_type and not endpoint_name:
```
vilit1: could be else

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+
+
+logger = get_logger(__name__)
+
+
+class MessageEndpoint(IoTHubResourceProvider):
+    def __init__(
+        self,
+        cmd,
+        hub_name: Optional[str] = None,
```
vilit1: hub_name should not be optional

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+    ):
+        super(MessageRoute, self).__init__(cmd, hub_name, rg)
+
+    def create(
+        self,
+        route_name: str,
+        source_type: str,
+        endpoint_name: str,
+        enabled: bool = True,
+        condition: Optional[str] = None,
```
vilit1: cant we just put str = 'true' here?

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+    ):
+        route = self.show(route_name=route_name)
+        route.source = route.source if source_type is None else source_type
+        route.endpoint_names = route.endpoint_names if endpoint_name is None else endpoint_name.split()
+        route.condition = route.condition if condition is None else condition
+        route.is_enabled = route.is_enabled if enabled is None else enabled
+
+        return self.client.iot_hub_resource.begin_create_or_update(
+            resource_group_name=self.hub_resource.additional_properties['resourcegroup'],
```
vilit1, line 67: need consistency in ' vs "

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+            iot_hub_description=self.hub_resource,
+            if_match=self.hub_resource.etag
+        )
+
+    def show(self, route_name: str):
+        routes = self.hub_resource.properties.routing.routes
+        for route in routes:
+            if route.name.lower() == route_name.lower():
+                return route
+        raise ResourceNotFoundError("No route found.")
```
vilit1, line 78: "Route {blah} not found in IoT Hub {blah}."

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
```diff
@@ -0,0 +1,906 @@
...
+    endpoint_list = cli.invoke(
+        "iot hub message-endpoint list -n {} -g {} -t {}".format(
+            iot_hub, iot_rg, "storage-container"
+        )
+    ).as_json()
+
+    assert endpoint_list == []
+
+
+def test_iot_cosmos_endpoint_lifecycle(provisioned_cosmosdb_with_identity_module):
```
vilit1, line 614: add skipif to make sure we do not test it on an azure cli core version that does not support this

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+            "body": body,
+            "appProperties": app_properties,
+            "systemProperties": system_properties
+        }
+
+        if route_name:
+            route = self.show(route_name)
+            test_route_input = {
+                "message": route_message,
+                "twin": None,
```
vilit1, line 125: Need to add a backlog item to add twin to route test

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+                "message": route_message,
+                "twin": None
+            }
+            return self.client.iot_hub_resource.test_all_routes(
+                iot_hub_name=self.hub_resource.name,
+                resource_group_name=self.hub_resource.additional_properties['resourcegroup'],
+                input=test_all_routes_input
+            )
+
+        # for all types, need to test all types one by one
```
vilit1, line 146: note that this is a major change from how `az iot hub route test` works

Do we like this functionality or no?

azext_iot/iothub/providers/message_route.py
```diff
@@ -0,0 +1,184 @@
...
+            if len(result) == 1 and result[0].properties.name == "$fallback":
+                fallback = result
+            else:
+                routes.extend(result)
+
+        if len(routes) == 0 and fallback:
+            routes = fallback
+        return {"routes": routes}
+
+    def show_fallback(self):
```
vilit1, line 171: fallback route is a new functionality that should be included in history

azext_iot/tests/helpers.py
```diff
@@ -127,6 +128,233 @@ def create_storage_account(
...
+            return cs_object["connectionString"]
+
+
+def create_managed_identity(cmd, name: str, rg: str):
+    return cmd(
+        f"identity create -n {name} -g {rg}"
+    ).get_output_in_json()
+
+
+def get_role_assignments(cmd, scope, role):
```
vilit1: we could add an optional assignee

azext_iot/tests/helpers.py
```diff
@@ -127,6 +128,233 @@ def create_storage_account(
...
+    return cmd(
+        'role assignment list --scope "{}" --role "{}"'.format(
+            scope, role
+        )
+    ).get_output_in_json()
+
+
+def assign_rbac_role(cmd, assignee: str, scope: str, role: str, max_tries: int = 10):
```
vilit1: This and get_role_assignments end up everywhere for tests. We should condense them into one function (prob a future task). Also another future task is to combine the other service creations in a neat way since they are shared btw dt, adu)

azext_iot/tests/helpers.py
```diff
@@ -127,6 +128,233 @@ def create_storage_account(
...
+                assignee, role, scope
+            )
+        )
+        sleep(10)
+        tries += 1
+
+
+def assign_cosmos_db_role(cmd, principal_id: str, role: str, cosmos_db_account: str, rg: str):
+    cmd(
+        "cosmosdb sql role assignment create -a {} -g {} --scope '/' -n '{}' -p {}".format(
```
vilit1: This is a blanket (everything) scope. Do we want to specify a more specific scope?

azext_iot/tests/iothub/__init__.py
```diff
@@ -48,6 +58,21 @@
 MAX_RBAC_ASSIGNMENT_TRIES = settings.env.azext_iot_rbac_max_tries or 10
 ROLE_ASSIGNMENT_REFRESH_TIME = 120
 
+# Endpoints
+EP_RG = settings.env.azext_iot_ep_rg or ENTITY_RG
```
vilit1: we should think of a neater way to create many constant variables. Prob a future task.

azext_iot/tests/iothub/__init__.py
```diff
@@ -210,24 +236,335 @@ def _create_storage_account(self):
...
 
         elif not settings.env.azext_iot_teststoragecontainer:
             self.cmd(
                 "storage container delete -n {} --connection-string '{}'".format(
-                    self.storage_account_name, self.storage_cstring
+                    STORAGE_ACCOUNT, self.storage_cstring
                 ),
             )
 
+    def _create_cosmos_db(self):
```
vilit1: I may have fixturized all of this so this code is prob not used.

Delete if not used

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
+# coding=utf-8
```
vilit1, line 1: we need a task to combine fixtures in a nice way

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+settings = DynamoSettings(req_env_set=REQUIRED_TEST_ENV_VARS)
+RG = settings.env.azext_iot_testrg
+
+
+def generate_hub_id() -> str:
+    return f"test-hub-{generate_generic_id()}"[:40]
+
+
+def generate_hub_depenency_id() -> str:
+    return f"testhubdep{generate_generic_id()}"[:24]
```
vilit1: are we ok with this naming for everything related to the hub? It will help with searching and mass deleting if needed

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+
+    hub_marker = get_closest_marker(request)
+    desired_location = None
+    desired_tags = None
+    desired_sys_identity = False
+    desired_user_identity = False
+
+    if hub_marker:
+        desired_location = hub_marker.kwargs.get("location")
+        desired_tags = hub_marker.kwargs.get("desired_tags")
```
vilit1, line 77: Note that this gets rid of the original test tag functionality and I do not have code to add more test tags

I will need to think about how to add them back in a neat fashion

azext_iot/iothub/_help.py
c-ryan-k: @digimaun just as an FYI - this throws a tiny wrench in the gears of hiding commands based on api-versions, we may also need to configure the help in a similar manner to make the command table / help linter happy for older CLI versions

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+def provisioned_only_iot_hub_module(request) -> Optional[dict]:
+    result = _iot_hub_provisioner(request)
+    yield result
+    if result:
+        _iot_hub_removal(result["name"])
+
+
+def _iot_hub_provisioner(request, provisioned_user_identity=None):
+    name = generate_hub_id()
+    base_create_command = f"iot hub create -n {name} -g {RG} --sku S1"
```
vilit1, line 67: This fixture will always create a new hub, eventhub, etc. Do we want that or no? 

We may need to add in code to allow a hub to be specified in the future.

azext_iot/iothub/_help.py
c-ryan-k: @vilit1 in case you missed this

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+        base_create_command += " --mi-system-assigned"
+    if desired_user_identity and provisioned_user_identity:
+        user_identity_id = provisioned_user_identity["id"]
+        base_create_command += f" --mi-user-assigned {user_identity_id}"
+    if desired_tags:
+        base_create_command = base_create_command + f" --tags {desired_tags}"
+    if desired_location:
+        base_create_command = base_create_command + f" -l {desired_location}"
+
+    return cli.invoke(base_create_command).as_json()
```
vilit1, line 91: Note that this returns the entire hub object and not the things needed for the test. This will help with generalizing the conftest but at the same time does have a lot of wasted data. 
Should we keep returning everything or just return what we need?

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+
+
+def _iot_hub_removal(name):
+    delete_result = cli.invoke(f"iot hub delete -n {name} -g {RG}")
+    if not delete_result.success():
+        logger.error(f"Failed to delete iot hub resource {name}.")
+
+
+@pytest.fixture(scope="module")
+def provisioned_user_identity_module() -> Optional[dict]:
```
vilit1, line 102: I made everything module but we may need fixtures with smaller scopes.

also add comments to denote sections/specific services

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+    ).as_json()
+
+
+def _user_identity_removal(name):
+    delete_result = cli.invoke(f"identity delete -n {name} -g {RG}")
+    if not delete_result.success():
+        logger.error(f"Failed to delete user identity resource {name}.")
+
+
+def _assign_rbac_role(assignee: str, scope: str, role: str, max_tries: int = 10):
```
vilit1, line 128: this function is everywhere!

azext_iot/iothub/_help.py
c-ryan-k: @vilit1 ^

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+                assignee, role, scope
+            )
+        )
+        sleep(10)
+        tries += 1
+
+
+@pytest.fixture(scope="module")
+def provisioned_storage_with_identity_module(request, provisioned_iot_hub_module, provisioned_storage_module):
+    role = "Storage Blob Data Contributor"
```
vilit1, line 148: prob move this to a constants for tests file (and similar roles)

azext_iot/tests/iothub/message_endpoint/conftest.py
```diff
@@ -0,0 +1,464 @@
...
+    scope = provisioned_storage_module["storage"]["id"]
+    hub_principal_id = provisioned_iot_hub_module["identity"]["principalId"]
+    user_identities = list(provisioned_iot_hub_module["identity"]["userAssignedIdentities"].values())
+    user_id = user_identities[0]["principalId"]
+    _assign_rbac_role(assignee=hub_principal_id, scope=scope, role=role)
+    _assign_rbac_role(assignee=user_id, scope=scope, role=role)
+    yield provisioned_iot_hub_module, provisioned_storage_module
```
vilit1, line 155: should we do tuples or dicts? dicts are easier and nicer but tuples are smaller

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,298 @@ def load_iothub_help():
...
+        type: command
+        short-summary: Delete all routes or a mentioned route in an IoT Hub.
+        examples:
+          - name: Delete a route from an IoT Hub.
+            text: >
+                az iot hub message-route delete -n {iothub_name} --route-name {route_name}
+          - name: Delete all routes  of source type "DeviceMessages" from an IoT Hub.
+            text: >
+                az iot hub message-route delete -n {iothub_name} --source DeviceMessages
+          - name: All route from an IoT Hub.
```
c-ryan-k:
```suggestion
          - name: Delete all routes from an IoT Hub.
```

azext_iot/iothub/common.py
c-ryan-k: I wonder if this value was just used to map from all possible return types from the service/SDK - I don't see why we'd need this value from the client perspective, no reason to create/test all routes with `invalid` types.

https://github.com/Azure/azure-cli/blob/99f07bc11069b66ed7e4dfbd01b86868371ee9ac/src/azure-cli/azure/cli/command_modules/iot/tests/latest/test_iot_commands.py#L365

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,360 @@
...
+                connection_string = get_servicebus_topic_cstring(
+                    cmd=self.cmd,
+                    namespace_name=endpoint_account_name,
+                    topic_name=entity_path,
+                    policy_name=endpoint_policy_name,
+                    rg=endpoint_resource_group,
+                    sub=endpoint_subscription_id
+                )
+            endpoints.service_bus_topics.append({
+                "connection_string": connection_string,
```
c-ryan-k: Just to clarify, service bus queues use `connectionString` - does topic use `connection_string` or does this need changed?

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Just off the top of my head, a few things to simplify:
- repetitive checks of `if authentication_type != AuthenticationType.IdentityBased.value and not connection_string:` could be resolved to a single variable like `should_create_cstring` (that name is probably bad but anyway)
- repetitive `append` with custom objects for a few specific endpoint types (all using the same attribute), could possibly create the endpoint object once and use it in a few places:
  ````
  new_endpoint = {
        "connectionString": connection_string,
        "name": endpoint_name,
        "subscriptionId": endpoint_subscription_id,
        "resourceGroup": endpoint_resource_group,
        "authenticationType": authentication_type,
        "endpointUri": endpoint_uri,
        "entityPath": entity_path,
        "identity": endpoint_identity
   }
  ...
  endpoints.event_hubs.append(new_endpoint)
  ...
  endpoints.service_bus_queues.append(new_endpoint)
  ...etc

azext_iot/iothub/command_map.py
```diff
@@ -33,6 +48,73 @@ def load_iothub_commands(self, _):
...
+            "servicebus-queue",
+            "message_endpoint_create_service_bus_queue",
+            transform=EndpointUpdateResultTransform(self.cli_ctx)
+        )
+        cmd_group.command(
+            "servicebus-topic",
+            "message_endpoint_create_service_bus_topic",
+            transform=EndpointUpdateResultTransform(self.cli_ctx)
+        )
+        cmd_group.command(
```
vilit1, line 82: double check if needs to be marked as preview

azext_iot/iothub/command_map.py
```diff
@@ -33,6 +48,73 @@ def load_iothub_commands(self, _):
...
+            'delete',
+            'message_route_delete',
+            transform=RouteUpdateResultTransform(self.cli_ctx),
+            confirmation=True
+        )
+        cmd_group.command(
+            'update', 'message_route_update', transform=RouteUpdateResultTransform(self.cli_ctx)
+        )
+        cmd_group.command('test', 'message_route_test')
```
vilit1, line 112: since the functionality here is different from normal test, this may need to be marked as preview

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
```diff
@@ -0,0 +1,906 @@
...
+    queue_instance = servicebus_obj["queue"]["name"]
+    topic_instance = servicebus_obj["topic"]["name"]
+    endpoint_uri = "sb:" + servicebus_obj["namespace"]["serviceBusEndpoint"].split(":")[1]
+    queue_cs = servicebus_obj["queueConnectionString"]
+    topic_cs = servicebus_obj["topicConnectionString"]
+    # create 6 names, 2 types of service bus endpoints * 3 auth types
+    endpoint_names = generate_ep_names(6)
+
+    # create topic endpoint first - connection string
+    cli.invoke(
```
vilit1, line 186: we should have checks for invalid commands

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: I think that works...

My understanding of the desired logic is:
- if name - delete a single endpoint with that name
- if type - delete all endpoints of that type
- if no name and no type - delete all endpoints

So the final else would be:
`not endpoint_name and not endpoint_type`, which I think is the desired outcome.

I'm a bit curious about the `endpoint_name and not endpoint_type` scenario - technically this loops through each type of endpoint removing all with the same name, though I believe we've clarified that endpoint names are unique.

I think the end result is the same, but we may be doing some extra work there.

azext_iot/iothub/providers/message_route.py
c-ryan-k: May need to test this, but if `None` is passed into `create()` - I think you still need to check `if not condition` or `if condition is None` later - because the default parameter only gets used if there is no value passed to the function, `None` is still considered a value iirc

azext_iot/iothub/providers/message_route.py
c-ryan-k: our preferred `black` formatter can help with that - although be warned it will format an entire document as it doesn't support "format selection"

azext_iot/iothub/providers/message_endpoint.py
vilit1: good q - both prob work

azext_iot/iothub/common.py
vilit1: ok will leave it in just incase some dev wants it but will add a class function to return a list without the invalid:
```
    @classmethod
    def list_valid_types(cls):
        return list(filter(lambda d: d != RouteSourceType.Invalid.value, map(lambda c: c.value, cls)))```

File: azext_iot/iothub/providers/message_endpoint.py (hunk 34 above)
Comment (vilit1) [reply azext_iot/iothub/providers/message_endpoint.py]: I guess we could just add an if found bool to short circuit but then we would have to choose which types go first (so in this case deleting eventhubs will be faster than cosmos db since to get to cosmos, since the if statement logic goes through each type one by one)

or we can leave as is and say everyone suffers

(tbh this prob wont even make a huge difference timing wise since the max number of endpoints in a normal hub is 10)

File: azext_iot/iothub/providers/message_route.py (hunk 36 above)
Comment (vilit1) [reply azext_iot/iothub/providers/message_route.py]: yeah this is the check:
`"condition": ('true' if condition is None else condition),`

File: azext_iot/iothub/providers/base.py (hunk 31 above)
Comment (vilit1) [reply azext_iot/iothub/providers/base.py]: Note that if we do, we should make sure that the discovery doesnt get fetch target for message route/endpoints to reduce run time

File: azext_iot/tests/helpers.py (hunk 43 above)
Comment (vilit1) [reply azext_iot/tests/helpers.py]: just killed the function

File: azext_iot/tests/helpers.py (hunk 44 above)
Comment (vilit1) [reply azext_iot/tests/helpers.py]: killed the function

File: azext_iot/tests/iothub/__init__.py (hunk 46 above)
Comment (vilit1) [reply azext_iot/tests/iothub/__init__.py]: killed these

File: azext_iot/tests/iothub/message_endpoint/conftest.py (hunk 48 above)
Comment (vilit1) [reply azext_iot/tests/iothub/message_endpoint/conftest.py:1]: same with get and assign role functions

File: azext_iot/iothub/providers/message_endpoint.py [hunk 62]
```diff
@@ -0,0 +1,376 @@
...
+                "In order to use an identity for authentication, you must select --auth-type as 'identityBased'."
+            )
+        elif identity and identity not in [IdentityType.none.value, SYSTEM_ASSIGNED_IDENTITY]:
+            endpoint_identity = ManagedIdentity(
+                user_assigned_identity=identity
+            )
+        elif authentication_type != AuthenticationType.IdentityBased.value and not connection_string:
+            # check for connection string args
+            error_msg = "Please provide a connection string '--connection-string/-c'"
+            if not(
```
c-ryan-k: Linter fix:
```suggestion
            if not (
```

azext_iot/tests/helpers.py
```diff
@@ -8,6 +8,7 @@
 import os
 
 from inspect import getsourcefile
+from time import sleep
```
c-ryan-k: Linter fix:
```suggestion
```

azext_iot/iothub/_help.py
c-ryan-k: Note - we discussed offline, this might be an issue just for our pipeline (assuming help commands don't show for commands that aren't loaded) and not actual customer usage - so we may be able to pop this in our local `linter_exclusions` to make our pipeline happy and be ok.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,320 @@ def load_iothub_help():
...
+          - name: Create an Event Hub endpoint for an IoT Hub using system assigned identity. The endpoint and entity path must be specified.
+            text: >
+                az iot hub message-endpoint create eventhub -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri}
+                --auth-type identityBased --entity-path {entity_path}
+          - name: Create an Event Hub endpoint for an IoT Hub using user assigned identity. The endpoint and entity path must be specified.
+            text: >
+                az iot hub message-endpoint create eventhub -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri}
+                --auth-type identityBased --entity-path {entity_path} --identity {user_identity_resource_id}
+    """
+
```
xujing-ms, line 379: @c-ryan-k @digimaun do we not need a --endpoint-type for each type of endpoint anymore? now it is just create cosmosdb-collection, create eventhub?

and also, do we still have the option to list a separate subscription id and resource group for each endpoint creation?

digimaun: Yea that's right, this new command group is conforming to our digital twins endpoint style where each endpoint has a dedicated create command rather than trying to support all endpoints in one. 

For the second one, each create should support `--endpoint-subscription-id` and `--endpoint-resource-group` (as well as `--connection-string`)

azext_iot/iothub/commands_message_endpoint.py
```diff
@@ -0,0 +1,227 @@
...
+    connection_string: Optional[str] = None,
+    authentication_type: Optional[str] = None,
+    endpoint_uri: Optional[str] = None,
+    primary_key: Optional[str] = None,
+    secondary_key: Optional[str] = None,
+    partition_key_name: Optional[str] = None,
+    partition_key_template: Optional[str] = None,
+    identity: Optional[str] = None,
+    resource_group_name: Optional[str] = None,
+):
```
xujing-ms, line 124: @digimaun  @c-ryan-k endpoint_accountname is quite different from to database/collection names? and it should not be optional as per swagger copied below. it has a lot more required fields than the other endpoints

iot hub swagger:
      "required": [
        "name",  <----- account name
        "endpointUri",  <---- also are we concatenating the strings together to make the endpoint uri or should we make it also required?
        "databaseName",
        "collectionName"
      ]

digimaun: In the change set, the reason endpointUri is not required is because it can be explicitly provided, or derived from a connection string.

For "name" in the swagger, I believe that is referring to the endpoint name rather than the account name https://github.com/Azure/azure-rest-api-specs/blob/main/specification/iothub/resource-manager/Microsoft.Devices/preview/2022-04-30-preview/iothub.json#L3218

In the command the optional `--endpoint-account` is used to obtain a connection string for key based integration if `--connection-string` has not been provided. 

We should make sure help / errors are clear about the different flows.

azext_iot/iothub/_help.py
vilit1: yup to both - 

sub id and rg is now optional, defaulting to dt sub and rg if not provided
image

azext_iot/tests/helpers.py
vilit1: blanket scope will make do for testing purposes

azext_iot/iothub/providers/message_endpoint.py
vilit1: did some refactoring

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: future task

azext_iot/iothub/commands_message_endpoint.py
vilit1: fixed error handling, the error now gives:
image

c-ryan-k: @vilit1 just curious- if they provide both `--endpoint-account` and `-c` - do they get a `MutuallyExclusiveArgumentError` or do we just use the CS and swallow the endpoint account parameter?

azext_iot/iothub/providers/message_endpoint.py
vilit1: Going to leave as is and let everyone suffer since the max number of endpoints is 10

azext_iot/iothub/_help.py
```diff
@@ -310,6 +310,323 @@ def load_iothub_help():
...
+                --endpoint-uri {endpoint_uri} --container {container} --db {database}
+          - name: Create a Cosmos DB Container endpoint for an IoT Hub using system assigned identity and a partition key name.
+                  The partition key template will be the default.
+            text: >
+                az iot hub message-endpoint create cosmosdb-container -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri}
+                --container {container} --db {database} --auth-type identityBased --pkn {partition_key_name}
+          - name: Create a Cosmos DB Container endpoint for an IoT Hub using user assigned identity, partition key name, and partition key template.
+            text: >
+                az iot hub message-endpoint create cosmosdb-container -n {iothub_name} --en {endpoint_name} --endpoint-uri {endpoint_uri}
+                --container {container} --db {database} --auth-type identityBased --pkn {partition_key_name} --pkt {partition_key_template}
```
digimaun: We should eliminate/deprecate `--auth-type` in favor of determining auth type implicitly based on provided identity params (or lack of them) --- similar to ADT UAI changes.

vilit1: I did that and did not clean up help
cleaning help rn

azext_iot/iothub/commands_message_endpoint.py
vilit1: we use the account name from endpoint account and the keys from the connection string

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
vilit1: oh but this is the embedded cli so the error messages will litter the logs - would we still want these

vilit1:
test run:
The job with messaging endpoints isnt failing but the other job is timing out so far: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7128&view=results

going to look into this more

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7138&view=results

azext_iot/tests/iothub/message_endpoint/conftest.py
c-ryan-k: I'm fine tagging things but maybe we make it a tiny bit more readable?
```suggestion
    return f"test_hub_dep_{generate_generic_id()}"[:24]
```

or possibly even link it to the hub it's being generated for:
```suggestion
    return f"{hub_id}_dependency_{generate_generic_id()}"[:24]
```

azext_iot/tests/iothub/message_endpoint/conftest.py
c-ryan-k: I guess it depends on the usage. We could also protect this with a preliminary call to see if a hub that we're looking for exists before creating it, or we overload this with a `force` param and adjust the logic to do the same.

azext_iot/tests/iothub/message_endpoint/conftest.py
c-ryan-k: I don't think returning extra data is going to harm our tests, unless you find it's causing extreme slowdowns or timeouts etc -assuming this is only called once per created hub, you can then parse out which properties you want to assert / test with after it returns.

azext_iot/tests/iothub/message_endpoint/conftest.py
c-ryan-k: If it's only returning two values in a test, I think a tuple is fine for now. It helps us not have to remember named properties, and we can just index the properties (again, since there's only two, `[0]` and `[1]` are fine)

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: There are two issues with this
1. not every service accepts _ (looking at event hubs) and some don't even accept non alphanumeric characters (looking at storage) in the name
2. I am also trying to keep the length shorter cause some services have shorter name limits than the iot hub

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Seems a bit cleaner (duplicate logic removed) - unsure why some of these branches of the logic create a `connection_string` value thought that's not used... are these supposed to be part of the `new_endpoint` object like `new_endpoint["connectionString"]`?

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: my solution (for figuring out what iot hub is associated) here was tagging - but I sorta removed that for this test and was planning on figuring out a better way in a future pr

vilit1: offline- decided aziotclitest as a prefix is better

azext_iot/iothub/providers/message_route.py
c-ryan-k: I'm not familiar with previous `route test` functionality.

Could you explain the major differences between the two versions (or at least the previous one)?
This one seems like it:

- Tests a route by name, if provided
- Tests all routes by source_type, if provided
- Tests all source types through all routes if neither are provided, with some logic to dance around the fallback route

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: Have not thought about if there is a hub already - will prob prefer rerolling the name over reusing the hub for now

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: alright will leave as is so it is more usable for other (future) tests

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: Will add pytest options in a future pr since this will need a restructuring of existing pytest variables and more documentation changes

vilit1: there is a work item already :D https://msazure.visualstudio.com/One/_sprints/taskboard/IoT-Platform-UPX/One/Custom/IoT/2212?workitem=15274556

azext_iot/iothub/providers/message_route.py
vilit1: The biggest change is in the test all source types cause now it actually tests all source types rather than pretend to (and only do device messages - so it never actually was able to test all routes)

image

old route test - 
image
only this one passes b/c it is enabled, a true condition, and **is of type DeviceMessages**


new route test - 
image

this grabs things that aren't of type devicemessages

vilit1:
better test run https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7145&view=results

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
vilit1: will omit for now

azext_iot/tests/iothub/message_endpoint/conftest.py
vilit1: added comments. Will keep as module as its prob the smallest scope that makes sense

azext_iot/iothub/providers/message_endpoint.py
vilit1: fixed

azext_iot/iothub/providers/message_route.py
vilit1: added to history

azext_iot/iothub/providers/message_route.py
vilit1: added to history

vilit1:
nightly passing:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7152&view=results
the min cli version does not support the cosmos db yet so that test is skipped

azext_iot/iothub/common.py
```diff
@@ -11,6 +11,11 @@
 from enum import Enum
 
 
+SYSTEM_ASSIGNED_IDENTITY = '[system]'
+
+# Message Endpoint Messages
+INVALID_CLI_CORE_FOR_COSMOS = "This version of the azure cli core does not support Cosmos Db Endpoints for IoT Hub."
```
c-ryan-k, line 18: Totally not something to solve in this PR, but just a thought that we have enough messages / constants spread across various `shared.py` and `common.py` - I'm wondering if it would be beneficial to separate (specifically, error message strings or other output formatting) into a `strings.py` module at some point.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,351 @@
...
+        database_name: Optional[str] = None,
+        primary_key: Optional[str] = None,
+        secondary_key: Optional[str] = None,
+        partition_key_name: Optional[str] = None,
+        partition_key_template: Optional[str] = None,
+        identity: Optional[str] = None
+    ):
+        resource_group_name = self.hub_resource.additional_properties["resourcegroup"]
+        if not endpoint_resource_group:
+            endpoint_resource_group = resource_group_name
```
c-ryan-k: tiny nit - i don't think `resource_group_name` is used outside this scope:
```suggestion
        if not endpoint_resource_group:
            endpoint_resource_group = self.hub_resource.additional_properties["resourcegroup"]
```

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,351 @@
...
+                raise ArgumentUsageError(
+                    error_msg + " or endpoint namespace '--endpoint-namespace', endpoint "
+                    "entity path '--entity-path', and policy name '--policy-name'."
+                )
+            elif not endpoint_account_name and endpoint_type.lower() in [
+                EndpointType.AzureStorageContainer.value, EndpointType.CosmosDBContainer.value
+            ]:
+                raise ArgumentUsageError(
+                    error_msg + " or endpoint account '--endpoint-account'."
+                )
```
c-ryan-k, line 92: Not to mess with the logic too much here, but assuming if you have an identity, you don't need a connection string, but here we have no identity, AND no connection string:
Is it possible to add something about `Please provide a connection string OR a managed identity, otherwise...` or `No managed identity or connection string provided...` ? Or am I maybe overthinking it?

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,351 @@
...
+                secondary_key = primary_key
+            if secondary_key and not primary_key:
+                primary_key = secondary_key
+            if not new_endpoint["endpointUri"]:
+                raise RequiredArgumentMissingError(
+                    "Endpoint uri via --endpoint-uri or connection string via --connection-string is required."
+                )
+            if partition_key_name and not partition_key_template:
+                partition_key_template = '{deviceid}-{YYYY}-{MM}'
+            del new_endpoint["connectionString"]
```
c-ryan-k, line 176: This value is deleted before it's used.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -0,0 +1,351 @@
...
+                    sub=endpoint_subscription_id
+                )
+            if not container_name:
+                raise RequiredArgumentMissingError("Container name is required.")
+            new_endpoint.update({
+                "containerName": container_name,
+                "encoding": encoding.lower() if encoding else EncodingFormat.AVRO.value,
+                "fileNameFormat": file_name_format,
+                "batchFrequencyInSeconds": batch_frequency,
+                "maxChunkSizeInBytes": (chunk_size_window * 1048576),
```
c-ryan-k: nit: Maybe we could pull this into a constant of `BYTES_PER_MEGABYTE`

azext_iot/tests/settings.py
```diff
@@ -15,7 +15,19 @@
...
+    "azext_iot_eventhub_instance",
+    "azext_iot_eventhub_policy",
+    "azext_iot_servicebus_namespace",
+    "azext_iot_servicebus_queue",
+    "azext_iot_servicebus_topic",
+    "azext_iot_servicebus_policy",
+    "azext_iot_cosmos_namespace",
+    "azext_iot_cosmos_database",
+    "azext_iot_cosmos_collection",
+    "azext_iot_user_identity"
```
c-ryan-k: Do these need added to [`set-testenv-sentinel.yml`](https://github.com/Azure/azure-iot-cli-extension/blob/dev/.azure-devops/templates/set-testenv-sentinel.yml)?

azext_iot/iothub/providers/message_endpoint.py
vilit1: yes cause only cosmos db doesn't want a connection string. I use the connection string to get the primary key/symmetric key and uri props instead

azext_iot/iothub/common.py
vilit1: probably. This will have to be a future discussion that involves clean up of our files.

azext_iot/iothub/providers/message_endpoint.py
vilit1: it is also used in the begin_create_or_update but I can just replace the variable

azext_iot/tests/settings.py
vilit1: image

Also I dont even use these so I will just remove them. This is what style of test variables I wanted to go for originally

vilit1: also my changes do not affect the smoke test so anything needed for the smoke test is still needed

so we can chuck everything that is empty except for:
AZEXT_IOT_CENTRAL_APP_ID
AZEXT_IOT_TESTHUB
AZEXT_IOT_TESTDPS

(if they are missing the smoke tests will still run but there will be stderrs in the console due to missing variables)
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7158&view=logs&j=1bcc213b-07aa-5d1d-552c-52c61405b0e0&t=a12d701a-629d-5e4a-299f-a5bcd56509b9

c-ryan-k (approved):

PR #561, opened 2022-08-09
adding control plane elements (certs, identities, endpoints & routes) to state functions

msmoot0:
---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/_help.py
```diff
@@ -633,6 +633,61 @@
 """
 '''
 
+helps[
```
digimaun, line 636: Right above this copy of the `iot hub state` help, there is the original commented out version. Lets remove the duplicate.

image

azext_iot/iothub/commands_state.py
```diff
@@ -15,10 +15,10 @@ def state_export(
     resource_group_name: Optional[str] = None,
     login: Optional[str] = None,
     auth_type_dataplane: Optional[str] = None,
-    overwrite_file: Optional[bool] = False
+    force: Optional[bool] = False
```
digimaun, line 18: The use of `Optional[x]` typing is when `None` is an acceptable value. But in this case `force` is either `True` or `False`, right? If so, you can remove `Optional[]` and just keep the `bool` typing.

azext_iot/iothub/command_map.py
```diff
@@ -49,7 +49,7 @@ def load_iothub_commands(self, _):
         cmd_group.command("send", "iot_c2d_message_send")
         cmd_group.command("purge", "iot_c2d_message_purge")
 
-    # with self.command_group("iot hub state", command_type=iothub_state_ops, is_preview=True) as cmd_group:
-    #     cmd_group.command("export", "state_export")
-    #     cmd_group.command("import", "state_import")
-    #     cmd_group.command("migrate", "state_migrate")
+    with self.command_group("iot hub state", command_type=iothub_state_ops, is_preview=True) as cmd_group:
```
digimaun, line 52: nit, can we change `is_preview=True` to `is_experimental=True`?

azext_iot/tests/iothub/devices/test_hub_control_plane_export_import_int.py
```diff
@@ -0,0 +1,305 @@
...
+
+        self.filename = generate_generic_id() + ".json"
+
+    @pytest.fixture(scope="class", autouse=True)
+    def setup_class(self):
+
+        self.clean_up_hub(self.entity_name, self.entity_rg)
+
+        # add identities
+        self.uid = "/subscriptions/a386d5ea-ea90-441a-8263-d816368c84a1/resourcegroups/mirabai/providers/" + \
```
digimaun, line 58: We do not want to hard-code any resources/infrastructure in the tests themselves. Ideally they should be provisioned in JIT fashion as part of the test setup. Furthermore something like a subscription Id is an internal identifier and this is a public project.

A mode 2 for tests that we often support, is usage of environment variables that point to an existing resource i.e. if a target resource is provided via environment variable use it (and skip deleting it if user provided), otherwise provision & configure a new resource, then delete if it was provisioned for the test.

docs/samples/hub-state-import-data-sample.json
```diff
@@ -611,312 +206,375 @@
...
+      ],
+      "device2": [
+        "edge-parent1",
+        "simDevice"
+      ]
+    }
+  },
   "modules": {
     "device1-child": [],
+    "edge-parent1": [],
```
digimaun, line 412: Is this a latest snapshot of the sample json? There still appears to be a repeat of device identities (where we discussed a module identity object already has the device Id as part of it).

azext_iot/tests/iothub/devices/test_hub_control_plane_export_import_int.py
```diff
@@ -0,0 +1,305 @@
...
+from azext_iot.tests.generators import generate_generic_id
+from azext_iot.tests.helpers import add_test_tag
+from azext_iot.tests.iothub import DATAPLANE_AUTH_TYPES
+
+settings = DynamoSettings(req_env_set=ENV_SET_TEST_IOTHUB_REQUIRED, opt_env_set=ENV_SET_TEST_IOTHUB_OPTIONAL)
+CWD = os.path.dirname(os.path.abspath(__file__))
+
+CERT_FILE = "test_cert.cer"
+KEY_FILE = "test_key.cer"
+DATAPLANE_AUTH_TYPES.remove("cstring")
```
digimaun, line 20: Better to define your own auth types list rather than modify the DATAPLANE_AUTH_TYPES list. See this example https://github.com/msmoot0/azure-iot-cli-extension/blob/mirabai/azext_iot/tests/iothub/devices/test_iothub_device_tracing.py#L14

azext_iot/_help.py
```diff
@@ -633,6 +633,61 @@
...
+    - name: Import the state from the default file to the specified hub, replacing the previous state of the hub.
+      text: >
+        az iot hub state import -n {iothub_name} -f {filename} --replace
+"""
+
+helps[
+    "iot hub state migrate"
+] = """
+    type: command
+    short-summary: Copy the state of one hub to another hub without saving to a file.
```
digimaun, line 679: If the term `migrate` is preferred to `copy` we should replace the first word here.

azext_iot/iothub/providers/state.py
```diff
@@ -276,22 +452,70 @@ def upload_module_identity(self, identity: dict):
...
+
     def upload_state(self, filename: str, replace: Optional[bool] = None):
         '''
         Uses device info from file to recreate the devices
         '''
 
         if replace:
+            if self.include_control_plane:
+                with self.suppress_stderr():
+                    self.delete_all_routes()
```
digimaun, line 515: If stderr is suppressed rather than being captured, if something fails in these operations within the context scope, will the user be notified on what failed (like API messages which may give hints for resolution)?

azext_iot/iothub/providers/state.py
```diff
@@ -213,6 +282,113 @@ def save_state(self, filename: str, overwrite_file=False):
         except FileNotFoundError:
             raise FileOperationError(f'File {filename} does not exist.')
 
+    @contextmanager
+    def suppress_stderr(self):
```
digimaun, line 286: Good thought on stream-lining embedded CLI usage

azext_iot/iothub/providers/state.py
```diff
@@ -213,6 +266,120 @@ def save_state(self, filename: str, overwrite_file=False):
         except FileNotFoundError:
             raise FileOperationError(f'File {filename} does not exist.')
 
+    @contextmanager
+    def capture_stderr(self):
+        temp_file = "stderr_file" + str(random.randrange(100000000))
```
digimaun, line 271: Nice idea, but what about refactoring to use an in-memory text buffer so we don't need to deal with writing/deleting files to disk https://docs.python.org/3/library/io.html#io.StringIO

azext_iot/common/_azure.py
```diff
@@ -46,6 +46,16 @@ def parse_iot_device_module_connection_string(cs):
     return _parse_connection_string(cs, validate, "Module")
 
 
+def parse_iot_endpoint_connection_string(cs):
+    validate = ["Endpoint", "SharedAccessKeyName", "SharedAccessKey"]
```
digimaun, line 50: Because this is generically named, can we add a comment that indicates which endpoint types this function supports?

azext_iot/iothub/providers/state.py
```diff
@@ -213,6 +266,120 @@ def save_state(self, filename: str, overwrite_file=False):
         except FileNotFoundError:
             raise FileOperationError(f'File {filename} does not exist.')
 
+    @contextmanager
+    def capture_stderr(self):
```
digimaun, line 270: After making the adjust to an in-memory buffer, I think this function is generically useful and should be moved to common.utility

azext_iot/tests/iothub/devices/test_hub_export_import_int.py
```diff
@@ -16,19 +17,23 @@
...
-    DATAPLANE_AUTH_TYPES,
     DEVICE_TYPES,
 )
 
 settings = DynamoSettings(req_env_set=ENV_SET_TEST_IOTHUB_REQUIRED, opt_env_set=ENV_SET_TEST_IOTHUB_OPTIONAL)
 CWD = os.path.dirname(os.path.abspath(__file__))
 
+# changing the order so the final exported file can be used for all import tests
+# (the cstring output doesn't include the keys for the control plane features)
+DATAPLANE_AUTH_TYPES = ["cstring", "key", "login"]
```
digimaun, line 28: nit, why not use AuthenticationTypeDataplane enum attributes in common.shared rather than hard coded strings?

azext_iot/iothub/commands_state.py
```diff
@@ -15,10 +15,10 @@ def state_export(
     resource_group_name: Optional[str] = None,
     login: Optional[str] = None,
     auth_type_dataplane: Optional[str] = None,
-    overwrite_file: Optional[bool] = False
+    force: bool = False
 ):
     sp = StateProvider(cmd=cmd, hub=hub_name, rg=resource_group_name, login=login, auth_type_dataplane=auth_type_dataplane)
-    sp.save_state(filename, overwrite_file)
+    sp.save_state(filename, force)
```
digimaun, line 21: This can be done in a future iteration but I see two pivots from here in overall design:
- A user should be able to choose to not include aspects of their hub for export.
  - i.e. I have endpoints and devices but I only want to export (or migrate) devices.
- We should return a dictionary corresponding to the export data structure to make use of standard Az CLI features (like output modifiers i.e. yaml format). Also its standard shell practice to redirect output from stdout to a file. As in the filename/--state-file argument is not a necessity but a nice to have.

.azure-devops/templates/set-testenv-sentinel.yml
```diff
@@ -31,6 +31,7 @@ steps:
             "azext_dt_ep_eventhub_namespace": os.environ.get("AZEXT_DT_EP_EVENTHUB_NAMESPACE", sentinel_value),
             "azext_dt_ep_eventhub_policy": os.environ.get("AZEXT_DT_EP_EVENTHUB_POLICY", sentinel_value),
             "azext_dt_ep_eventhub_topic": os.environ.get("AZEXT_DT_EP_EVENTHUB_TOPIC", sentinel_value),
+            "azext_dt_ep_eventhub_queue": os.environ.get("AZEXT_DT_EP_EVENTHUB_QUEUE", sentinel_value),
```
digimaun, line 34: Future iteration comment - because these originally for digital twin endpoint resources can/are used across iot service areas, we should think about renaming them to something more generic rather than dt focused.

azext_iot/assets/user_messages.py
```diff
@@ -16,7 +18,7 @@ def error_no_hub_or_login_on_input(entity_type="IoT Hub"):
             "Please provide an {0} entity name (via the '{1}' or '-n' parameter)"
             " or {0} connection string via --login..."
         ).format(entity_type, PARAM_MAPPING[entity_type])
-    return ("Please provide either a hub name or a DPS name.")
+    raise InvalidArgumentValueError("Please provide either a hub name or a DPS name.")
```
digimaun, line 21: _If_ this error _should only_ raise during development (I believe that was mentioned in a prior iteration PR), then it can be a [TypeError](https://docs.python.org/3/library/exceptions.html#TypeError) with msg something like "entity_type requires a str value". the `if entity_type:` condition can be modified to `if entity_type and isinstance(entity_type, str):`. Also would be good to add the type hint to entity_type,  `entity_type: str`.

docs/samples/hub-state-import-data-sample.json
digimaun: This comment has not been resolved yet, I'm still seeing deviceId as the key to a module array

azext_iot/_params.py
```diff
@@ -297,7 +297,7 @@ def load_arguments(self, _):
         )
         context.argument(
```
vilit1, line 298: assuming this param only applies to `az iot hub state`, I would move it to be under a with statement for "az iot hub state" rather than "iot hub"

azext_iot/_params.py
```diff
@@ -306,8 +306,8 @@ def load_arguments(self, _):
             help="If this flag is set, then the command will overwrite the state of the destination hub."
         )
         context.argument(
-            "overwrite_file",
```
vilit1, line 309: assuming this param only applies to `az iot hub state`, I would move it to be under a with statement for "az iot hub state" rather than "iot hub"

vilit1: Furthermore, move the help for `az iot hub state` to azure_iot/iothub/params.py

.azure-devops/templates/set-testenv-sentinel.yml
vilit1: Added a work item https://msazure.visualstudio.com/One/_sprints/taskboard/IoT-Platform-UPX/One/Custom/IoT/2208?workitem=15274556

azext_iot/common/_azure.py
vilit1: Add it as a doc string - under the def, use """ """

azext_iot/common/utility.py
```diff
@@ -613,3 +615,21 @@ def is_valid_dtmi(dtmi):
     if not pattern.match(dtmi):
         return False
     return True
+
+
+@contextmanager
+def capture_stderr():
```
vilit1, line 621: add a doc string describing what this is for

azext_iot/assets/user_messages.py
```diff
@@ -10,13 +10,13 @@
 }
 
 
-def error_no_hub_or_login_on_input(entity_type="IoT Hub"):
-    if entity_type:
+def error_no_hub_or_login_on_input(entity_type: str = "IoT Hub"):
+    if entity_type and isinstance(entity_type, str):
```
vilit1, line 14: what is the reason for checking if the instance is an str? When should this be called with something that isnt an str? (the type hint in the param should tell users that entity_type is str so it should not be an issue)

azext_iot/iothub/command_map.py
```diff
@@ -49,7 +49,7 @@ def load_iothub_commands(self, _):
         cmd_group.command("send", "iot_c2d_message_send")
         cmd_group.command("purge", "iot_c2d_message_purge")
 
-    # with self.command_group("iot hub state", command_type=iothub_state_ops, is_preview=True) as cmd_group:
-    #     cmd_group.command("export", "state_export")
-    #     cmd_group.command("import", "state_import")
-    #     cmd_group.command("migrate", "state_migrate")
+    with self.command_group("iot hub state", command_type=iothub_state_ops, is_experimental=True) as cmd_group:
```
vilit1, line 52: We may make you re-comment this out before you - keep this comment open for now.

digimaun: This is in a feature branch now (dev reverted), so it is OK to keep it uncommented.

azext_iot/assets/user_messages.py
digimaun: To reduce mistakes and providing a clear dev error. The type hint, isinstance, and TypeError were refactored-in, in prior iteration of this PR. We aren't enforcing type checks/constraints in CI, if you think there is no chance of an issue (like there was here in the past), then I'm fine removing the isinstance check.

azext_iot/iothub/providers/state.py
```diff
@@ -51,7 +56,15 @@ def __init__(
             login=login,
             auth_type_dataplane=auth_type_dataplane
         )
+
+        if not login:
+            self.include_control_plane = True
+            self.rg = self.target["resourcegroup"]
```
vilit1, line 62: does this actually get populated? the connection string should not have a resource group and the discovery builds on that, so this should be same as self.rg = None

put prints and double check

azext_iot/iothub/providers/state.py
```diff
@@ -51,7 +56,15 @@ def __init__(
             login=login,
             auth_type_dataplane=auth_type_dataplane
         )
+
+        if not login:
```
vilit1, line 60: its easier on the eyes if you swap the if and else (so do a if login, .... else ....)

azext_iot/iothub/providers/state.py
```diff
@@ -100,27 +117,76 @@ def process_hub(self, target: dict):
 
             if id["capabilities"]["iotEdge"]:
                 children = _iot_device_children_list(target=target, device_id=id["deviceId"])
-                hub_state["children"][id["deviceId"]] = [c["deviceId"] for c in children]
+                hub_state["devices"]["children"][id["deviceId"]] = [c["deviceId"] for c in children]
 
-            hub_state["devices"][id["deviceId"]] = identities[i]
+            identities[i]["properties"].pop("reported")
+            [identities[i]["properties"]["desired"].pop(key) for key in ["$metadata", "$version"]]
```
vilit1, line 123: just expand the for loop for this and the next one - you are not saving the new list anywhere so it is awk to make a list (as you are doing here) and then throw it out.

azext_iot/iothub/providers/state.py
```diff
@@ -100,27 +117,76 @@ def process_hub(self, target: dict):
...
+
+            hub_state["certificates"] = self.cli.invoke("iot hub certificate list --hub-name {} -g {}"
+                                                        .format(hub_name, rg)).as_json()["value"]
+
+            pbar = tqdm(total=len(hub_state["certificates"]), desc="Saving certificates", file=sys.stdout)
+            pbar.update(len(hub_state["certificates"]))
+            pbar.close()
+
+            hub_state["identities"] = control_plane["identity"]
+            num_identities = (1 if hub_state["identities"]["principalId"] else 0) + \
```
vilit1, line 164: its easier on the eyes if you have something like
```
num_identites = 0
if hub_state["identities"]["prinicipalId"]:
   num_identities += 1
if hub_state["identites"]["userAssignedIdentities"]:
  num_identites += len(hub_state["identites"]["userAssignedIdentities"])
```

if you want to shorten this, then I would shorten the first if statement but not both.

azext_iot/iothub/commands_state.py
msmoot0: I'll put it in the presentation

azext_iot/assets/user_messages.py
digimaun: Talked with @vilit1 offline, we'll keep it.

azext_iot/iothub/providers/state.py
msmoot0: Not sure I understand. **self.rg = self.target["resourcegroup"]** is only running if the connection string isn't provided. If the user only gives the hub name and not the resource group, this is getting the name of the resource group.

azext_iot/iothub/providers/state.py
```diff
@@ -173,34 +239,50 @@ def upload_hub_from_dict(self, hub_state: dict):
...
             _iot_device_children_add(target=self.target, device_id=parentId, child_list=child_list)
 
-    def save_state(self, filename: str, overwrite_file=False):
+        # CONTROL PLANE
+
+        if self.include_control_plane:
+
+            with capture_stderr():
+
+                temp_cert_file = f"cert{random.randrange(100000000)}.cer"
```
vilit1, line 255: the user may have a file with the name already - you should put in the (long) help that temporary certificate files would be generated with cert{number}.cer naming pattern. These would be removed after the completion of the command.

azext_iot/iothub/providers/state.py
```diff
@@ -213,6 +295,98 @@ def save_state(self, filename: str, overwrite_file=False):
...
+            self.cli.invoke("iot hub identity assign --name {} -g {} --user-assigned {}".format(self.hub_name, self.rg, userIds))
+            pbar.update(len(identities))
+
+        pbar.close()
+
+    def upload_endpoint(self, ep, ep_type):
+
+        if ep["connectionString"]:
+
+            if ep_type == "azurestoragecontainer":
```
vilit1, line 312: azurestoragecontainer should be a constant

azext_iot/iothub/providers/state.py
```diff
@@ -100,27 +117,77 @@ def process_hub(self, target: dict):
 
             if id["capabilities"]["iotEdge"]:
                 children = _iot_device_children_list(target=target, device_id=id["deviceId"])
-                hub_state["children"][id["deviceId"]] = [c["deviceId"] for c in children]
+                hub_state["devices"]["children"][id["deviceId"]] = [c["deviceId"] for c in children]
 
-            hub_state["devices"][id["deviceId"]] = identities[i]
+            identities[i]["properties"].pop("reported")
+            [identities[i]["properties"]["desired"].pop(key) for key in ["$metadata", "$version"]]
+            [identities[i].pop(key, None) for key in ["cloudToDeviceMessageCount", "configurations", "deviceEtag", "deviceScope",
```
vilit1: the list of keys to look for should be a constant

azext_iot/iothub/providers/state.py
```diff
@@ -100,27 +117,77 @@ def process_hub(self, target: dict):
...
-                    module.pop('connection_state_updated_time')
-                    module.pop('last_activity_time')
-
                     module_twin = _iot_device_module_twin_show(target=target, device_id=id["deviceId"],
                                                                module_id=module["module_id"])
 
                     module2 = _iot_device_module_show(target=target, device_id=id["deviceId"], module_id=module["module_id"])
                     module["authentication"] = module2["authentication"]
 
+                    [module.pop(key) for key in ["connection_state_updated_time", "last_activity_time", "etag",
```
vilit1: The list of keys should be a constant, there are more things within this file that should be constants

azext_iot/iothub/providers/state.py
```diff
@@ -173,34 +239,50 @@ def upload_hub_from_dict(self, hub_state: dict):
...
 
         # set parent-child relationships
-        for parentId in hub_state["children"]:
-            child_list = hub_state["children"][parentId]
+        for parentId in hub_state["devices"]["children"]:
+            child_list = hub_state["devices"]["children"][parentId]
             _iot_device_children_add(target=self.target, device_id=parentId, child_list=child_list)
 
-    def save_state(self, filename: str, overwrite_file=False):
+        # CONTROL PLANE
```
vilit1, line 249: When we have the control plane sdk in the extension, this will drastically change (as we may be able to do more steps at once rather than one by one)

azext_iot/iothub/providers/state.py
```diff
@@ -276,22 +449,79 @@ def upload_module_identity(self, identity: dict):
...
+        eventHubs = endpoints["eventHubs"]
+        serviceBusQueues = endpoints["serviceBusQueues"]
+        serviceBusTopics = endpoints["serviceBusTopics"]
+        storageContainers = endpoints["storageContainers"]
+
+        num_endpoints = len(eventHubs) + len(serviceBusQueues) + len(serviceBusTopics) + len(storageContainers)
+        pbar = tqdm(total=num_endpoints, desc="Deleting endpoints from destination hub", file=sys.stdout)
+
+        for ep in eventHubs:
+            self.cli.invoke("iot hub routing-endpoint delete --hub-name {} -g {} --endpoint-name {} --endpoint-type \
```
vilit1, line 478: just end the string and start a new one. Dont add a bunch of empty space to strings

vilit1:
Changing to draft because this will drastically change with the generated sdk

Issue #364, opened 2021-06-03
[bug] `az iot central device-template create` does not accept DTDL v2 interfaces


rido-min:
`az iot central device-template create` does not accept DTDL v2 interfaces

**To Reproduce**

Install the latest with `az extension remove azure-iot && az extension add azure-iot`

Follow the [reference docs to create a device template](https://docs.microsoft.com/en-us/cli/azure/iot/central/device-template?view=azure-cli-latest#az_iot_central_device_template_create-required-parameters)

Use the next DTDL v2 Model as the input file

```json
{
  "@context": "dtmi:dtdl:context;2",
  "@id": "dtmi:com:example:basic;1",
  "@type": "Interface",
  "displayName": "basic",
  "contents": [
    {
      "@type": "Telemetry",
      "name": "temperature",
      "schema": "double"
    }
  ]
}
```

**Expected behavior**
A device template is added to my central app

**Observed behavior**

```text
{'code': 'InvalidBody', 
'message': "The provided request body contains the following errors: 

should have required property 'capabilityModel'; 

['@type'] should be array. 
```

**Environment (please complete the following information):**

```
C:\code> az --version
azure-cli                         2.19.1 *

core                              2.19.1 *
telemetry                          1.0.6

Extensions:
azure-iot                        0.10.13
interactive                        0.4.4

Python location 'C:\Program Files (x86)\Microsoft SDKs\Azure\CLI2\python.exe'
Extensions directory 'C:\Users\rido2\.azure\cliextensions'

Python (Windows) 3.6.8 (tags/v3.6.8:3c6b436a57, Dec 23 2018, 23:31:17) [MSC v.1916 32 bit (Intel)]
```

yang-hai-feng:
Try this
`{
	"displayName": "Themostat Template 2",
	"capabilityModel": {
		"@id": "dtmi:com:test:Thermostat;1",
		"@type": "Interface",
		"contents": [
			{
				"@type": [
					"Telemetry",
					"Temperature"
				],
				"description": "Temperature in degrees Celsius.",
				"displayName": "Temperature",
				"name": "temperature",
				"schema": "double",
				"unit": "degreeCelsius"
			}
		],
		"description": "Reports current temperature and provides desired temperature control.",
		"displayName": "Thermostat"
	},
	"@id": "dtmi:com:test:mythemostat",
	"@type": [
		"ModelDefinition",
		"DeviceModel"
	],
	"@context": [
[...]

PR #591, opened 2022-10-27
Edge device creation, configuration, and device bundles

c-ryan-k:
Adds `az iot edge devices create` command that allows users to create and configure multiple edge device scenarios (and associated device configuration settings) using a configuration file (YAML or JSON) or inline device arguments.

The command performs the following actions after processing input:
- Verify parent-child relationships between devices
- [Optional] Delete all devices on the hub before creation
- Create all new device-identities
- Assign parent scopes to devices
- Set deployments / modules on devices
- Creates device config bundle (.tgz) that contains:
  - Device certificates
  - Root CA certificate (self-signed or imported)
  - Install script and readme

---

### General Guidelines

Intent for Production

Basic expectations


c-ryan-k:
I realized there's going to be some line-length linter errors for the (quite verbose) scripts I threw in `constants` instead of copying the files from extension source to disk.

Will investigate a better solution / place for these.

azext_iot/common/shared.py
```diff
@@ -304,5 +309,43 @@ class SHAHashVersions(Enum):
     """
     Supported SHA types for generating the certificate thumbprint.
     """
+
     SHA1 = 1
     SHA256 = 256
+
+
+class CertInfo(TypedDict):
```
digimaun: We currently support Python 3.7 ...TypedDict was introduced in Python 3.8

azext_iot/constants.py
```diff
@@ -54,3 +54,127 @@
 # Initial Track 2 SDK version for DPS
 IOTDPS_MGMT_SDK_PACKAGE_NAME = "azure-mgmt-iothubprovisioningservice"
 IOTDPS_TRACK_2_SDK_MIN_VERSION = "1.0.0"
+
+# Edge device TOML default values
+DEVICE_CONFIG_TOML = {
```
digimaun: Can we put these edge config and script related attributes in a separate edge oriented module?

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,38 @@ def load_iothub_help():
...
+            az iot edge hierarchy create -n {hub_name} --clean --visualize
+            --device id=parent1 deployment=/path/to/parentDeployment1.json
+            --device id=child1 parent=parent1 deployment=/path/to/child_deployment1.json
+            --device id=parent2 deployment=/path/to/parentDeployment2.json
+            --device id=child2 parent=parent2 deployment=/path/to/child_deployment2.json
+
+        - name: Create a simple nested edge device hierarchy (deleting existing devices)
+            with an existing root CA, using x509 auth, and specify a custom device bundle output path.
+          text: >
+            az iot edge hierarchy create -n {hub_name} --clean --root-cert "root_cert.pem" --root-key "root_key.pem" --device-auth x509_ca
```
c-ryan-k: I need to add a help example for extra `--device` arguments (hostname, edge_agent, container_auth)

azext_iot/iothub/commands_device_identity.py
```diff
@@ -0,0 +1,68 @@
...
+        clean=clean,
+        visualize=visualize,
+        auth_type=device_auth_type,
+        root_cert_path=root_cert_path,
+        root_key_path=root_key_path,
+        output_path=bundle_output_path,
+    )
+
+
+def iot_delete_devices(
```
c-ryan-k, line 56: May remove this provider method until we decide on a suitable namespace for the command, like `az iot hub device-identity clean/delete --all` etc.

azext_iot/tests/iothub/devices/hierarchy_configs/fake_edge_container_auth.json
```diff
@@ -0,0 +1,5 @@
+{
```
c-ryan-k: Add to credscan suppressions / document mock creds to avoid pipeline failures

test.yml
```diff
@@ -0,0 +1,31 @@
+config_version: "1.0"
```
c-ryan-k: Will remove this file for merge

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,658 @@
...
+        Process edge config file schema dictionary
+        """
+        # TODO - version / schema validation
+        version = content.get("config_version", None)
+        hub_config = content.get("iothub", None)
+        devices_config = content.get("edgedevices", [])
+        for check, err in [
+            (version, "No schema version specified in configuration file"),
+            (hub_config, "No `iothub` properties specified in configuration file")
+            # (len(devices_config), "No devices specified in configuration file")
```
c-ryan-k: We should probably throw an exception if a config file is used without any devices - going to uncomment this and update tests.

azext_iot/common/shared.py
vilit1: one option is to say we only support 3.8 now

.gitignore
```diff
@@ -289,8 +289,8 @@ __pycache__/
 
 # Virtual environment
 env/
-env27/
```
vilit1, line 292: has python 2 not been removed?

c-ryan-k: Not in our .gitignore, at least - I can remove the newly added line if we want, just figured maybe there's someone out there with a 2.x env folder floating around :)

azext_iot/common/shared.py
digimaun: Disagree, we should deprecate python versions when they aren't supported anymore. Python 3.7 EOL is in 8 months.

c-ryan-k: The beauty of TypedDict is that we can just replace it with `Dict[str, str]` or `Dict[str, Any]` - it was just a convenience type for the certificate object format we pass around. I'll get rid of the TypedDict for now.

azext_iot/constants.py
c-ryan-k: I noticed that in another PR we have a new `common.py` file created in `azext_iot/iothub` - do you think this is the right spot for these, or should we:
- Add a `common` folder with (examples):
  - `hub.py`
  - `devices.py`
  - `edge.py`
 - Create a new `edge_common.py` strictly for edge-related common code.
 We could also move some of the `device_identity.py` provider functions into this file , like `_process_nested_edge_config_content` or `create_edge_device_config`

azext_iot/iothub/commands_device_identity.py
digimaun: Since it takes a list of device Id's (its not always all devices) I could see it being useful now, as in `az iot hub device-identity delete` today takes a single deviceId , we could flex it to take 1 or more values. Someone could delete 1 or 1000 in a single command execution. Have to think about the etag though. 2 immediate thoughts are to keep an optional etag single value like now (where it would be useful for a single device delete), or also flex etag to take 1 or more values, where the etag value index corresponds to the device id value index. If the length of values mismatch raise an error.

azext_iot/constants.py
digimaun: What is different about these newly introduced elements is they are focused around device side scripting & configuration which IMO does not harmonize with the organization of common modules in iothub, deviceupdate, digitaltwins, central etc -- those are in the realm of enums (often times to align service & client options), resource ids, naming functions, generic messages (arguably can be moved) and namespace specific client settings.

More prefer the second option, but for these specific attributes, something like `edge_device_config.py` seems appropriate because just reading the module name I have an idea it will be edge device configuration centric where `edge_common` can be ambiguous with respect to what part of edge overall (service side?) + what I was saying earlier where it wouldn't be organized like our other namespace specific common modules.  I also do like the thought of moving the edge config specific functions/classes into an edge config based module rather than mixing with device_identity.

azext_iot/common/shared.py
digimaun: Makes sense. I'm all for using the best facilities of Python we can, but we have to balance with compatibility and casting the widest (safe/secure) net of usage. We saw offline there are **still** notable usage of legacy/out-of-support Py versions.

c-ryan-k: Yup - totally agree. I actually didn't see your comment when I typed mine, so I was just reiterating support for waiting to use it until we're able to ensure compatibility. If anything, it was worth using during dev for the intellisense alone 😆

azext_iot/iothub/commands_device_identity.py
c-ryan-k: Hmm..we could use `eTag` as `*` in this command for the user (letting users know it won't check eTags, of course) or maybe support eTags as another param like `az iot hub device-identity delete --device id=test eTag=_etag_`

digimaun: For the first idea, right now for the device identity delete command, if the user does not provide an etag then * will be assumed - are you thinking if multiple device id's are provided, then the only choice is '*' + let users know of that etag behavior change?

For the second one, an initial though is it _could_ be confusing have both `--device-id` and `--device k=v` styles but we we can simmer on it.

We don't need to make a decision in this PR.

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,81 @@ def load_iothub_arguments(self, _):
...
+        )
+        context.argument(
+            "bundle_output_path",
+            options_list=[
+                "--output-path",
+                "--out",
+            ],
+            help="Directory path to output device configuration bundles.",
+        )
+        context.argument(
```
digimaun, line 311: Any opportunities for organizing related arguments as arg groups? For example an arg group for cert related args.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,43 @@ def load_iothub_help():
         type: command
         short-summary: Upload a local file as a device to a pre-configured blob storage container.
     """
+
+    helps["iot edge hierarchy"] = """
```
digimaun: We discussed the command name change making sure there is ample examples of hierarchy based + non-hierarchy based edge device configuration examples.

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,409 @@
...
+        super(DeviceIdentityProvider, self).__init__(
+            cmd=cmd,
+            hub_name=hub_name,
+            rg=rg,
+            login=login,
+            auth_type_dataplane=auth_type_dataplane,
+        )
+
+        self.hub_resolver = SdkResolver(target=self.target)
+        self.service_sdk = self.hub_resolver.get_sdk(SdkType.service_sdk)
```
digimaun: SdkResolver is already available in the base IoTHubProvider class, you can use the base method `get_sdk()`

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,409 @@
...
+
+    def create_edge_hierarchy(
+        self,
+        devices: Optional[List[List[str]]] = None,
+        config_file: Optional[str] = None,
+        visualize: Optional[bool] = False,
```
digimaun: nit, because the default value False is assigned, I don't see the semantic value that Optional provides for the bool.

azext_iot/common/utility.py
```diff
@@ -146,6 +150,44 @@ def process_json_arg(content: str, argument_name: str = "content", preserve_orde
...
+def process_toml_content(path: str) -> Dict[str, Any]:
+    """Primary processor of TOML file input"""
+
+    if not os.path.exists(path):
+        raise FileOperationError(
+            f"TOML file not found - Please ensure the path '{path}' is correct."
+        )
+
+    try:
+        import toml
```
digimaun: Where is the `toml` dependency coming from? If from az cli which part of it?

azext_iot/common/utility.py
```diff
@@ -146,6 +150,44 @@ def process_json_arg(content: str, argument_name: str = "content", preserve_orde
         )
 
 
+def process_yaml_arg(path: str) -> dict:
```
digimaun: nit, might as well use a Dict generic for return type like in the process toml function

azext_iot/common/utility.py
```diff
@@ -146,6 +150,44 @@ def process_json_arg(content: str, argument_name: str = "content", preserve_orde
...
+
+        with open(path, "rb") as f:
+            return yaml.load(f, Loader=yaml.SafeLoader)
+    except Exception as ex:
+        raise InvalidArgumentValueError(
+            "Failed to parse yaml for file '{}' with exception:\n{}".format(path, ex)
+        )
+
+
+def process_toml_content(path: str) -> Dict[str, Any]:
```
digimaun: nit, do you think ending in `_content` is appropriate? Just curious since we have process_yaml_**arg** and process_json_**arg**.

azext_iot/common/shared.py
```diff
@@ -304,5 +309,33 @@ class SHAHashVersions(Enum):
     """
     Supported SHA types for generating the certificate thumbprint.
     """
+
     SHA1 = 1
     SHA256 = 256
+
+
+# Utility classes for nested edge config file values and device arguments
+class EdgeContainerAuth(NamedTuple):
```
digimaun: Is a EdgeContainerAuth valuable if no attributes are set? Typing shows nothing is required (everything optional).

azext_iot/common/fileops.py
```diff
@@ -0,0 +1,47 @@
...
+    write_content = bytes(content, "utf-8") if isinstance(content, str) else content
+    with open(file_path, "wb") as f:
+        f.write(write_content)
+
+
+def tar_directory(
+    target_directory: str,
+    tarfile_path: str,
+    tarfile_name: str,
+    overwrite: Optional[bool] = False,
```
digimaun: Another place (and above function) where I think we can remove Optional.

azext_iot/tests/iothub/devices/hierarchy_configs/invalid/duplicate_device_config.yml
```diff
@@ -0,0 +1,16 @@
+config_version: "1.0"
+# We don't need iothub info from the config schema
```
digimaun: We talked about exposing one or more markdown document at least in our project docs that shows or links to example(s) of the config file format and describes key usage information.

To do this, we should concretely pin the structure of the format and version it. Best practice conventions should be applied, and because the CLI supports either json or yaml, I don't think its a bad idea to align on camelCase since that is an accepted practice for both json and yaml.

azext_iot/common/utility.py
```diff
@@ -146,6 +150,44 @@ def process_json_arg(content: str, argument_name: str = "content", preserve_orde
...
+
+    if not os.path.exists(path):
+        raise FileOperationError(
+            f"TOML file not found - Please ensure the path '{path}' is correct."
+        )
+
+    try:
+        import toml
+
+        with open(path, "rt", encoding="utf-8") as f:
```
digimaun: Think "r" is a synonym for "rt" but its good to be explicit.

azext_iot/common/shared.py
```diff
@@ -304,5 +309,33 @@ class SHAHashVersions(Enum):
...
+    serveraddress: Optional[str] = None
+    username: Optional[str] = None
+    password: Optional[str] = None
+
+
+class NestedEdgeDeviceConfig(NamedTuple):
+    device_id: str
```
digimaun: The minimum operation with just a device Id is to create an edge enabled device, is that right?

azext_iot/common/shared.py
```diff
@@ -304,5 +309,33 @@ class SHAHashVersions(Enum):
     """
     Supported SHA types for generating the certificate thumbprint.
     """
+
     SHA1 = 1
     SHA256 = 256
+
+
+# Utility classes for nested edge config file values and device arguments
```
digimaun: I think these utility namedtuples should be located within the `iothub` namespace. This module is legacy and is in large part supporting the classic operations/* implementation style. Placing them either in `edge_device_config` or a new `.../iothub/common.py` would be more optimal.

azext_iot/common/certops.py
```diff
@@ -125,3 +134,198 @@ def open_certificate(certificate_path: str) -> str:
         raise ValueError("Certificate file type must be either '.pem' or '.cer'.")
     # Remove trailing white space from the certificate content
     return certificate.rstrip()
+
+
+# TODO - Unit test, compare with test_utils::_generate_root_certificate
```
digimaun: Tiny note to remove TODOs **if** they have been satisfied for less future confusion.

azext_iot/common/certops.py
```diff
@@ -125,3 +134,198 @@ def open_certificate(certificate_path: str) -> str:
...
+def create_root_certificate(
+    subject: Optional[str] = "Azure_IoT_CLI_Extension_Cert",
+    valid_days: Optional[int] = 365,
```
digimaun: Is `Optional` appropriate here? What does `valid_days = None` mean?

azext_iot/common/certops.py
```diff
@@ -125,3 +134,198 @@ def open_certificate(certificate_path: str) -> str:
         raise ValueError("Certificate file type must be either '.pem' or '.cer'.")
     # Remove trailing white space from the certificate content
     return certificate.rstrip()
+
+
+# TODO - Unit test, compare with test_utils::_generate_root_certificate
+def create_root_certificate(
+    subject: Optional[str] = "Azure_IoT_CLI_Extension_Cert",
```
digimaun: A subject str is always required right? We should only use `Optional` if `None` is allowed.

azext_iot/common/certops.py
```diff
@@ -125,3 +134,198 @@ def open_certificate(certificate_path: str) -> str:
...
+    valid_days: Optional[int] = 365,
+    key_size: Optional[int] = 4096,
```
digimaun: optional key_size makes sense if say the parameter was assigned a value of None, then if falsey, the function would assign  key_size the value of 4096.

azext_iot/iothub/edge_device_config.py
```diff
@@ -0,0 +1,469 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
```
digimaun: Seems appropriate to put this module in providers/, thoughts?

azext_iot/iothub/edge_device_config.py
```diff
@@ -0,0 +1,469 @@
...
+    echo "Invalid hostname $hostname"
+    exit 1
+fi
+
+sed -i "s/{{HOSTNAME}}/$hostname/" /etc/aziot/config.toml
+"""
+EDGE_CONFIG_SCRIPT_PARENT_HOSTNAME = """
+# ======================= Set Parent Hostname =======================================
+
+read -p "Enter the parent hostname to use: " parent_hostname
```
digimaun, line 85: What about ensuring a device or its parent hostname is known? As in are `EDGE_CONFIG_SCRIPT_PARENT_HOSTNAME` and `EDGE_CONFIG_SCRIPT_HOSTNAME` must haves?

azext_iot/iothub/edge_device_config.py
```diff
@@ -0,0 +1,469 @@
...
+
+echo "To check the edge runtime status, run 'iotedge system status'. To validate the configuration, run 'sudo iotedge check'"
+"""
+
+EDGE_SUPPORTED_OS_LINK = (
+    "https://docs.microsoft.com/en-us/azure/iot-edge/support?view=iotedge-2020-11"
+)
+
+EDGE_LINUX_TUTORIAL_LINK = (
+    "https://learn.microsoft.com/en-us/azure/iot-edge/how-to-provision-single-device-linux-symmetric"
```
digimaun: I wonder if these links should be aka.ms ones, unless they will stay the same forever.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,43 @@ def load_iothub_help():
...
+        type: group
+        short-summary: Commands to deploy and manage nested IoT Edge device hierarchies.
+    """
+
+    helps["iot edge hierarchy create"] = """
+        type: command
+        short-summary: |
+          Deploy nested edge hierarchies to an IoT Hub.
+          This can be done with inline device arguments or by supplying a nested edge configuration file in YAML or JSON format.
```
digimaun: `long-summary` can be used where there is ample enough information to provide in help. Short-summary for each command will be shown in its entirety when showing help for a command group.

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,43 @@ def load_iothub_help():
...
+          Review examples and parameter descriptions for details on how to fully utilize this operation.
+
+        examples:
+        - name: Delete all existing devices in the Hub, and create a simple nested edge device hierarchy
+            containing 2 parent devices with 1 child device each. Also specifies module deployments
+            to configure the devices and visualizes the operations as they progress.
+          text: |
+            az iot edge hierarchy create -n {hub_name} --clean --visualize
```
digimaun: General question, currently the device bundle output is always produced right? If so, should there be an opt-in way to activate, or a way to deactivate? 

Scenario I'm thinking - what if a user wants to create or update a set of edge device identities but don't need the device side setup scripts?

azext_iot/common/utility.py
c-ryan-k: I can change it to `_arg` - I think I chose this initially since it was a bit different than JSON (which can be passed in as a string or a path) - though YAML is processed the same way so I just missed one of these 🙃

azext_iot/common/shared.py
c-ryan-k: Sorry that was a bit of laziness left over from some object prototyping, I *believe* all of these are required, but there's a chance that user/pass may be optional - will investigate and update.

azext_iot/common/certops.py
c-ryan-k: Like you've probably seen elsewhere in my usage of `Optional`, this is likely my Typescript brain fighting with my Python brain :)

I'll make another pass at updating method signatures and object/tuple definitions.

azext_iot/iothub/_help.py
c-ryan-k: I agree, this was something that I realized as well - we should either:
- Specify that the device bundle output always happens unless a `--no-output`-esque param is provided
- Only create device bundle output if a path is provided.

azext_iot/common/utility.py
c-ryan-k: I could have sworn it was coming from azure-cli but I think I just realized it's coming from `azdev` - might need to add this to setup / etc

c-ryan-k: After some brief searching, I may look into `tomli/tomli-w` as well, I was only using `toml` since I figured it was already in our dependency list

azext_iot/common/shared.py
c-ryan-k: Yes, correct, technically that is the only property you need in the config to create devices.
For this scenario, they're all assumed edge devices, but this could easily be `DeviceConfig` with a property of `edge: bool`

azext_iot/common/utility.py
digimaun: Sounds good `tomli[-w]` looks cool

azext_iot/iothub/edge_device_config.py
c-ryan-k: Now that I've finished a bunch of other renaming / changes - I'm still thinking about this.

Right now the providers folder only contains actual provider modules (`base`, `device_identity`, `device_messaging`, etc.).

Since this code is used almost exclusively by the new `device_identity` provider, but isn't a provider itself, it seems odd to just throw it in that folder... would it make sense to move some providers into folders (`providers/device_identity/device_identity_provider.py`) in case we extend them? Or maybe just another folder inside `providers` - `providers/[utility | common | helpers]`?

digimaun: The first option (`providers/device_identity/device_identity_provider.py`) is not my favorite because it looks too nested and repeated.

`from iothub.providers.device_identity.device_identity_provider` ...I know I'm importing an iothub provider from the first 2 segments.

And IMO we don't have enough independent modules around device identity at this stage to create a new nested directory.

Then `common` is overloaded -  I'm good with putting it in a new helpers directory nested under providers or even as a sibling to providers i.e.  ........`iothub/providers/helpers/edge_device_config.py` or `iothub/helpers/edge_device_config.py`

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
         raise ValueError("Certificate file type must be either '.pem' or '.cer'.")
     # Remove trailing white space from the certificate content
     return certificate.rstrip()
+
+
+def create_root_certificate(
```
vilit1: can we combine this with create_self_signed_certificate and/or replace it? There is some customization that would need to be added (such as file prefix, etc)

It may also be helpful to add docstrings here as to what this will be used for (ex: create a root certificate that will be uploaded to the iot hub)

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+    ).decode("utf-8")
+
+    return {
+        "certificate": certificate,
+        "privateKey": private_key,
+        "thumbprint": thumbprint,
+    }
+
+
+def create_signed_cert(
```
vilit1: It may also be helpful to add docstrings here as to what this will be used for (ex: create a device certificate signed by a root certificate in the iot hub and use that for device registration)

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+        .add_extension(key_usage, critical=True)
+        .sign(key, hashes.SHA256())
+    )
+    certificate = cert.public_bytes(serialization.Encoding.PEM).decode("utf-8")
+    thumbprint = cert.fingerprint(hashes.SHA256()).hex().upper()
+    private_key = key.private_bytes(
+        encoding=serialization.Encoding.PEM,
+        format=serialization.PrivateFormat.TraditionalOpenSSL,
+        encryption_algorithm=serialization.NoEncryption(),
+    ).decode("utf-8")
```
vilit1: are we not dumping/writing root certificate and key?

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+def create_signed_cert(
+    subject: str,
+    ca_public: str,
+    ca_private: str,
+    cert_output_dir: Optional[str] = None,
+    cert_file: Optional[str] = None,
+    valid_days: Optional[int] = 365,
+) -> Dict[str, str]:
+
+    private_key = rsa.generate_private_key(public_exponent=65537, key_size=4096)
```
vilit1: how come we can customize key_size for root but not here? Do we need to be able to customize key_size?

azext_iot/common/certops.py
vilit1: also this looks like something that can replace the other signed cert function

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+    }
+
+
+def load_ca_cert_info(
+    cert_path: str, key_path: str, password: Optional[str] = None
+) -> Dict[str, str]:
+    for path in [cert_path, key_path]:
+        if not exists(path):
+            raise FileOperationError(
+                "Error loading certificates. " f"No file found at path '{path}'"
```
vilit1: why not one string

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
         raise ValueError("Certificate file type must be either '.pem' or '.cer'.")
     # Remove trailing white space from the certificate content
     return certificate.rstrip()
+
+
+def create_root_certificate(
+    subject: str = "Azure_IoT_CLI_Extension_Cert",
+    valid_days: int = 365,
+    key_size: int = 4096,
+) -> Dict[str, str]:
```
vilit1: do we want to be able to create certs with passwords? (this applies to the function after this too)

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+            file_name=f"{cert_file or subject}.key.pem",
+        )
+    return {
+        "certificate": certificate,
+        "thumbprint": thumbprint,
+        "privateKey": privateKey,
+    }
+
+
+def load_ca_cert_info(
```
vilit1, line 272: add a short docstring to describe this vs open_certificate uses

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+        "privateKey": key,
+    }
+
+
+def make_cert_chain(
+    certs: List[str],
+    output_dir: Optional[str] = None,
+    output_file: Optional[str] = None,
+) -> str:
+    cert_content = "".join(certs)
```
vilit1, line 337: I am assuming this works cause I do not remember if the certs need to have a newline between all of them...

also interesting that you have this separate from creating signed certificates - even though it does make the signed certificates easier to understand cause one thing is being done, usually the signed certificates (specifically device) need a cert chain...

azext_iot/common/certops.py
```diff
@@ -125,3 +134,201 @@ def open_certificate(certificate_path: str) -> str:
...
+        "thumbprint": thumbprint,
+        "privateKey": key,
+    }
+
+
+def make_cert_chain(
+    certs: List[str],
+    output_dir: Optional[str] = None,
+    output_file: Optional[str] = None,
+) -> str:
```
vilit1, line 325: for things as scary as certs, it is safer to have a short docstring with what this is doing and where it will be used

azext_iot/common/fileops.py
```diff
@@ -0,0 +1,47 @@
...
+    if exists(file_path) and not overwrite:
+        raise FileOperationError(f"File already exists at path: {file_path}")
+    if overwrite and destination:
+        makedirs(destination, exist_ok=True)
+    write_content = bytes(content, "utf-8") if isinstance(content, str) else content
+    with open(file_path, "wb") as f:
+        f.write(write_content)
+
+
+def tar_directory(
```
vilit1, line 37: why do we need tar files...

azext_iot/common/utility.py
```diff
@@ -146,6 +150,44 @@ def process_json_arg(content: str, argument_name: str = "content", preserve_orde
...
+        )
+
+    try:
+        import yaml
+
+        with open(path, "rb") as f:
+            return yaml.load(f, Loader=yaml.SafeLoader)
+    except Exception as ex:
+        raise InvalidArgumentValueError(
+            "Failed to parse yaml for file '{}' with exception:\n{}".format(path, ex)
```
vilit1: YAML?

Also we could make these errors into constants since these and the corresponding toml ones are mostly the same; this will be helpful if we decide to support other file types too

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
         type: command
         short-summary: Upload a local file as a device to a pre-configured blob storage container.
     """
+
+    helps["iot edge devices"] = """
+        type: group
+        short-summary: Commands to create and manage IoT Edge devices.
```
vilit1: right now there is only create - is this expected?
Can this summary just say manage (without the create)

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
...
+    """
+
+    helps["iot edge devices create"] = """
+        type: command
+        short-summary: Create configuratations of edge devices in an IoT Hub.
```
vilit1: creating configurations of edge devices != creating edge devices, right?

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
             help="MIME Type of file.",
         )
+
+    with self.argument_context("iot edge devices") as context:
+        context.argument(
+            "devices",
+            options_list=["--device", "-d"],
+            nargs="+",
+            action="append",
+            help="Space-separated key=value pairs corresponding to properties of the edge device to create. "
```
vilit1, line 258: what is the max amount of devices a person can add at a time?

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
...
+
+        - name: Delete all existing devices on a hub and create new devices based on a configuration file (with progress bars and visualization).
+          text: >
+            az iot edge devices create -n {hub_name} --cfg path/to/config_yml_or_json -c -v
+
+        - name: Create a group of nested edge devices with custom module deployments - containing 2 parent devices with 1 child device each, using inline arguments.
+            Also specifies output path for device certificate bundles.
+          text: |
+            az iot edge devices create -n {hub_name} --out {device_bundle_path}
+            --device id=parent_1 deployment=/path/to/parentDeployment_1.json
```
vilit1, line 353: we may need a warning in help (long summary?) for a person trying to add too many devices and hitting the command line input limit

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
...
+        - name: Create a couple of edge devices using symmetric key auth (default)
+          text: |
+            az iot edge devices create -n {hub_name} --device id=device_1 --device id=device_2
+
+        - name: Create a flat list of edge devices using self-signed certificate authentication with various edge property configurations, using inline arguments.
+          text: |
+            az iot edge devices create -n {hub_name} --device-auth x509_thumbprint --default-edge-agent {default_edge_agent}
+            --device id=device_1 hostname={FQDN}
+            --device id=device_2 edge_agent={custom_agent}
+            --device id=parent hostname={FQDN} edge_agent={custom_agent} container_auth={path_or_json_string}
```
vilit1: I hate commands that have key=value args but this is probably for the best here

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
...
+        - name: Create a group of nested edge devices with custom module deployments - containing 2 parent devices with 1 child device each, using inline arguments.
+            Also specifies output path for device certificate bundles.
+          text: |
+            az iot edge devices create -n {hub_name} --out {device_bundle_path}
+            --device id=parent_1 deployment=/path/to/parentDeployment_1.json
+            --device id=child_1 parent=parent_1 deployment=/path/to/child_deployment_1.json
+            --device id=parent_2 deployment=/path/to/parentDeployment_2.json
+            --device id=child_2 parent=parent_2 deployment=/path/to/child_deployment_2.json
+
+        - name: Create a simple nested edge device configuration with an existing root CA, using x509 auth, and specify a custom device bundle output path.
```
vilit1, line 358: Is there a way to do this while generating a root CA for the hub? (Assuming that the user does not have the root ca)

What if the user has the device certs and wants to use those?

azext_iot/iothub/_help.py
```diff
@@ -309,3 +309,50 @@ def load_iothub_help():
...
+            az iot edge devices create -n {hub_name} --device id=device_1 --device id=device_2
+
+        - name: Create a flat list of edge devices using self-signed certificate authentication with various edge property configurations, using inline arguments.
+          text: |
+            az iot edge devices create -n {hub_name} --device-auth x509_thumbprint --default-edge-agent {default_edge_agent}
+            --device id=device_1 hostname={FQDN}
+            --device id=device_2 edge_agent={custom_agent}
+            --device id=parent hostname={FQDN} edge_agent={custom_agent} container_auth={path_or_json_string}
+
+        - name: Delete all existing devices on a hub and create new devices based on a configuration file (with progress bars and visualization).
```
vilit1: cant we just tell the user "run the delete command first" (this will require creating said delete command) rather than slapping this into create and making it have more switches/toggles?

azext_iot/iothub/common.py
```diff
@@ -0,0 +1,34 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+# Utility classes for edge device configs
```
vilit1: move this comment to the actual named tuples

also it would be nice to make a named tuple for certificate props that we care about (tho it can be a future task)

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
+            "The following key values are supported: `id`, `deployment` (inline json or path to file), `hostname`, "
+            "`parent`, `edge_agent`, and `container_auth` (inline json or path to file). "
```
vilit1: is there a reference with respect to what each of these key values mean?


Here is a location help from `az cosmosdb create` that may help with formatting
image

vilit1: will look for a better example

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
+            "--device can be used 1 or more times. Review help examples for parameter usage."
+        )
+        context.argument(
+            "clean",
+            options_list=["--clean", "-c"],
+            arg_type=get_three_state_flag(),
+            help="Deletes all devices in target hub before creating new devices.",
```
vilit1, line 269: this sounds like a command that needs a prompt... This goes back to why can't we just have a delete command

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
+            "(device creation, setting parents, updating configs, etc).",
+        )
+        context.argument(
+            "config_file",
+            options_list=["--config-file", "--config", "--cfg"],
+            help="Path to devices configuration file. Sample configuration file: "
+            "https://github.com/Azure/azure-iot-cli-extension/tree/dev/docs/samples/sample_devices_config.yaml",
+        )
+        context.argument(
+            "device_auth_type",
```
vilit1, line 285: can a person create a sym key device and a cert device in the same command? 

Not sure if we should be explicit here with "Device to hub authorization mechanism for all devices created in this command run" (will need to tweak wording)

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
+            "visualize",
+            options_list=["--visualize", "--vis", "-v"],
+            arg_type=get_three_state_flag(),
+            help="Shows visualizations of devices and progress of various tasks "
+            "(device creation, setting parents, updating configs, etc).",
+        )
+        context.argument(
+            "config_file",
+            options_list=["--config-file", "--config", "--cfg"],
+            help="Path to devices configuration file. Sample configuration file: "
```
vilit1, line 281: are there any specifics for config files with respect to devices? Ex: each device must have a config associated with it

azext_iot/iothub/params.py
```diff
@@ -246,3 +246,88 @@ def load_iothub_arguments(self, _):
...
+            help="Device to hub authorization mechanism.",
+        )
+        context.argument(
+            "default_edge_agent",
+            options_list=["--default-edge-agent", "--default-agent", "--dea"],
+            help="Default edge agent for created Edge devices if not specified individually.",
+        )
+        context.argument(
+            "device_config_template",
+            options_list=["--device-template", "--dt"],
```
vilit1: whats the difference between this and config file?

setup.py
```diff
@@ -53,7 +53,10 @@
     "jsonschema~=3.2.0",
     "importlib_metadata;python_version<'3.8'",
     "azure-iot-device~=2.11",
+    "tomli",
+    "tomli-w",
     "tqdm~=4.62",
+    "treelib",
```
vilit1: would like to see a pipeline run with 3.7 to make sure everything still works

azext_iot/iothub/_help.py
digimaun: Hate is a strong word, why so? Assuming key value pairs are well defined and there are examples -  there is flexibility being able to support both inline and from file device definition.

azext_iot/common/certops.py
c-ryan-k: Not sure, this might have just been an autoformat / linting miss.

azext_iot/common/fileops.py
c-ryan-k: The end result of `edge devices create` if an `--output` is provided is both the Hub configuration changes (create device, set parents, set modules, etc) as well as a device bundle to install on edge devices themselves, to setup local edge agent configs and certificate paths.

c-ryan-k: So since most edge use-cases are on linux endpoints/devices - we decided to use tar to compress the device bundles.

azext_iot/iothub/_help.py
c-ryan-k: Yeah I can see how this could be confused with "Edge Configurations" but this is creating... (not hierarchies, but... topologies?)

I need to think of a better word than "collection", "group", or "a bunch of"

azext_iot/iothub/_help.py
c-ryan-k: We need to be careful not to confuse the Edge device root cert (which is different than a certificate stored on the hub) with the hub root CA - the certificates created by this command are purely for either: edge-to-edge communication (all signed by the same self-signed or imported root certificate) or for edge-to-hub device identity certificates.

Regarding bring-your-own device certs, currently you can regenerate all the device certs based off of an existing root cert you import with the command, but there is no mechanism to individually specify existing certificates per-device.

azext_iot/iothub/params.py
c-ryan-k: There is technically no defined maximum, you are only limited by input configuration / shell / how long your device arguments are, etc.

azext_iot/iothub/params.py
c-ryan-k: This is the main configuration file that contains your YAML or JSON list of devices, think of it as a "device deployment playbook".

azext_iot/iothub/params.py
c-ryan-k: > can a person create a sym key device and a cert device in the same command?

Not in the same command, no. You can make one run of symmetric key devices, then another run for cert devices, etc - but only one auth type per `devices create` command.

azext_iot/iothub/params.py
c-ryan-k: This is the edge `config.toml` device configuration template used as a basis for each edge [device's config](https://learn.microsoft.com/en-us/azure/iot-edge/how-to-provision-single-device-linux-x509?view=iotedge-1.4&tabs=azure-portal%2Cubuntu#provision-the-device-with-its-cloud-identity):

See [device_config.toml](https://github.com/Azure/azure-iot-cli-extension/pull/591/files/db13535a20ae1d5f600b1fd6b67c127d79226cca#diff-7ca392c8d88a5cc372d64ce3e35b9ede99a5ef0f7951edd65f264e9bf4808cf1)

azext_iot/iothub/_help.py
c-ryan-k: Correct, this simply adds the ability to create groups of new edge devices on a hub (and at the same time clean the hub if necessary). Can definitely change to manage, felt like create was the star here so hence it's inclusion :)

azext_iot/iothub/_help.py
c-ryan-k: For the sake of this command, since it's being used to create specifically designated device scenarios, I would argue it's acceptable to have a `--clean` switch here. That's not to say we shouldn't also eventually have a `devices clean` command, but I don't think adding one extra switch to a pretty hefty "infra as config" type command complicates it too much

azext_iot/iothub/params.py
c-ryan-k: These are the same values that are in the configuration yaml/json schema per-device, so perhaps we could just add a note near the link to that sample file that declares that they are the same properties / types as in that schema.

azext_iot/iothub/params.py
c-ryan-k: Agree, we have talked offline about adding a confirmation to this command if `--clean` is provided.

azext_iot/common/certops.py
c-ryan-k: For these commands (since we create a lot of certs) I made a separate `fileops` command to`write_content_to_file`, so we can do it separately, doesn't necessarily need to be an overload on this command.

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,426 @@
...
+
+        # Show the device tree
+        if visualize:
+            tree.show()
+
+        # Query existing devices
+        query_args = ["SELECT deviceId FROM devices"]
+        query_method = self.service_sdk.query.get_twins
+        existing_devices = _execute_query(query_args, query_method)
+        existing_device_ids = list(map(lambda x: x["deviceId"], existing_devices))
```
vilit1: `[x for x["deviceId"] in existing_devices]` does not work?

azext_iot/common/utility.py
c-ryan-k: Not sure if this is a question on the capitalization (or lack thereof) of `YAML` or a question on the usage of it as a file format 😆

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -0,0 +1,577 @@
...
+iotedge config apply -c /etc/aziot/config.toml
+
+echo "To check the edge runtime status, run 'iotedge system status'. To validate the configuration, run 'sudo iotedge check'"
+"""
+
+EDGE_SUPPORTED_OS_LINK = "https://aka.ms/iotedge-supported-systems"
+EDGE_LINUX_TUTORIAL_LINK = "https://aka.ms/iotedge-provision-linux-device"
+EDGE_WINDOWS_TUTORIAL_LINK = "https://aka.ms/iotedge-provision-windows"
+
+DEVICE_README = f"""
```
vilit1, line 208: should this be a separate file rather than a constant?

vilit1: lots of these constants are scripts which make me question if they should be files instead

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -0,0 +1,577 @@
...
+    # load default device TOML object or custom path
+    device_toml = (
+        process_toml_arg(device_config_path)
+        if device_config_path
+        else DEVICE_CONFIG_TOML
+    )
+
+    device_toml[
+        "trust_bundle_cert"
+    ] = f"file:///etc/aziot/certificates/{EDGE_ROOT_CERTIFICATE_FILENAME}"
```
vilit1, line 260: is this file path going to always be the same for any system?

c-ryan-k: yes, this is the path the edge agent uses to store certificates and a copy of the device config toml.

azext_iot/iothub/providers/helpers/edge_device_config.py
c-ryan-k: Had a brief conversation about this before you left re: storing files in the extension, and then reading / writing them to device configs, or storing them as a string in code so we can just write to new files without reading the existing ones.

@digimaun curious if you have any thoughts here? I'm unsure of implications of including additional source files for these scripts. It would certainly make them easier to edit and format (and would probably save the linter some struggles) but is that worth the additional read steps every time we need to write these to a device bundle?

azext_iot/tests/utility/test_iot_file_operations.py
```diff
@@ -0,0 +1,59 @@
...
+    )
+    def test_write_content_to_file(self, set_cwd, content, destination, file_name, overwrite, error, remove_after_test):
+        try:
+            write_content_to_file(content=content, destination=destination, file_name=file_name, overwrite=overwrite)
+            with open(join(destination, file_name), "rt", encoding="utf-8") as f:
+                assert f.read() == content
+            if remove_after_test:
+                os.remove(join(destination, file_name))
+                if destination not in current_or_empty_dirs:
+                    os.rmdir(destination)
```
vilit1, line 38: Add a check that the error did actually get raised and caught in the exception - 
`assert error is None`

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,426 @@
...
+                    target_directory=device_cert_output_directory,
+                    tarfile_path=bundle_output_directory,
+                    tarfile_name=device_id,
+                    overwrite=True,
+                )
+                # delete uncompressed files
+                rmtree(device_cert_output_directory)
+
+        # Give device registry a chance to catch up
+        sleep(1)
```
vilit1: this may not be enough - we should have retries with sleeps or have a better error so users wont be confused (and understand that the devices got created but the parent child relationships were not:
image

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,426 @@
...
+
+        # Set parent / child relationships
+        device_to_parent_iterator = (
+            tqdm(device_to_parent_dict, desc="Setting device parents")
+            if visualize
+            else device_to_parent_dict
+        )
+        for device_id in device_to_parent_iterator:
+            # get device properties
+            device = self.service_sdk.devices.get_identity(id=device_id)
```
vilit1, line 407: why not use existing commands to set device parents?

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,426 @@
...
+                    id=device_id, content=deployment_content
+                )
+
+        # Print device bundle details after other visuals
+        if bundle_output_directory:
+            num_bundles = len(config.devices)
+            bundle_plural = '' if num_bundles == 1 else 's'
+            print(f"{num_bundles} device bundle{bundle_plural} created in folder: {abspath(bundle_output_directory)}")
+
+    # TODO - Unit test
```
vilit1: todo

azext_iot/iothub/providers/helpers/edge_device_config.py
digimaun: As we've discussed in the past (at length), in general I think device side configuration scripts on a service side focused client tool (meant to be integrated) is an odd fit, but as this is the path we are experimenting down...

There is convenience organizing all the edge device side elements in a single module. Focusing on the scripts, I see that wrapping scripts as strings makes them easier to use in generating device bundles, but there are trade-offs as you say like formatting, linting and I'm thinking even vulnerability vectors may be harder to detect.

As a data point, we have prior art in picking up non-python code/json schema assets as part of the packaging configuration  (though I think maintaining a python dict representing the schema in code is OK) and they being used for edge deployment validation but

Assuming these scripts are already vetted and used elsewhere, I would re-iterate this changeset is classified as experimental and will be released as such, where things are dynamic and there could be significant changes in functionality/implementation. So in conclusion given the vetted scripts assumption I don't feel strongly on shifting the pattern in this change set.

azext_iot/iothub/_help.py
vilit1: Why is "Create edge devices in an IoT Hub" not good? Otherwise collection would be ok or even multiple devices. Group can be confused with device groups

azext_iot/iothub/_help.py
vilit1: talked about this - the solution is better help

azext_iot/iothub/params.py
vilit1: I see that if devices and config file are present then the command will error.

azext_iot/iothub/params.py
vilit1: as long as a clueless person can figure out the typing I am good

azext_iot/iothub/_help.py
vilit1: certs are confusing

azext_iot/iothub/_help.py
vilit1: actually the shell will prob handle it and it should be pretty obvious if we get an issue about this. This can be ignored then

azext_iot/iothub/_help.py
vilit1: Why is "Create edge devices in an IoT Hub" not good? Otherwise collection would be ok or even multiple devices (instead of a bunch of). Group can be confused with device groups...

azext_iot/common/utility.py
vilit1: it is two things:
1. the lack of capitalization 
2. the error messages are repeats

I can also question why YAML and then get the wrath of people who like it more than json

azext_iot/tests/iothub/devices/test_iot_edge_devices_create_int.py
```diff
@@ -0,0 +1,327 @@
...
+        # file has cert auth, call with overrides: keyAuth, custom config path, default edge agent
+        self.cmd(
+            f"iot edge devices create -n {self.entity_name} -g {self.entity_rg} -c --cfg {config_path} --dt {config_template} "
+            f"--out device_bundles --device-auth {override_auth_type} --default-edge-agent {default_edge_agent}"
+        )
+
+        self._validate_results(devices, "device_bundles", cert_auth=False, custom_device_template=True)
+
+    def _validate_results(self, devices, output_path, cert_auth=False, custom_device_template=False):
+        # device_tuple[0] == device_id
```
vilit1: im surprised that you didnt use a named tuple here

azext_iot/tests/iothub/devices/test_iot_edge_devices_create_unit.py
```diff
@@ -0,0 +1,1147 @@
...
+        )
+
+        if output:
+            assert exists(output)
+            for device in devices:
+                device_id = device[0].split("=")[1]
+                assert exists(join(output, f"{device_id}.tgz"))
+
+            rmtree(output)
+        pass
```
vilit1: pass

vilit1 (approved):

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,426 @@
...
+                else existing_device_ids
+            )
+            self.delete_device_identities(delete_iterator, confirm=True)
+        else:
+            # If not cleaning the hub, ensure no duplicate device ids
+            duplicates = list(
+                filter(lambda id: id in device_config_dict, existing_device_ids)
+            )
+            if any(duplicates):
+                raise InvalidArgumentValueError(
```
vilit1, line 234: not sure if there is a test for this yet

azext_iot/iothub/providers/device_identity.py
c-ryan-k: At the time (before I was doing a separate GET for the device identity before the PUT), I already had the entire device object constructed and there was no need to waste queries / throttle limits on extra calls (`set-parent` does a GET for both devices before the PUT on the new device property) - at this point it's probably a wash, though I think the goal with this provider is to eventually move over all of the `hub.py` device-identity commands into it, so eventually this will be able to make use of more provider/class methods

azext_iot/common/certops.py
c-ryan-k: I've added docstrings to the certificate commands. For the time being, my recommendation would be to leave these as separate calls, as we may iterate on some of the Edge device certificate generation, and until we know we have one function to rule them all (certs) I think we should leave them separate just in case.

azext_iot/common/certops.py
c-ryan-k: I don't think so. I think that's best left to the users to create their own certs / keys with passwords, and then they can BYOK with the `--root-cert`, `--root-key`, and `root-pass` parameters.

I think it's a possibility and a nice-to-have, but the way the logic works now, we'd end up overloading `--root-pass` with a "set root password" if creating a self-signed cert, or "use root password" if using an existing protected key.

azext_iot/common/certops.py
c-ryan-k: Yeah, this is pretty much just a helper method for `"".join(any_string_arr) -> write_content_to_file` - it just so happened it was explicitly useful for writing certificate files so it takes a `list[cert_str]` and writes them all to a single file.

setup.py
c-ryan-k: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7095&view=logs&j=6b37a07b-ae07-5f7f-46c3-5ce679ba41f2&t=6495c3e0-3a6f-5c64-f017-67dfb92d7cfa

vilit1 (approved):

azext_iot/common/certops.py
```diff
@@ -122,3 +132,252 @@ def open_certificate(certificate_path: str) -> str:
             certificate = base64.b64encode(certificate).decode("utf-8")
     # Remove trailing white space from the certificate content
     return certificate.rstrip()
+
+
+def create_v3_self_signed_root_certificate(
+    subject: str = "Azure_IoT_CLI_Extension_Cert",
```
digimaun: Are there consolidation opportunities between this function and other cert creation functions in this module? This function supports v3 extensions while there are more parameters/options for create_self_signed_certificate. There is also the currently named create_ca_signed_certificate below.

If there is some new extension or cert configuration down the line, does that necessitate a new create cert function?

Basically looking to reduce the need of new certificate creation functions in favor of less, but flexible ones with common code if possible.

azext_iot/common/certops.py
```diff
@@ -122,3 +132,252 @@ def open_certificate(certificate_path: str) -> str:
...
+
+
+def create_ca_signed_certificate(
+    subject: str,
+    ca_public: str,
+    ca_private: str,
+    cert_output_dir: Optional[str] = None,
+    cert_file: Optional[str] = None,
+    key_size: int = 4096,
+    valid_days: Optional[int] = 365,
```
digimaun: Is `valid_days` really Optional? What happens if you pass in `None`?

azext_iot/common/certops.py
```diff
@@ -122,3 +132,252 @@ def open_certificate(certificate_path: str) -> str:
...
+    return {
+        "certificate": certificate,
+        "privateKey": private_key,
+        "thumbprint": thumbprint,
+    }
+
+
+def create_ca_signed_certificate(
+    subject: str,
+    ca_public: str,
```
digimaun: nit, since description states ca_public / ca_private are expecting keys adding _key to the names makes the parameters more self describing.

azext_iot/common/fileops.py
```diff
@@ -0,0 +1,47 @@
+from os import makedirs, remove, listdir
```
digimaun, line 7: module is missing headers (coding utf8 (less important for py3 but lets be consistent & explicit) and license.

digimaun (commented):
Chunk of review

azext_iot/common/certops.py
```diff
@@ -122,3 +132,252 @@ def open_certificate(certificate_path: str) -> str:
...
+    ).decode("utf-8")
+
+    return {
+        "certificate": certificate,
+        "privateKey": private_key,
+        "thumbprint": thumbprint,
+    }
+
+
+def create_ca_signed_certificate(
```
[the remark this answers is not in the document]
digimaun: It makes sense.

azext_iot/common/certops.py
c-ryan-k: There certainly are.

In developing the v3 extension version, since it was so specifically tailored for edge, I wanted to avoid stepping on existing certificate code until it was finalized. Now that we have it successfully validated for Edge / module certificate communication, there's definitely an opportunity to consolidate.

I'll spike on this for a bit - For the initial self-signed cert, either a way for us to pass extra properties (v3 extensions, etc) into the existing `create_self_signed_certificate` or possibly seeing if we can just apply the v3 extension configuration to existing certificates and make sure that doesn't break anything.

I'll see if that leads to any insight on how to consolidate the signing / cert chaining process for edge certs too

azext_iot/iothub/_help.py
```diff
@@ -316,6 +316,53 @@ def load_iothub_help():
...
+        short-summary: Commands to manage IoT Edge devices.
+    """
+
+    helps["iot edge devices create"] = """
+        type: command
+        short-summary: Create multiple edge devices in an IoT Hub.
+        long-summary: |
+          This operation accepts inline device arguments or an edge devices configuration file in YAML or JSON format.
+          Inline command args (like '--device-auth') will take precedence and override configuration file properties if they are provided.
+          A sample configuration file can be found here: https://github.com/Azure/azure-iot-cli-extension/tree/dev/docs/samples/sample_devices_config.yaml
```
digimaun: nit/side thought, should we use an aka.ms link for the sample? 

Unrelated to this PR I want to use an aka.ms link initially pointing to our wiki dedicated page for inline json help for json heavy commands.

c-ryan-k: Yeah, this absolutely should be an aka.ms link - not only can we then update URLs if anything changes (branch/folder structure, filename) without updating help docs, but it's also shorter.

azext_iot/iothub/_help.py
```diff
@@ -316,6 +316,53 @@ def load_iothub_help():
...
     """
 
+    helps["iot edge devices"] = """
+        type: group
+        short-summary: Commands to manage IoT Edge devices.
+    """
+
+    helps["iot edge devices create"] = """
+        type: command
+        short-summary: Create multiple edge devices in an IoT Hub.
```
digimaun: nit, "Create **and configure**  multiple edge devices in an IoT Hub."

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,437 @@
...
+
+        # Query existing devices
+        query_args = ["SELECT deviceId FROM devices"]
+        query_method = self.service_sdk.query.get_twins
+        existing_devices = _execute_query(query_args, query_method)
+        existing_device_ids = [x["deviceId"] for x in existing_devices]
+
+        # Clear devices if necessary
+        if clean and len(existing_device_ids):
+            if not yes and not prompt_y_n(msg=f"Confirm you want to delete all devices in '{self.hub_name}'", default='y'):
```
digimaun: Since this is destructive, should the default be no/n instead?

azext_iot/iothub/commands_device_identity.py
```diff
@@ -0,0 +1,75 @@
...
+        root_key_path=root_key_path,
+        root_cert_password=root_cert_password,
+        output_path=bundle_output_path,
+    )
+
+
+def iot_delete_devices(
+    cmd,
+    device_ids: List[str],
+    confirm: bool = True,
```
digimaun: This is not command publicly exposed right? Looking at confirm with default value of `True` here.

c-ryan-k: Oh, that's right. When I was initially creating the method (the internal provider method `delete_device_identities`), instead of `confirm` being used as a way to let the user confirm the deletion, this was an option on the method on whether we should confirm that all devices *were* in fact deleted.

Since it's not publicly exposed (yet) - I don't think this will cause parameter confusion with our other `confirm` options, but I can at least make the parameter in the internal method more explicit and remove it from the provider-facing entrypoint.

azext_iot/iothub/params.py
```diff
@@ -249,6 +249,98 @@ def load_iothub_arguments(self, _):
...
+            options_list=["--visualize", "--vis", "-v"],
+            arg_type=get_three_state_flag(),
+            help="Shows visualizations of devices and progress of various tasks "
+            "(device creation, setting parents, updating configs, etc).",
+        )
+        context.argument(
+            "config_file",
+            options_list=["--config-file", "--config", "--cfg"],
+            help="Path to devices configuration file. Sample configuration file: "
+            "https://github.com/Azure/azure-iot-cli-extension/tree/dev/docs/samples/sample_devices_config.yaml",
```
digimaun: Another spot to replace with aka.ms link

azext_iot/iothub/params.py
```diff
@@ -249,6 +249,98 @@ def load_iothub_arguments(self, _):
...
+            "device_auth_type",
+            arg_type=get_enum_type(
+                [DeviceAuthType.shared_private_key.value, DeviceAuthType.x509_thumbprint.value]
+            ),
+            options_list=["--device-auth-type", "--device-auth"],
+            help="Device to hub authorization mechanism.",
+        )
+        context.argument(
+            "default_edge_agent",
+            options_list=["--default-edge-agent", "--default-agent", "--dea"],
```
digimaun, line 294: `--dea` :smile:

azext_iot/iothub/params.py
```diff
@@ -249,6 +249,98 @@ def load_iothub_arguments(self, _):
...
+            help="Default edge agent for created Edge devices if not specified individually.",
+        )
+        context.argument(
+            "device_config_template",
+            options_list=["--device-template", "--dt"],
```
digimaun: Just an observation that `--device-template`  is somewhat overloaded,  in that "device templates" are used in central [1](https://learn.microsoft.com/en-us/azure/iot-central/core/concepts-device-templates) [2](https://github.com/Azure/azure-iot-cli-extension/issues/364), DTDL (where it transitioned to `model` as the preferred term) and ...

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -0,0 +1,579 @@
...
+    # Dynamic, AlwaysOnStartup, OnErrorOnly
+    device_toml["auto_reprovisioning_mode"] = "Dynamic"
```
digimaun, line 262: Are AlwaysOnStartup and OnErrorOnly not settable?

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -0,0 +1,579 @@
...
+            hostname=hostname,
+            parent_hostname=parent_hostname,
+            edge_agent=edge_agent,
+            container_auth=container_auth,
+        )
+        config.devices.append(device_config)
+    return config
+
+
+def assemble_nargs_to_dict(hash_list: List[str]) -> Dict[str, str]:
```
digimaun: If this is exactly the same as the classmethod in ..../deviceupdate/providers/base (looks like it is) we can plug it in common/utility (TBD future reorganization) and reference it from both places

azext_iot/iothub/providers/device_identity.py
```diff
@@ -0,0 +1,437 @@
...
+
+    def delete_device_identities(self, device_ids: List[str], confirm: bool = False):
+        for id in device_ids:
+            try:
+                self.service_sdk.devices.delete_identity(id=id, if_match="*")
+            except Exception as err:
+                raise AzureResponseError(err)
+        if confirm:
+            existing_devices = self.service_sdk.devices.get_devices()
+            if len(existing_devices):
```
digimaun: minor thing - `list` with an item is True (and empty list is False) so you could do `if existing_devices:`

azext_iot/tests/utility/test_iot_file_operations.py
```diff
@@ -0,0 +1,59 @@
+from azext_iot.common.fileops import tar_directory, write_content_to_file
```
digimaun, line 7: Missing headers, I'm thinking we should add a module linter job to help with this kind of thing

setup.py
```diff
@@ -53,7 +53,10 @@
     "jsonschema~=3.2.0",
     "importlib_metadata;python_version<'3.8'",
     "azure-iot-device~=2.11",
+    "tomli",
```
digimaun: It's good practice to constrain dependency so we stay within the major version range we tested with. The concern is a new major version makes a breaking change.

digimaun (approved):
Nice work and solid tests 

Lets make sure to show evidence of full test pass prior to merge and capture follow ups on our backlog

azext_iot/iothub/commands_device_identity.py
digimaun: Understood

azext_iot/iothub/providers/helpers/edge_device_config.py
c-ryan-k: That was my understanding (at least of how the existing tool functions) after a few sessions with the edge team to debug the resultant config.toml / E2E device deployments - the comments here are just a note of the allowed model values from Edge in Rust. I can connect with them to see if it makes sense to make this a parameterized value, or if we should move this `set` to just the default config template and let an existing TOML file use its own value.

digimaun: Ok gotcha...conceptually (abstract from this particular comment) the more settings we hard code / remove the customization from users the more specificity of scenario this tooling solution becomes. Not sure how important customization of this auto reprovisioning mode setting is.

c-ryan-k:
First round of tests - have one final one (hopefully) coming that updates unit and integration tests so I'll have another run starting shortly.

https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7310&view=results

c-ryan-k:
Ok, [one last full pipeline run to go](https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7315&view=results) once the PR pipeline passes, but here's some local testing evidence:
image
image

azext_iot/iothub/providers/helpers/edge_device_config.py
c-ryan-k: Have a note to follow up in a subsequent PR - I'll change this to:
-  Only set `Dynamic` in our default loaded config, or if there is no value in a supplied user template
-  Check supplied template values to ensure this value is valid
-  Add tests to confirm

azext_iot/iothub/providers/device_identity.py
c-ryan-k: Have a note to address this in a follow-up PR

azext_iot/tests/utility/test_iot_file_operations.py
c-ryan-k: Will address any gaps in test coverage (by lines etc) in a follow-up PR

azext_iot/iothub/commands_device_identity.py
c-ryan-k: Have a note to follow-up on the implementation of this command to allow users to delete multiple device-identities at once.

c-ryan-k:
[Looks like tests were mostly successful](https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7315&view=ms.vss-test-web.build-test-results-tab), save some of our recent recurring issues. I'm good with this.

digimaun (approved):
LGTM

PR #605, opened 2022-12-15
Update help text for adt invoke command example according to sample DTDL

Elsie4ever:

### General Guidelines

Intent for Production

Basic expectations

This pr address this issue[issue](https://github.com/Azure/azure-cli/issues/18461), according to the [DTDL sample](https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json), the payload passed in successfully when it's only "5" instead of '{"delay":5}'

### DIFF
--- a/azext_iot/iothub/_help.py
+++ b/azext_iot/iothub/_help.py
@@ -107,7 +107,7 @@ def load_iothub_help():
           text: >
             az iot hub digital-twin invoke-command --command-name {command_name} -n {iothub_name} -d {device_id} --payload '{"property_name": "property_value"}'
 
-        - name: Invoke root level command "reboot" which takes a payload named "delay" conforming to DTDL model:
+        - name: Invoke root level command "reboot" which takes a payload named "delay" conforming to DTDL model -
           https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json.
           text: >
             az iot hub digital-twin invoke-command --command-name reboot -n {iothub_name} -d {device_id} --payload 5
--- a/azext_iot/iothub/_help.py
+++ b/azext_iot/iothub/_help.py
@@ -107,8 +107,9 @@ def load_iothub_help():
           text: >
             az iot hub digital-twin invoke-command --command-name {command_name} -n {iothub_name} -d {device_id} --payload '{"property_name": "property_value"}'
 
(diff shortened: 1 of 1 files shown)


azext_iot/iothub/_help.py
```diff
@@ -105,7 +105,7 @@ def load_iothub_help():
         examples:
         - name: Invoke root level command "reboot" which takes a payload that includes the "delay" property.
           text: >
-            az iot hub digital-twin invoke-command --command-name reboot -n {iothub_name} -d {device_id} --payload '{"delay":5}'
+            az iot hub digital-twin invoke-command --command-name reboot -n {iothub_name} -d {device_id} --payload "5"
```
digimaun: We should make sure this works with a PnP device client sample

Elsie4ever: thanks will test it out!

Elsie4ever: @digimaun for PnP device client sample do you mean device defined using something like this https://github.com/Azure/opendigitaltwins-dtdl/blob/master/DTDL/v2/samples/TemperatureController.json? Just to confirm if I'm verify using the right way. I firstly used `az iot device simulate -n dfsdf -g jiacjutest -d device1 --model-id 'dtmi:com:example:TemperatureController;1'` to simulate the device using corresponding model id, and then send the command using `az iot hub digital-twin invoke-command --command-name reboot -n dfsdf -d device1 --payload "5"`, and the response I got is:
{
  "payload": {
    "methodName": "reboot",
    "methodRequestId": "1",
    "methodRequestPayload": 5
  },
  "status": "200"
}

would this wipe your concern?😋

vilit1: We discussed this on friday and the result is to have two examples, one with the older json and another with the new one with the help text explaining what the difference is between the two (as in what models would work with them)

Elsie4ever: @vilit1 example updated!😀

YingXue: According to the TemperatureController.json dtdl. delay is an integer. we should be sending 5 directly instead of a string.

YingXue (approved):

azext_iot/iothub/_help.py
Elsie4ever: resolved offline

PR #606, opened 2022-12-19
ADT UAI Extra Updates

vilit1:
Goals of this pr:
1. add  UAI for data history
2. fix some nit wordings in help from previous UAI pr
3. try to make the deletion of UAI's smoother by disassociating them from the dt's before deleting them

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/digitaltwins/providers/connection/builders.py
+++ b/azext_iot/digitaltwins/providers/connection/builders.py
@@ -63,9 +63,11 @@ class AdxConnectionValidator(object):
         if self.dt.identity is None and identity == SYSTEM_IDENTITY:
             raise BadRequestError(DT_SYS_IDENTITY_ERROR)
         elif (
-            self.dt.identity is None
-            or identity not in self.dt.identity.user_assigned_identities
-            and identity != SYSTEM_IDENTITY
+            identity != SYSTEM_IDENTITY and (
+                self.dt.identity is None
+                or self.dt.identity.user_assigned_identities is None
+                or identity not in self.dt.identity.user_assigned_identities
+            )
         ):
             raise BadRequestError(DT_UAI_IDENTITY_ERROR)
         # set the identity to be principal id for ease
--- a/azext_iot/digitaltwins/providers/connection/builders.py
+++ b/azext_iot/digitaltwins/providers/connection/builders.py
@@ -60,14 +60,15 @@ class AdxConnectionValidator(object):
         self.yes = yes
         self.dt = dt_instance
         # Check that the identity is associated with the dt
-        if self.dt.identity is None and identity == SYSTEM_IDENTITY:
+        if identity == SYSTEM_IDENTITY and (
+            not self.dt.identity
+            or not self.dt.identity.principal_id
+        ):
             raise BadRequestError(DT_SYS_IDENTITY_ERROR)
-        elif (
-            identity != SYSTEM_IDENTITY and (
-                self.dt.identity is None
-                or self.dt.identity.user_assigned_identities is None
-                or identity not in self.dt.identity.user_assigned_identities
-            )
+        elif identity != SYSTEM_IDENTITY and (
+            not self.dt.identity
+            or not self.dt.identity.user_assigned_identities
+            or identity not in self.dt.identity.user_assigned_identities
         ):
             raise BadRequestError(DT_UAI_IDENTITY_ERROR)
         # set the identity to be principal id for ease
--- a/azext_iot/digitaltwins/providers/connection/builders.py
+++ b/azext_iot/digitaltwins/providers/connection/builders.py
@@ -5,18 +5,27 @@
 # --------------------------------------------------------------------------------------------
 
 import json
+from typing import Optional
 
 from azure.cli.core.azclierror import BadRequestError, CLIInternalError, ManualInterrupt
 from azext_iot.common.embedded_cli import EmbeddedCLI
-from azext_iot.sdk.digitaltwins.controlplane.models import AzureDataExplorerConnectionProperties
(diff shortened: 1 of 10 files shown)


vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7190&view=results

azext_iot/digitaltwins/providers/connection/builders.py
```diff
@@ -44,20 +53,37 @@ def __init__(
...
+        if self.dt.identity is None and identity == SYSTEM_IDENTITY:
+            raise BadRequestError(DT_SYS_IDENTITY_ERROR)
+        elif (
+            identity != SYSTEM_IDENTITY and (
+                self.dt.identity is None
+                or self.dt.identity.user_assigned_identities is None
+                or identity not in self.dt.identity.user_assigned_identities
+            )
+        ):
+            raise BadRequestError(DT_UAI_IDENTITY_ERROR)
```
c-ryan-k: nit - I think we can simply the logic here a bit with a single nested `if`:
Another nit/question - do we need to check `is None` or is `not self.dt.identity` sufficient?

```suggestion
        # Check that the identity is associated with the dt
        if not self.dt.identity:
            if identity == SYSTEM_IDENTITY:
                raise BadRequestError(DT_SYS_IDENTITY_ERROR)
            elif (
                not self.dt.identity.user_assigned_identities
                or identity not in self.dt.identity.user_assigned_identities
            ):
                raise BadRequestError(DT_UAI_IDENTITY_ERROR)
```

azext_iot/tests/digitaltwins/__init__.py
```diff
@@ -528,6 +528,15 @@ def delete_servicebus_resources(self):
     def delete_user_identity(self):
```
c-ryan-k, line 531: nit (I think) - I know this method is only called during teardown - but it seems strange for this to remove `all` user-assigned identities from all `dt` instances - do we need to specify a specific identity in the embedded call, or could we rename the function `delete_all_user_identities` or something?

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -76,11 +83,15 @@ def test_dt_data_history_adx(self):
         table_name = f"tb_{generate_generic_id()}"
         consumer_group = f"cg-{generate_generic_id()}"
         self.add_eventhub_consumer_group(consumer_group=consumer_group)
+        user_identity = self.ensure_user_identity()["id"]
+        # TODO: lower sleep time to necessary amount
+        sleep(50)
```
c-ryan-k: Is this still a `TODO`? Is there anything we could loop / poll for?

c-ryan-k (commented):
Do we need a history update for this? It looks like we added at least one param (user_identity)

azext_iot/tests/digitaltwins/__init__.py
vilit1: right now, I am only making one user identity (and thus deleting one user identity) and the tests that use user identities also only create one dt instance

the only reason for the for loop to remove it from all dt instances is for future proofing if there is a test that makes multiple dt instances.

Since there is only one user identity, saying `delete_all_user_identities` is awkward. I don't think there will be a test scenario with multiple user identities in the near future either.

vilit1:
> Do we need a history update for this? It looks like we added at least one param (user_identity)

I will add one when it is closer to actually being merged in (I hate merge conflicts with history)

Here is the blurb that I will put:

```
**Digital Twins**

 * Addition of User Assigned Identities for data history connections. The command `az dt data-history connection create adx` now can take an extra parameter `--mi-user-assigned` to use an associated User Assigned Identity for the connection creation rather than the system assigned identity for the Digital Twin.
 * Some minor improvements to command documentation involving managed identities.
```

azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
vilit1: This is a copy paste for having the User Assigned Identity be fully created (and associated in the graph)... Sadly I have no idea how to check if the UAI has been fully created aside from trying to use it in another command (and have the command fail). 50 seconds is the sweet spot I hit for previous UAI tests

None of the fields suggest anything I can poll for: 
image


I kept the TODO so it is easier to find if I think of a better way for the UAI tests

Ill move it to the method that creates the UAI but I will keep the todo as a reminder for myself and/or anyone else

azext_iot/digitaltwins/providers/connection/builders.py
vilit1: Your if will not work... but I can change the `is None` to just having a `not`

This logic takes me a bit to process every time...

1. if there is no dt.identity, then you cannot do anything
  a. if you are trying for system, show the system error
  b. if you are trying for user, show the user error
2. if there is an identity, and you are trying for user, you need to check if there are any user identities associated with the dt and the user identity asked for is there too
  a. if there is no user identities or the user identity asked for isnt there, show the user error

The logic above will result in having `raise BadRequestError(DT_UAI_IDENTITY_ERROR)` in two places,  so I reworked that so it would be there in one place (aka current code). If you think this logic is more reasonable, I can change the code back to it (we will have to deal with the same error raised in two places next to each other)

c-ryan-k: You're correct, I missed one step of the `self.dt.identity` check that would end up resulting in managing two separate exception cases 👍

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7211&view=results

c-ryan-k (approved):

PR #609, opened 2022-12-28
Created test queries command for hub configuration

Elsie4ever:
Created test queries command for hub configuration, it help to validate target condition and custom metric queries

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -21,6 +21,11 @@ unreleased
   
   To learn more about this transition, visit http://aka.ms/iot-ca-updates/.
 
+* Added test queries commands for configuration, it validates target condition and custom metric queries for a
+  configuration on the IoT Hub:
+
+    - az iot hub configuration test-queries
+
 
 **IoT Central updates**
 
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -5,6 +5,12 @@ Release History
 
 unreleased
 +++++++++++++++
+**IoT Hub updates**
+
+* Added test queries commands for configuration, it validates target condition and custom metric queries for a
+  configuration on the IoT Hub:
+
+    - az iot hub configuration test-queries
 
 0.18.3
 +++++++++++++++
@@ -20,12 +26,6 @@ unreleased
   - az iot hub certificate root-authority set
   
   To learn more about this transition, visit http://aka.ms/iot-ca-updates/.
-
-* Added test queries commands for configuration, it validates target condition and custom metric queries for a
-  configuration on the IoT Hub:
-
-    - az iot hub configuration test-queries
-
 
 **IoT Central updates**
 
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -829,16 +829,15 @@ helps[
       This command supports both inline json and file input for custom metric queries `--custom-metric-queries`.
       Review examples and parameter description for details on how to fully utilize the operation.
 
-      - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
-      - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
-      - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
-      - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
+      Read more about using quotation marks and escapes characters in different shells here:
+      https://github.com/Azure/azure-iot-cli-extension/wiki/Tips#use-quotation-mark-properly-for-command-continuation
 
     examples:
     - name: Validate the target condition and custom metric queries with inline json.
       text: >
         az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
-        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''""}'
+        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''"",
+        \\"mertic\\":\\"select *\\"}'
     - name: Validate the target condition and custom metric queries with json file.
       text: >
         az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
--- a/azext_iot/_help.py
+++ b/azext_iot/_help.py
@@ -821,6 +821,31 @@ helps[
 """
 
 helps[
+    "iot hub configuration test-queries"
+] = """
+    type: command
+    short-summary: Validates target condition and custom metric queries for a configuration on the IoT Hub
+    long-summary: |
+      This command supports both inline json and file input for custom metric queries `--custom-metric-queries`.
+      Review examples and parameter description for details on how to fully utilize the operation.
+
+      - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
+      - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
+      - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
+      - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
+
+    examples:
+    - name: Validate the target condition and custom metric queries with inline json.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''""}'
+    - name: Validate the target condition and custom metric queries with json file.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '/path/to/file'
+"""
+
+helps[
     "iot hub distributed-tracing"
 ] = """
     type: group
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -819,6 +819,13 @@ def load_arguments(self, _):
             type=int,
             help="Maximum number of configurations to return. By default all configurations are returned.",
         )
+        context.argument(
+            "custom_metric_queries",
+            options_list=["--custom-metric-queries"],
+            type=dict,
+            help="Map of custrom metric queries for configuration. "
+            'Format example: {"mertic1":"select *", "metric2":"select moduleId from devices.modules where tags.location=''US''"}',
+        )
 
     with self.argument_context("iot edge") as context:
         context.argument(
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -822,9 +822,7 @@ def load_arguments(self, _):
         context.argument(
             "custom_metric_queries",
             options_list=["--custom-metric-queries"],
-            type=dict,
-            help="Map of custrom metric queries for configuration. "
-            'Format example: {"mertic1":"select *", "metric2":"select moduleId from devices.modules where tags.location=''US''"}',
+            help="Inline custom metric queries JSON or file path to custom metric queries JSON. "
         )
 
     with self.argument_context("iot edge") as context:
--- a/azext_iot/deviceupdate/_help.py
+++ b/azext_iot/deviceupdate/_help.py
@@ -842,13 +842,8 @@ def load_deviceupdate_help():
           `--related-file`. Review examples and parameter descriptions for details on how
           to fully utilize the operation.
 
-          - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
-          - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
-          - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
-          - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
-
-          For a detailed explanation of shell quoting rules please goto
-            https://learn.microsoft.com/en-us/cli/azure/use-cli-effectively
+          Read more about using quotation marks and escapes characters in different shells here
+            https://github.com/Azure/azure-iot-cli-extension/wiki/Tips#use-quotation-mark-properly-for-command-continuation
 
         examples:
         - name: Initialize a minimum content import manifest. Inline json optimized for `bash`.
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -49,6 +49,7 @@ from azext_iot._factory import SdkResolver, CloudError
 from azext_iot.operations.generic import _execute_query, _process_top
 from typing import Optional
 import pprint
+from azext_iot.common.utility import validate_key_value_pairs
 
 logger = get_logger(__name__)
 printer = pprint.PrettyPrinter(indent=2)
@@ -1641,8 +1642,13 @@ def iot_hub_configuration_test_queries(
     try:
         headers = {}
         headers["If-Match"] = '"{}"'.format(etag if etag else "*")
+        if custom_metric_queries:
+            custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="content")
         result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries)
-        print(result.target_condition_error)
+        if result.target_condition_error is not None:
+            logger.error('Target condition validation failed: "%s".', result.target_condition_error)
+        if result.custom_metric_query_errors is not None:
+            logger.error('Target condition validation failed: "%s".', result.custom_metric_query_errors)
     except CloudError as e:
         handle_service_exception(e)
 
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -1625,7 +1625,7 @@ def iot_hub_configuration_test_queries(
     hub_name=None,
     resource_group_name=None,
     login=None,
-    auth_type_dataplane=None,
+    auth_type_dataplane=None
 ):
     discovery = IotHubDiscovery(cmd)
     target = discovery.get_target(
@@ -1641,14 +1641,17 @@ def iot_hub_configuration_test_queries(
         if custom_metric_queries:
             custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="custom_metric_queries")
 
-        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries)
+        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries, raw=True).response.json()
+        validation_errors = []
 
-        if not result.target_condition_error and not result.custom_metric_query_errors:
+        if not result['targetConditionError'] and not result['customMetricQueryErrors']:
             return 'Validation passed!'
-        if result.target_condition_error:
-            return 'Target condition validation failed: {}.'.format(result.target_condition_error)
-        if result.custom_metric_query_errors:
-            return 'Custom metric query validation failed: {}.'.format(result.custom_metric_query_errors)
+        if result['targetConditionError']:
+            validation_errors.append('Target condition validation failed: {}.'.format(result['targetConditionError']))
+        if result['customMetricQueryErrors']:
+            validation_errors.append('Custom metric query validation failed: {}.'.format(result['customMetricQueryErrors']))
+
+        return validation_errors
     except CloudError as e:
         handle_service_exception(e)
 
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -49,7 +49,6 @@ from azext_iot._factory import SdkResolver, CloudError
 from azext_iot.operations.generic import _execute_query, _process_top
 from typing import Optional
 import pprint
-from azext_iot.common.utility import validate_key_value_pairs
 
 logger = get_logger(__name__)
 printer = pprint.PrettyPrinter(indent=2)
@@ -1627,7 +1626,6 @@ def iot_hub_configuration_test_queries(
     resource_group_name=None,
     login=None,
     auth_type_dataplane=None,
-    etag=None,
 ):
     discovery = IotHubDiscovery(cmd)
     target = discovery.get_target(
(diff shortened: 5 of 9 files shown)


HISTORY.rst
```diff
@@ -21,6 +21,11 @@ unreleased
   
   To learn more about this transition, visit http://aka.ms/iot-ca-updates/.
 
+* Added test queries commands for configuration, it validates target condition and custom metric queries for a
```
vilit1: put this under the unreleased since it wasn't released yet

azext_iot/_help.py
```diff
@@ -820,6 +820,31 @@
...
+    type: command
+    short-summary: Validates target condition and custom metric queries for a configuration on the IoT Hub
+    long-summary: |
+      This command supports both inline json and file input for custom metric queries `--custom-metric-queries`.
+      Review examples and parameter description for details on how to fully utilize the operation.
+
+      - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
+      - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
+      - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
+      - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
```
vilit1: We should have this in a separate readme/file or wiki page that we link to the main page (and also any commands that we think needs it). 

Wanna write a "best practices for az cli beyond the given resources"? Let's start with adding this info (and any other weird quirks of the command shells) to the wiki page

Our wiki page to add to:
https://github.com/Azure/azure-iot-cli-extension/wiki/Tips

Leverage this doc (and or tell people to read this):
https://learn.microsoft.com/en-us/cli/azure/use-cli-effectively?tabs=bash%2Cbash2

vilit1: Also make sure you test these things out in a non powershell terminal - ex: ubuntu 

Google for how to add ubuntu to windows terminal
https://rkstrdee.medium.com/how-to-add-ubuntu-tab-to-windows-10s-new-terminal-271eb6dfd8ee

azext_iot/_help.py
```diff
@@ -820,6 +820,31 @@
...
+
+    examples:
+    - name: Validate the target condition and custom metric queries with inline json.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''""}'
+    - name: Validate the target condition and custom metric queries with json file.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '/path/to/file'
```
vilit1, line 844: Does the file have only one query?

azext_iot/_help.py
```diff
@@ -820,6 +820,31 @@
...
+      - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
+      - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
+      - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
+      - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
+
+    examples:
+    - name: Validate the target condition and custom metric queries with inline json.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''""}'
```
vilit1: You may need to also update the param file - have you tried running your command with -h to see how it looks?

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,42 @@ def iot_hub_configuration_delete(
...
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    try:
+        if custom_metric_queries:
+            custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="content")
+
+        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries)
+
+        if not result.target_condition_error and not result.custom_metric_query_errors:
+            print("Validation passed!")
```
vilit1: What does string/message does the service return if the queries pass? What does the portal use to say that the queries worked?

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,42 @@ def iot_hub_configuration_delete(
...
+        resource_group_name=resource_group_name,
+        login=login,
+        auth_type=auth_type_dataplane,
+    )
+    resolver = SdkResolver(target=target)
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    try:
+        if custom_metric_queries:
+            custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="content")
```
vilit1: argument name is custom metric queries here, no?

azext_iot/operations/hub.py
vilit1: Also you can return the message if it makes sense

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,42 @@ def iot_hub_configuration_delete(
...
+
+        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries)
+
+        if not result.target_condition_error and not result.custom_metric_query_errors:
+            print("Validation passed!")
+            return
+        if result.target_condition_error:
+            logger.error('Target condition validation failed: "%s".', result.target_condition_error)
```
vilit1: why log and not return?

azext_iot/tests/iothub/configurations/test_iot_config_int.py
```diff
@@ -676,6 +677,29 @@ def test_device_configurations(self):
...
+            # Validate target condition for configuration
+            target_condition = "from devices.modules where tags.building=9"
+            self.cmd(
+                self.set_cmd_auth_type(
+                    "iot hub configuration test-queries -n {} --target-condition \"{}\"".format(
+                        self.entity_name, target_condition
+                    ),
+                    auth_type=auth_phase,
+                ),
+                expect_failure=False,
```
vilit1: if expect_failure is false, you don't need to specify that input - magic of python and default values

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,64 @@ def test_config_apply_edge_error(
...
+        service_client.return_value = build_mock_response(mocker, request.param, {})
+        return service_client
+
+    @pytest.mark.parametrize(
+        "hub_name, target_condition, custom_metric_queries, etag",
+        [
+            (
+                mock_target["entity"],
+                "tags.building=43 and tags.environment='test'",
+                "{\"hi\": \"sd\"}",
```
vilit1: you can just do `'{"hi":"sd"}'` here

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,64 @@ def test_config_apply_edge_error(
...
+                [generate_generic_id(), None],
+            )
+        ],
+    )
+    def test_config_queries(self, fixture_cmd, serviceclient, hub_name, etag, target_condition, custom_metric_queries):
+        subject.iot_hub_configuration_test_queries(
+            cmd=fixture_cmd,
+            hub_name=hub_name,
```
vilit1: considering that the hub_name will be the same, you can set the variable here rather than parameterizing it

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,64 @@ def test_config_apply_edge_error(
...
+
+    @pytest.mark.parametrize(
+        "hub_name, target_condition, custom_metric_queries, etag",
+        [
+            (
+                mock_target["entity"],
+                "tags.building=43 and tags.environment='test'",
+                "{\"hi\": \"sd\"}",
+                [generate_generic_id(), None],
+            )
```
vilit1: more conditions? weird to parameterize with only one test case - same with the one below

azext_iot/_help.py
Elsie4ever: nope, you can have multiple queries, the input is a dict[str, str]

azext_iot/_help.py
Elsie4ever: yep I did run -h

azext_iot/operations/hub.py
Elsie4ever: this service call is a bit different, the return type is ConfigurationQueriesTestResponse. when it fails, it will not throw exception, the status code will be 200, so I need to check the value of target_condition_error  and custom_metric_query_errors. On the other hand, when validation passed, there is no message from service call, the target_condition_error  and custom_metric_query_errors will just be NONE. So i think it would be make sense for me to write a message telling user that the validation pass, otherwise nothing will be happened?😀

azext_iot/tests/iothub/configurations/test_iot_config_int.py
Elsie4ever: ah I see, will update!

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
Elsie4ever: I was referencing test_config_apply_edge_error that also only having one test case that is using parameterize, so what would be the best practice to initialize the parameters when there is only one case?

azext_iot/operations/hub.py
vilit1: do a return since it will be more useful for developers than a print (cause then they can do stuff like save the output vs having to save whatever prints to the console)

azext_iot/_help.py
vilit1: how does it work if there are multiple queries?

lets say that the first query passes, second fails, third passes - what would the result be?
what if all queries fail?

azext_iot/_params.py
```diff
@@ -819,6 +819,13 @@ def load_arguments(self, _):
             type=int,
             help="Maximum number of configurations to return. By default all configurations are returned.",
         )
+        context.argument(
+            "custom_metric_queries",
+            options_list=["--custom-metric-queries"],
+            type=dict,
+            help="Map of custrom metric queries for configuration. "
```
vilit1: custom

azext_iot/_params.py
```diff
@@ -819,6 +819,13 @@ def load_arguments(self, _):
...
+            'Format example: {"mertic1":"select *", "metric2":"select moduleId from devices.modules where tags.location=''US''"}',
```
vilit1: usually we have format examples in the help rather than the param help.

azext_iot/_help.py
vilit1: ^ on this note, add an example with multiple inline queries

azext_iot/_params.py
vilit1: also this can be a file too - use wording like this example for az dt model create 
image

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,41 @@ def iot_hub_configuration_delete(
...
+        if custom_metric_queries:
+            custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="custom_metric_queries")
+
+        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries)
+
+        if not result.target_condition_error and not result.custom_metric_query_errors:
+            return 'Validation passed!'
+        if result.target_condition_error:
+            return 'Target condition validation failed: {}.'.format(result.target_condition_error)
+        if result.custom_metric_query_errors:
```
vilit1: will there ever be a condition where we have result.target_condition_error and result.custom_metric_query_errors? We want to make sure we return everything if possible

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,40 @@ def test_config_apply_edge_error(
...
+    def serviceclient(self, mocker, fixture_ghcs, fixture_sas, request):
+        service_client = mocker.patch(path_service_client)
+        service_client.return_value = build_mock_response(mocker, request.param, {})
+        return service_client
+
+    def test_config_queries(self, fixture_cmd, serviceclient):
+        subject.iot_hub_configuration_test_queries(
+            cmd=fixture_cmd,
+            hub_name=mock_target["entity"],
+            target_condition="tags.building=43 and tags.environment='test'",
```
vilit1: you can keep the target_condition and custom metric queries as parameterized - 

I thought of some extra tests:
have some failure cases (you want to hit both if statements so you will need at least two)
have some that are multiple queries that pass
have some multiple queries that are a mixed of fail/pass

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,40 @@ def test_config_apply_edge_error(
...
+
+    def test_config_queries_invalid_args(
+        self, fixture_cmd, serviceclient
+    ):
+        with pytest.raises(CLIError):
+            subject.iot_hub_configuration_test_queries(
+                cmd=fixture_cmd,
+                hub_name=mock_target["entity"],
+                target_condition="targetCondition",
+                custom_metric_queries="hello.txt",
```
vilit1, line 1014: target_condition and custom metric queries can also be parameterized here

more tests - are there invalid target conditions?

what about invalid custom metric queries (lets say a bad json)

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -899,3 +899,40 @@ def test_config_apply_edge_error(
...
+        assert "/configurations/testQueries" in url
+        assert method == "POST"
+
+        assert body.get("targetCondition") == "tags.building=43 and tags.environment='test'"
+        assert body.get("customMetricQueries") == {'hi': 'sd'}
+
+    def test_config_queries_invalid_args(
+        self, fixture_cmd, serviceclient
+    ):
+        with pytest.raises(CLIError):
```
vilit1, line 1009: we should start checking what error message it throws too but that can be a future pr.

azext_iot/operations/hub.py
Elsie4ever: got it! now it's returned😁

azext_iot/_params.py
Elsie4ever: true! forgot about the file, thanks!

azext_iot/_params.py
Elsie4ever: ok got it, I was following the one above, will remove that

azext_iot/operations/hub.py
Elsie4ever: there will, let me make a change

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
Elsie4ever: test updated!

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
Elsie4ever: test update!

azext_iot/_help.py
Elsie4ever: updated and tested in multiple shells

azext_iot/_help.py
Elsie4ever: updated with multiple inline queries. the return type of query error is dict[str, str], therefore when we have multiple queries with mixed fail/pass, it will have error under the specific queries name

Elsie4ever: here's an example:

(env) PS D:\azure-iot-cli-extension> az iot hub configuration test-queries -n dfsdf --custom-metric-queries '{\"mymetric\":\"sd\", \"mymetric2\":\"select moduleId from devices.modules where tags.location=''US''\"}'
[
  "Custom metric query validation failed: {'mymetric': \"SQL query parse failed due to following errors: \\n\\tLine 1:0 missing SELECT at 'sd'\"}."
]

azext_iot/deviceupdate/_help.py
```diff
@@ -842,13 +842,8 @@ def load_deviceupdate_help():
...
 
-          - For bash inline json format use '{"key":"value"}' and \\ (backslash) for command continuation.
-          - For powershell inline json format use '{\\"key\\":\\"value\\"}' and ` (tilde) for command continuation.
-          - For cmd inline json format use \"{\\"key\\":\\"value\\"}\" and ^ (caret) for command continuation.
-          - For file based json input use "@/path/to/file". File based input avoids shell quotation issues.
-
-          For a detailed explanation of shell quoting rules please goto
-            https://learn.microsoft.com/en-us/cli/azure/use-cli-effectively
+          Read more about using quotation marks and escapes characters in different shells here
+            https://github.com/Azure/azure-iot-cli-extension/wiki/Tips#use-quotation-mark-properly-for-command-continuation
```
digimaun, line 846: We should create a new page in our wiki that is dedicated for inline json help - linking to that is safer. I'm hesitant about using an anchor where the header content could slightly change breaking the link.

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,44 @@ def iot_hub_configuration_delete(
...
+        auth_type=auth_type_dataplane,
+    )
+    resolver = SdkResolver(target=target)
+    service_sdk = resolver.get_sdk(SdkType.service_sdk)
+
+    try:
+        if custom_metric_queries:
+            custom_metric_queries = process_json_arg(custom_metric_queries, argument_name="custom_metric_queries")
+
+        result = service_sdk.configuration.test_queries(target_condition, custom_metric_queries, raw=True).response.json()
```
digimaun, line 1644: What happens when target_condition is None?

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,44 @@ def iot_hub_configuration_delete(
...
+        validation_errors = []
+
+        if not result['targetConditionError'] and not result['customMetricQueryErrors']:
+            return 'Validation passed!'
```
digimaun, line 1648: In most cases we should not return bare strings in stdout, this can be a debug or info logger item if desired. The result code being 0 indicates success.

azext_iot/operations/hub.py
```diff
@@ -1618,6 +1618,44 @@ def iot_hub_configuration_delete(
...
+        if result['targetConditionError']:
+            validation_errors.append('Target condition validation failed: {}.'.format(result['targetConditionError']))
+        if result['customMetricQueryErrors']:
```
digimaun, line 1651: Is there something non-optimal or confusing in regards to the data model returned by the service where we need to organize it in a different way?

azext_iot/_help.py
```diff
@@ -820,6 +820,30 @@
...
+      Review examples and parameter description for details on how to fully utilize the operation.
+
+      Read more about using quotation marks and escapes characters in different shells here:
+      https://github.com/Azure/azure-iot-cli-extension/wiki/Tips#use-quotation-mark-properly-for-command-continuation
+
+    examples:
+    - name: Validate the target condition and custom metric queries with inline json.
+      text: >
+        az iot hub configuration test-queries -n {iothub_name} --target-condition "from devices.modules where tags.building=9"
+        --custom-metric-queries '{\\"mymetric\\":""select moduleId from devices.modules where tags.location=''US''"",
```
digimaun, line 839: json input by file is fine, but have we thought about alternative to inline json ways to take user input to reduce the users need for dealing with shell quotation issues? I know the edge deployment create command has a similar pattern to this (whose implementation is legacy) , but I'm thinking of alternatives like an nargs kvp style input, something like:

```
az iot hub configuration test-queries \
-n myiothub \
--target-condition "from devices.modules where tags.building=9" \
--custom-metric-queries mymetric1="select moduleId from devices.modules where tags.location='US'" mymetric2="select moduleId from devices.modules where tags.location='EU'" 
```

Thoughts?

azext_iot/deviceupdate/_help.py
Elsie4ever: sure, I will move it to a new page!

azext_iot/operations/hub.py
Elsie4ever: when target_condition is None, it will go ahead validate custom_metric_queries; if custom_metric_queries is also None, then the return will be {'Message': 'ErrorCode:ArgumentInvalid;BadRequest', 'ExceptionMessage': 'Tracking ID:cffd76d19a36464ca394469eda49b357-G:0-TimeStamp:01/04/2023 18:04:02'}

azext_iot/operations/hub.py
Elsie4ever: got it will update!

azext_iot/_help.py
Elsie4ever: I agree this is much more clear and easy to use, let me dive into how to implement these, thanks for the insight!

azext_iot/operations/hub.py
Elsie4ever: do you mean why I chose to use raw=True? yeah that's because when I try to mock response, Vicky and I tried different formats to pass in target_condition_error and custom_metric_query_errors, but all of them goes to 'additional properties' instead of the expected place. And using raw=True as Vicky suggested did fix this issue since the fixture now don't do assumption of where to put the properties that we mock anymore. Please let me know if there is better insight to fix this, thanks!😀

azext_iot/deviceupdate/_help.py
Elsie4ever: New page added, how about this? https://github.com/Azure/azure-iot-cli-extension/wiki/Inline-JSON-helps

azext_iot/operations/hub.py
digimaun: I'm not talking about raw=True, I'm referring to why we can't return `result` / the service response as is - instead, here we are manipulating the response

azext_iot/operations/hub.py
digimaun: Seems like low hanging fruit to ensure the users request meets the minimum, as in add validation to ensure at least target condition or custom metric queries has a value. The client error message would be more clear and better devEx compared to "BadRequest" when the command currently doesn't require either.

azext_iot/_help.py
digimaun: We can also think about (in a separate task) retrofitting the existing commands that support custom metric queries with this pattern for consistency.

Elsie4ever:
According to offline conversation, this cmd is not quite useful to cli, therefore scratching it

PR #625, opened 2023-02-07
Fix integration tests for existed resources

Elsie4ever:
Fix integration tests when the pytest.ini have stated exist resource
Issue found and fixed:
image
pipeline build: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7614&view=results
---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/tests/iothub/devices/test_iot_edge_devices_create_int.py
```diff
@@ -275,12 +275,6 @@ def _validate_results(
         cert_auth: bool = False,
         custom_device_template: bool = False,
     ):
-        # get all devices in hub
-        device_list = self.cmd(
-            f"iot hub device-identity list -n {self.entity_name} -g {self.entity_rg}"
-        ).get_output_in_json()
-        # make sure all devices were created
-        assert len(device_list) == len(devices)
```
c-ryan-k, line 283: This is a vital part of the test - essentially, if I pass a config file with 6 devices, I need to ensure that 6 devices were created after running the test.

If you're trying to run this same test multiple times on an existing hub, we may need to add a "delete all devices" step before it (or after, as part of cleanup) - but I don't think the solution should be to remove this assertion / test validation.

c-ryan-k (changes requested):
I see a few places in this PR that remove checks for ensuring hubs/resources are in a clean state (no devices, no configs, etc) before continuing - or as part of the test itself.

I think, instead of allowing existing resources to have old devices and configs, if we use an existing resource - we should add a task to clean it as part of test setup (remove all devices, remove all configs, etc).

Otherwise this can lead to stray / orphaned devices and configs building up on existing resources.

azext_iot/tests/iothub/devices/test_iot_edge_devices_create_int.py
Elsie4ever: thanks @c-ryan-k, you are right I deleted the length check since for existed hub, there might already be some devices so the length will not be equaled, and also I thought the code below seems like it's also checking that all devices are created successfully:
`device = self.cmd(
                f"iot hub device-identity show -d {device_id} -n {self.entity_name} -g {self.entity_rg}"
            ).get_output_in_json()` 
`assert device`

I'm thinking a way to not touch the original hub state and also able to check the length, how about save the device initial state for the hub and check here using something like `assert len(device_list) - len(original_device_list) == len(devices)`?😀

vilit1: I would also recommend looking at how the dt tests handle multiple dt instances. They keep track of all the dt instances created during a test run and in the end delete all of them

This may or may not be necessary for devices, modules, etc. (Think about what happens if you let the test run fully, cancel mid test, etc)

azext_iot/tests/deviceupdate/conftest.py
```diff
@@ -308,6 +309,25 @@ def _iothub_provisioner(request) -> Optional[dict]:
         if desired_instance_count:
             hub_id_map = {}
             hub_names = []
+
+            # Assign Data Contributor role to resource group if it does not exist
+            account = cli.invoke("account show").as_json()
+            scope_id = "/subscriptions/{}/resourceGroups/{}".format(
```
vilit1, line 315: 1. Should we be lazy and do this for all services? For other tests, we assign permissions one by one but if we can just add all the permissions needed to the rg and not worry about them after, it would be faster. 

But this idea is also a bad security practice and may end up with extra assignments that are not cleaned up in the end

2. Should we try to clean up resource permission assignments - as you got permission for x for a test, do we want to remove it post test? This sounds very annoying though.

@c-ryan-k  thoughts?

azext_iot/tests/iothub/__init__.py
```diff
@@ -94,31 +96,38 @@ def __init__(self, test_scenario, add_data_contributor=True):
...
+                    self._add_data_contributor(target_hub)
 
-                    if tries == MAX_RBAC_ASSIGNMENT_TRIES:
-                        raise Exception(
-                            "Reached max ({}) number of tries to assign RBAC role. Please re-run the test later "
-                            "or with more max number of tries.".format(MAX_RBAC_ASSIGNMENT_TRIES)
-                        )
+        else:
+            # when the hub instance is existed, make sure to add data contributor role
+            account = self.cmd("account show").get_output_in_json()
```
vilit1, line 103: why not make the _add_data_contributor handle this role assignment part too?

azext_iot/tests/iothub/__init__.py
```diff
@@ -130,41 +139,74 @@ def __init__(self, test_scenario, add_data_contributor=True):
...
                 self.cmd(
-                    "iot hub device-identity delete -d {} -n {} -g {}".format(
-                        device, self.entity_name, self.entity_rg
+                    "iot hub device-identity delete -d {} --login {}".format(
+                        device, self.connection_string
                     ),
                     checks=self.is_empty(),
                 )
 
+            for device in device_ids:
```
vilit1, line 181: why pop the first device if you are going to iterate anyways?

azext_iot/tests/iothub/__init__.py
```diff
@@ -94,31 +96,38 @@ def __init__(self, test_scenario, add_data_contributor=True):
...
+                'role assignment list --assignee "{}" --scope "{}" --role "{}"'.format(
+                    user["name"], target_hub["id"], USER_ROLE
+                )
+            ).get_output_in_json()
+
+            if len(role_assignments) == 0:
+                self._add_data_contributor(target_hub)
+
+            # Note down current devices and configs so they can be excluded from tear down when tests are done
+            self.original_device_list.extend(d["deviceId"] for d in self.cmd(
```
vilit1, line 120: In a real life test scenario you would want to also make sure that none of the new devices, deployments, etc would share a name with the current devices, deployments, etc 

But this is **very** unlikely since the generated ids for the devices should end up being different enough. So I would not mind if you skip this.

azext_iot/tests/iothub/__init__.py
```diff
@@ -94,31 +96,38 @@ def __init__(self, test_scenario, add_data_contributor=True):
...
+                f"iot hub device-twin list -n {self.entity_name} -g {self.entity_rg}"
+            ).get_output_in_json())
+
+            self.original_config_list.extend(c["id"] for c in self.cmd(
+                f"iot edge deployment list -n {self.entity_name} -g {self.entity_rg}"
+            ).get_output_in_json())
+
+            self.original_config_list.extend(c["id"] for c in self.cmd(
```
vilit1, line 128: does `iot hub configuration list` include configs and deployments or only adm configs?

azext_iot/tests/iothub/__init__.py
```diff
@@ -130,41 +139,74 @@ def __init__(self, test_scenario, add_data_contributor=True):
...
+            sleep(10)
 
-            for device in device_ids:
+        if tries == MAX_RBAC_ASSIGNMENT_TRIES:
+            raise Exception(
+                "Reached max ({}) number of tries to assign RBAC role. Please re-run the test later "
+                "or with more max number of tries.".format(MAX_RBAC_ASSIGNMENT_TRIES)
+            )
+
+    def clean_up(self, device_ids: List[str] = None, config_ids: List[str] = None):
```
vilit1, line 169: you should not need to keep track of created devices and original devices during the test. Pick one and then use the list commands to get all devices and you can figure out the other half from there.

azext_iot/tests/iothub/certificate/test_iothub_certificate_root_authority_int.py
```diff
@@ -9,23 +9,41 @@
...
+        super(TestIotHubCertificateRoot, self).tearDown()
+
+    def test_certificate_root(self):
+        authority = self.cmd(
+            f"iot hub certificate root-authority show -n {self.entity_name} -g {self.entity_rg}"
+        ).get_output_in_json()
+
+        self.original_enableRootCertificateV2 = authority["enableRootCertificateV2"]
+
+        if authority["enableRootCertificateV2"] is False:
```
vilit1, line 31: nit: you can do 
if not authority["enableRootCertificateV2"]:

azext_iot/tests/iothub/certificate/test_iothub_certificate_root_authority_int.py
```diff
@@ -9,23 +9,41 @@
 
 class TestIotHubCertificateRoot(IoTLiveScenarioTest):
     def __init__(self, test_case):
+        self.original_enableRootCertificateV2 = False
```
vilit1, line 12: so if the test is canceled/failed before the first show, the teardown will try to change the hub to v1 no matter if it was v1 or v2.

something I use is hasattr to check if an attribute exists for an object. This could solve the issue.

azext_iot/tests/iothub/configurations/test_iot_config_int.py
```diff
@@ -480,11 +480,12 @@ def test_edge_deployments(self):
             )
 
             config_list_check = [
-                self.check("length([*])", config_count),
                 self.exists("[?id=='{}']".format(config_ids[0])),
```
vilit1, line 483: why get rid of the length check?

azext_iot/tests/iothub/__init__.py
Elsie4ever: good point, how about I do check when generating new device id to make it's not repetitive to the old(existed device) names? Also, this kind of come back to @c-ryan-k's suggestion, should we clean up the devices for existed resource before running these tests?

azext_iot/tests/iothub/__init__.py
Elsie4ever: this was the original code actually, not sure why it shows that I added it, probably because I moved the line?

azext_iot/tests/iothub/certificate/test_iothub_certificate_root_authority_int.py
Elsie4ever: oh thanks! will try it out!

azext_iot/tests/iothub/certificate/test_iothub_certificate_root_authority_int.py
Elsie4ever: got it!

azext_iot/tests/iothub/__init__.py
vilit1: What do you think? First decide if the hub should be cleaned or not: Do you think if you give a hub, it should be cleaned before it is used (will help with weirdness, like lets say you have a config that would affect every device; the test resource is a test resource so it should not contain anything that you want to always keep) or should the current dataplane be kept (maybe you want to test out how that config would react with newly created devices in the test; maybe you are proud of your 10k device hub)?

azext_iot/tests/iothub/core/test_iot_messaging_int.py
```diff
@@ -42,6 +42,12 @@ def tearDown(self):
                 except OSError as e:
                     logger.error(f"Failed to remove {cert}. {e}")
 
+        self.cmd(
```
vilit1, line 45: root isnt that unique for a cert name so I can see conflicts here. This would entail either wiping the hub before hand (and not caring about old certs) or changing the root name to be some generic id (that would require finding what test makes it and changing that too)

azext_iot/tests/iothub/core/test_iothub_utilities_int.py
```diff
@@ -137,13 +137,13 @@ def test_iothub_connection_string_show(self):
...
+            devices = self.cmd(
                 self.set_cmd_auth_type(
                     f'iot hub query --hub-name {self.entity_name} -q "select * from devices"',
                     auth_type=auth_phase,
-                ),
-                checks=[self.check("length([*])", 0)],
-            )
+                )
+            ).get_output_in_json()
+            assert len(devices) >= 0
```
vilit1, line 146: I would prefer the length of devices to be something more concrete rather than >= 0 because the length of any list will always be >= 0

azext_iot/tests/iothub/core/test_iothub_storage_int.py
```diff
@@ -29,13 +29,15 @@
 JOB_POLL_MAX_ATTEMPTS = 3
 SETUP_SLEEP_INTERVAL = 10
 IDENTITY_SLEEP_INTERVAL = 60
+REMOVE_HUB_STORAGE_CONNECTION = False
```
vilit1, line 32: do you use this anywhere?

Elsie4ever: yeah I did in the same file, is github showing weird again?😥 it's here to track if I should remove storage connection at the teardown

azext_iot/tests/iothub/__init__.py
Elsie4ever: only configs, what is "adm" stands for? deployment management?

PR #626, opened 2023-02-07
Update Responses to use 0.22.0

vilit1:
Update to pin responses to the newest 0.22.0. Only tests were updated.

The biggest change is that now 500 return a different error - msrest.exceptions.ClientRequestError. This is similar to what the service returns for a hub with a non working dataplane.

Query has also been changed: if you have a 500 response with a next link, the responses package will go to the next page rather than failing right away. Tests have been updated to reflect this.

Note that 0.21.0 is skipped because the 500's return urllib3.exceptions.MaxRetryError

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/tests/digitaltwins/test_dt_twin_unit.py
+++ b/azext_iot/tests/digitaltwins/test_dt_twin_unit.py
@@ -11,7 +11,7 @@ import json
 from knack.cli import CLIError
 from azext_iot.digitaltwins import commands_twins as subject
 from msrest.paging import Paged
-from urllib3.exceptions import MaxRetryError
+from msrest.exceptions import ClientRequestError
 from azext_iot.tests.digitaltwins.dt_helpers import (
     etag,
     generate_generic_id,
@@ -245,7 +245,7 @@ class TestTwinCreateTwin(object):
         yield mocked_response
 
     def test_create_twin_error(self, fixture_cmd, service_client_error):
-        with pytest.raises((CLIError, MaxRetryError)) as e:
+        with pytest.raises((CLIError, ClientRequestError)) as e:
             subject.create_twin(
                 cmd=fixture_cmd,
                 name_or_hostname=hostname,
@@ -254,7 +254,7 @@ class TestTwinCreateTwin(object):
                 properties=None
             )
         if service_client_error.calls[0].response.status_code == 500:
-            assert isinstance(e.value, MaxRetryError)
+            assert isinstance(e.value, ClientRequestError)
         else:
             assert isinstance(e.value, CLIError)
 
@@ -309,7 +309,7 @@ class TestTwinShowTwin(object):
         [None, resource_group]
     )
     def test_show_twin_error(self, fixture_cmd, service_client_error, resource_group_name):
-        with pytest.raises((CLIError, MaxRetryError)) as e:
+        with pytest.raises((CLIError, ClientRequestError)) as e:
             subject.show_twin(
                 cmd=fixture_cmd,
                 name_or_hostname=hostname,
@@ -317,7 +317,7 @@ class TestTwinShowTwin(object):
                 resource_group_name=resource_group_name
             )
         if service_client_error.calls[0].response.status_code == 500:
-            assert isinstance(e.value, MaxRetryError)
+            assert isinstance(e.value, ClientRequestError)
         else:
             assert isinstance(e.value, CLIError)
 
@@ -424,7 +424,7 @@ class TestTwinUpdateTwin(object):
         yield mocked_response
 
     def test_update_twin_error(self, fixture_cmd, service_client_error):
-        with pytest.raises((CLIError, MaxRetryError)) as e:
+        with pytest.raises((CLIError, ClientRequestError)) as e:
             subject.update_twin(
                 cmd=fixture_cmd,
                 name_or_hostname=hostname,
@@ -436,7 +436,7 @@ class TestTwinUpdateTwin(object):
         status_codes = [call.response.status_code for call in service_client_error.calls]
         # Retries 5 times
         if status_codes == [202] + [500] * 5:
-            assert isinstance(e.value, MaxRetryError)
+            assert isinstance(e.value, ClientRequestError)
         else:
             assert isinstance(e.value, CLIError)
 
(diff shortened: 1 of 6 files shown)


azext_iot/tests/digitaltwins/test_dt_twin_unit.py
```diff
@@ -420,6 +433,12 @@ def test_update_twin_error(self, fixture_cmd, service_client_error):
                 resource_group_name=None,
                 etag=None
             )
+        status_codes = [call.response.status_code for call in service_client_error.calls]
+        # Retries 5 times
```
digimaun, line 437: What about having the service_client_error fixture patch the service client so it doesn't retry?

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
```diff
@@ -168,6 +169,10 @@ def test_query_twins_error(self, fixture_cmd, service_client_error):
                 show_cost=False,
                 resource_group_name=None
             )
+            # For some reason the new responses pkg doesnt fail on the first 500...
+            status_codes = [call.response.status_code for call in service_client_error.calls]
+            assert status_codes == [500, 200]
+            raise CLIError()
```
digimaun, line 175: Does the service_client_error fixture need to be constructed in a different way with the new version? It is suspicious to have to raise CLIError in the test to fulfill what the application logic is supposed to throw.

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
```diff
@@ -902,10 +903,14 @@ def test_config_list_invalid_args(self, fixture_cmd, top):
 
     def test_config_list_error(self, fixture_cmd, service_client_generic_errors):
         service_client_generic_errors.assert_all_requests_are_fired = False
-        with pytest.raises(CLIError):
+        with pytest.raises((CLIError, ClientRequestError)) as e:
```
digimaun, line 906: The error tests need to assert what the application logic will throw rather than the service client unless the command logic does not catch the error, why would a new responses version change that?

vilit1: The ClientRequestError is not caught by the command logic and this seems to align with my testing with a broken IoT Hub.

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
vilit1: This is for the test case that has the first request as a 500 (with a next link) and the second request as a 200 (from the next link). The result is that the command succeeds even though there is a 500.

I cannot test this in a real service...

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
vilit1: In a real service call, a 500 ends with the retries so this is closer to a real life situation. Plus the retries in the unit test are still very fast.

digimaun: Why is it not consistent where all other service_client_error based unit tests don't require this same setup for retries? And why does updating responses necessitate this?

vilit1: hmm ok let me look through the other retries and update them so the tests look for the retry calls

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
digimaun: Raising a CLIError in the test code that is supposed to validate the application code raises CLIError is wrong and not testing what its supposed to.

Looking at the respective digitaltwins client operation, if the response status code is not 200 it should raise models.ErrorResponseException.

image

You don't need a real service to test this flow, and at a minimum, one can change the status code of the response in a breakpoint before the response is inspected in further logic.

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
digimaun: The client is supposed to throw a `CloudError` if the response is not 200

image

And that is what the command handler is expecting

image

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
vilit1: The issue is that the code skips over the 500 (since there is a next link) and thus the final response status code is 200. So the case where the first link is a 500 with a next link to a 200 doesn't throw an error but succeeds...

Example:
`pytest azext_iot/tests/digitaltwins/test_dt_twin_unit.py::TestTwinQueryTwins::test_query_twins_error[service_client_error2]`

pdb prints:
image

vilit1: But I do understand that raising a CLIError in this case is very weird since the code does not error. My other suggestion is to remove this test case.

digimaun: I don't see how the code skips over the 500, when a 500 will interrupt page accumulation that happens with `accumulate_result`.

When I run the same test (with responses 0.20.0)

Right before the first API call (notice I have 2 break points here, before and after the call)

image


After getting the first response by the digitaltwins client
image

It's a 500 so raise models.ErrorResponseException ....I will continue the debugger from here to the next breakpoint
image

The next breakpoint is the one that wraps accumulate result and uses `handle_service_exception` to throw something derived from CLIError.
image

Which passes the existing test definition.

image

azext_iot/tests/digitaltwins/test_dt_twin_unit.py
digimaun: It's more about understanding the why this behavior in _this responses version_ before applying it everywhere.

azext_iot/tests/iothub/configurations/test_iot_config_unit.py
vilit1: The line above - aka the 273 response, now throws the error for a maxretry (500) service error

So the code never gets to checking the response.status_code and the error is not caught by our handle_service_exception

digimaun: now implying in responses 0.22.0 where it works as expected in 0.20.0

vilit1:
Closing this since I found a better way

PR #629, opened 2023-02-09
Tox Updates

c-ryan-k:
This overhauls our existing `tox.ini` to support the following local and pipeline test infrastructure.
I highly advise giving their [docs](https://tox.wiki) a quick overview, as it is much more in depth than this preview will cover.

## Local Tox
This changes nothing about how you currently run tests - local pytest and virtualenvs are untouched and unrelated.
In order to perform any of these new tasks, you must install tox (currently in `dev_requirements`, so already part of dev setup) and have the azure-cli repo cloned alongside your extension repo.

It supports local test configurations for the following environments (passed to tox with `-e "env"`):

- Linting
  - In order to run flake8 and pylint locally on your code, simply run `tox -e lint`
- Various Python and AZ CLI Versions
  - The tox environment string (passed to `-e`) will be parsed as such:
    `py{thon,3.7,3.8,3.9,3.10}-az{min,cur,dev}-{int,unit}`
  - Multiple options can be used inside of each `{matrix}` section (or just pick one): 
    `tox -e "py{3.7,3.10}-az{min,dev}-unit"`
  - Current CLI version support is as follows:
    - `azmin` installs the minimum supported CLI version (currently `2.32.0`)
    - `azcur` installs the latest released CLI version from PyPi
    - `azdev` installs the CLI from your local (`../azure-cli`) CLI instance
  - In order to list all environments, you can type `tox -av`, which will display them all in a list:
    
image

If you choose not to select a specific python version (which is also the current default), you can use `python` instead, to invoke whichever version `python` invokes in your environment.

### Important notes
- If you simply run `tox`, the current default environments (which can be changed inside tox.ini) are:
  - `lint`: which runs flake8 and pylint
  - `python-azdev-unit`: which runs unit test only, on your default python version, against your local (pathed to `../azure-cli`) CLI core version.
- The first time you run a new environment in tox, it will perform some setup tasks and dependency installation. It can look for changes in tox.ini and other related settings, but in order to rebuild a local environment you need to run tox with the `-r` switch.

## Tox Pipeline 
This also introduces a new `tox` pipeline (more work to come), which adds switches for each python version and azure CLI version, and distributes those tasks to simultaneous jobs:
image

This is currently a standalone pipeline, but the goal is to eventually configure this pipeline to utilize the existing built wheel and test SDK from our current pipelines to ensure our tests pass on all possible python and Azure CLI versions with very minimal user configuration.

---

### General Guidelines

Intent for Production

Basic expectations


c-ryan-k:
[Here's an example of the tox pipeline running](https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7635&view=results)

vilit1:
do we need to update dev_requirements (ex: add tox)

Also would be nice if you can put the pr description into the page we have about testing

c-ryan-k:
> do we need to update dev_requirements (ex: add tox)
> 
> Also would be nice if you can put the pr description into the page we have about testing

[tox is already in our dev_requirements](https://github.com/Azure/azure-iot-cli-extension/blob/706f394657caea065d6af490b424cb1204d516bb/dev_requirements#L14) - but I can absolutely work on a "tox testing guide", with this PR description as a base (and some examples of course)

vilit1 (approved):

digimaun:
I'm thrilled about this contribution

azext_iot/sdk/product/models/test_run.py
```diff
@@ -31,6 +31,9 @@ class TestRun(Model):
     :vartype certification_badge_results: list[object]
     """
 
+    # c-ryan-k - disable pytest trying to collect this file
+    __test__ = False
```
digimaun, line 35: Alternatively all the py2 models can be deleted from the code base

tox.ini
```diff
@@ -4,12 +4,67 @@
...
+commands =
+    python --version
+    # install to tox extension dir
+    pip install -U --target {envsitepackagesdir}/azure-cli-extensions/azure-iot .
+    # validate az and extension version
+    az -v
+    # run tests
+    # You can pass additional positional args to pytest using `-- {args}`
+    unit: pytest -k _unit ./azext_iot/tests {posargs}
+    int: pytest -k _int ./azext_iot/tests {posargs}
```
digimaun, line 72: have we thought about leveraging python-xdist as we do today in existing pipelines to parallelize worker processes for IT's ?

.azure-devops/tox.yml
```diff
@@ -0,0 +1,121 @@
...
+  - template: templates/clone-cli-repo.yml
+  - template: templates/setup-tox.yml
+  - template: templates/set-testenv-sentinel.yml
+  - task: UsePythonVersion@0
+    inputs:
+      versionSpec: $(python)
+      architecture: 'x64'
+    displayName: 'Use cached Python $(python) for tests.'
+
+  - script: $(pyTools.pythonLocation)/bin/tox -e "$(toxEnvStr)" -vv
```
digimaun: What's the advantage of executing tox via `$(pyTools.pythonLocation)/bin/tox` ? Seems more OS specific. Could `python -m tox ....` invoke the same tox module in a more agnostic fashion?

.azure-devops/tox.yml
```diff
@@ -0,0 +1,121 @@
+trigger: none
```
digimaun, line 1: As a next step we could replace the merge pipeline ....this pipeline is unit tests and linters only right?

.azure-devops/templates/setup-tox.yml
```diff
@@ -0,0 +1,14 @@
...
+  default: '3.8'
+
+steps:
+  - task: UsePythonVersion@0
+    displayName: Get Python for Python tools.
+    inputs:
+      versionSpec: ${{ parameters.pythonVersion }}
+      addToPath: false
+    name: pyTools
+  - script: $(pyTools.pythonLocation)/bin/pip install --upgrade tox
```
digimaun: similar question/comment as the [script tox invocation](https://github.com/Azure/azure-iot-cli-extension/pull/629/files#r1103260813) , does `python -m pip install ...` not work?

tox.ini
```diff
@@ -4,12 +4,67 @@
...
+deps =
+    {[base]deps}
     azure-cli
 commands =
-    pip install -r dev_requirements
-    pip install --upgrade ../azure-cli/src/azure-cli-testsdk
-    pytest azext_iot/tests
+    flake8 azext_iot/
+    pylint azext_iot/ --rcfile=.pylintrc
+
```
digimaun, line 46: What are your thoughts about incorporating the `azdev` linter? Specifically thinking of the help/ref docs linting.

digimaun:
> > do we need to update dev_requirements (ex: add tox)
> > Also would be nice if you can put the pr description into the page we have about testing
> 
> [tox is already in our dev_requirements](https://github.com/Azure/azure-iot-cli-extension/blob/706f394657caea065d6af490b424cb1204d516bb/dev_requirements#L14) - but I can absolutely work on a "tox testing guide", with this PR description as a base (and some examples of course)

Will that tox testing guide be part of this PR or a follow up PR?

azext_iot/sdk/product/models/test_run.py
digimaun: Just pointing it out, doesn't have to be in this PR.

tox.ini
digimaun: Looks like `tox` has some parallelization built-in  https://tox.wiki/en/latest/user_guide.html#parallel-mode

tox.ini
```diff
@@ -4,12 +4,67 @@
...
+[testenv:lint]
+description = run linter
+deps =
+    {[base]deps}
     azure-cli
 commands =
-    pip install -r dev_requirements
-    pip install --upgrade ../azure-cli/src/azure-cli-testsdk
-    pytest azext_iot/tests
+    flake8 azext_iot/
```
digimaun: for flake8 in existing processes we use `--statistics` to count errors & warnings, we also explicitly set the config but I suppose flake8 is finding it.

tox.ini
c-ryan-k: I haven't gotten too far into integration test parallelization yet, as I'm thinking we may want to plan out some more `testenv` separation / distinction for individual or groups of tests (similar to how we've moved tests into aspect folders).

This PR mostly captures the pipeline work for unit tests (with the capability to run int tests locally as part of tox as well).
In order to run tox for integration tests locally, I also believe we need to add our sizeable environment variable list into the tox environment `passenv` section as well. More to come on this in future iterations.

I've had mixed results with Tox's parallelization running unit tests - for a few reasons, but namely that debug output isn't printed during runs (which doesn't hurt as much locally as it does in a pipeline) and also some weird inconsistencies that could be due to how i'm setting up the separate test environments.

.azure-devops/templates/setup-tox.yml
c-ryan-k: Most likely, and I think that's the right solution. This was likely an artifact of some samples I was using when trying to "recreate" a version of [this pipeline](https://github.com/tox-dev/azure-pipelines-template/blob/master/run-tox-env.yml) (from tox) to manage multiple tox jobs using specific python versions.

This allows me to more explicitly say "hey use that very specific version of python we just asked for" (`pyTools`) "and make sure that we're running tox specifically from that interpreter for consistency"

Now that I think about it, once we hard-set a python version using `UsePythonVersion` - it should update path so that `python` points to that location as well, so we may not need to be this explicit.

I'll work on testing this change

.azure-devops/tox.yml
c-ryan-k: Correct. This is just unit tests and linters, with parameters to turn on/off individual python and CLI core testing versions.

One thing I plan to adjust eventually (in order to replace the merge pipeline) is to [replace the `pip install -U --target ...` command](https://github.com/Azure/azure-iot-cli-extension/pull/629/files/ebe777c0763d09091ed18f3cb05117548abb3d70#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R64) (as well as the [testSDK step](https://github.com/Azure/azure-iot-cli-extension/pull/629/files/ebe777c0763d09091ed18f3cb05117548abb3d70#diff-ef2cef9f88b4fe09ca3082140e67f5ad34fb65fb6e228f119d3812261ae51449R60)) with a step to install the artifacts from the primary build step in the merge pipeline.

tox.ini
c-ryan-k: I think it's definitely the direction to head.
I have some more small changes in this PR to at least get the initial pipeline / task squared away (python binary location, flake arguments, etc) - I think this change will slot in nicely once I begin some secondary changes to match feature parity of the current merge pipeline.

digimaun (approved):

.azure-devops/tox.yml
c-ryan-k: Going to merge this as-is, next step will be to replace unit tests in merge pipeline with this tech.

tox.ini
c-ryan-k: This will be part of a next PR that puts tox testing in the merge pipeline.

PR #634, opened 2023-02-22
Add fixtures for provision existed resource and factor out role assignment functions

Elsie4ever:
Add fixtures for provision existed resource and factor out role assignment helper functions in test dir

pipeline: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7836&view=results
---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/tests/digitaltwins/test_dt_data_history_lifecycle_int.py
```diff
@@ -439,10 +440,10 @@ def test_dt_data_history_adx_wait(self):
 
         # Check role assignments - needed once
         principal_id = create_output.get("identity").get("principalId")
-        assert len(self.get_role_assignment(
+        assert len(get_role_assignments(
```
vilit1, line 443: if you no longer use self.get_role_assignment, get rid of it

azext_iot/tests/iothub/conftest.py
```diff
@@ -72,11 +48,73 @@ def assign_iot_hub_dataplane_rbac_role(hub_results):
...
+                assignee=user["name"],
+                scope=target_hub_id,
+                role=USER_ROLE,
+                max_tries=MAX_RBAC_ASSIGNMENT_TRIES
             )
 
 
+@pytest.fixture()
+def fixture_provision_existing_hub_certificate(request):
+    if HUB_NAME:
```
vilit1: settings.env.azext_iot_testhub or hub_name?

azext_iot/tests/iothub/conftest.py
```diff
@@ -72,11 +48,73 @@ def assign_iot_hub_dataplane_rbac_role(hub_results):
...
+        # Make sure root certificate is Baltimore(Default)
+        authority = cli.invoke(
+            f"iot hub certificate root-authority show -n {HUB_NAME} -g {RG}"
+        ).as_json()
+
+        if authority["enableRootCertificateV2"]:
+            # Transition to Baltimore (initial transition)
+            cli.invoke(
+                f"""iot hub certificate root-authority set -n {HUB_NAME}
```
vilit1: no need for 3 " here

azext_iot/tests/helpers.py
```diff
@@ -156,6 +159,72 @@ def get_agent_public_ip():
     return requests.head("https://www.wikipedia.org").headers["X-Client-IP"]
 
 
+def get_role_assignments(
+    scope: str,
+    assignee: str = None,
+    role: str = None,
+):
```
digimaun: Might as well add return type hints to new functions we call into

azext_iot/tests/helpers.py
```diff
@@ -156,6 +159,127 @@ def get_agent_public_ip():
...
+        f'role assignment list --scope "{scope}" {role_flag} {assignee_flag}'
+    ).as_json()
+
+
+def assign_role_assignment(
+    role: str,
+    scope: str,
+    assignee: str,
+    max_tries=10,
+    wait=10
```
vilit1, line 189: wait should prob have a default of 60 here since it seems like a decent amount of calls use that as the wait time. Then you can prob remove some wait=60 in calls

Elsie4ever: ok thanks! will take a look!

vilit1 (approved):

azext_iot/tests/helpers.py
```diff
@@ -156,6 +159,127 @@ def get_agent_public_ip():
...
+        role_assignment_principal_ids = [assignment["principalId"] for assignment in role_assignments]
+        if assignee in role_assignment_principal_ids:
+            break
+        # else assign role to scope and check again
+        output = cli.invoke(
+            f'role assignment create --assignee "{assignee}" --role "{role}" --scope "{scope}"'
+        )
+        sleep(wait)
+        tries += 1
+    return output
```
c-ryan-k, line 207: In the spirit of tightening up our test reliability - 
If this fails after `max_tries` - Should we throw an error or log a warning?
I see we return the output, which is then parsed in some locations. Theoretically, this command could:
- Succeed, but not show in `get_role_assignments` inside our polling window
- Fail, but require us to wait the entire polling window anyway

In the first situation (succeed but not complete), I expect the output would have all of the correct parameters (role ID, etc), but future calls will fail with perms / auth errors. In the second situation, it seems we'd see an error trying to parse the `assign_role_assignment` output, since I don't think we check for `success()` first.

azext_iot/tests/helpers.py
```diff
@@ -156,6 +159,127 @@ def get_agent_public_ip():
...
+    ).as_json())
+
+    connection_string = cli.invoke(
+        "iot hub connection-string show -n {} -g {} --policy-name {}".format(
+            hub_name, rg, "iothubowner"
+        )
+    ).as_json()["connectionString"]
+
+    if device_list:
+        device = device_list.pop()
```
c-ryan-k, line 253: I'm a little confused by the logic here, it looks like we:
- Check if there are devices to delete
- If so:
  - pop the first device
  - delete it using `--login`
  - loop through all other devices
    - delete these using hub name and resource group
    
  The same logic applies below to `config_list`, but could we simplify the logic by just going through `device in device_list` instead of the extra steps for the first item?

azext_iot/tests/helpers.py
```diff
@@ -156,6 +159,127 @@ def get_agent_public_ip():
...
+    hub_name: str,
+    rg: str
+):
+    device_list = []
+    device_list.extend(d["deviceId"] for d in cli.invoke(
+        f"iot hub device-twin list -n {hub_name} -g {rg}"
+    ).as_json())
+
+    config_list = []
+    config_list.extend(c["id"] for c in cli.invoke(
```
c-ryan-k, line 238: Should this be kept in a separate array for `deployment_list`, etc?

Elsie4ever: yeah we can! will update in another open pr

azext_iot/tests/helpers.py
Elsie4ever: Hi Ryan, this was the original logic before I touch the code and it's showing green since a factored them out. I was not clear about the history of the --login for first device so I didn't touch this. But after trying out your suggestion I don't see any issue removing the extra first step, therefore will update!😀

azext_iot/tests/helpers.py
Elsie4ever: thanks! will add check for success()!

PR #638, opened 2023-03-01
Add max-models-per-batch parameter to dt models creation and fix test

Elsie4ever:
---
**Upon investigation, the model create from directory was failing across multiple version of ubuntu and can be obeserved from multiple pipeline build, and reducing the max-models-per-batch default to 30 fix the issue(it will then take more time to make sure models are pull in right order) therefore to prevent the user from ubuntu encounter error when dt service is flaky in the future, adding this as a parameter that default to 30. Also added this value to environment variable so we can adjust it from pipeline if this issue happen again.
Pipeline build:
Ubuntu 18.04: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7884&view=results
Ubuntu 20.04: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7831&view=results
Ubuntu 22.04: https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7833&view=results**

### General Guidelines

Intent for Production

Basic expectations


azext_iot/digitaltwins/params.py
```diff
@@ -456,6 +456,14 @@ def load_digitaltwins_arguments(self, _):
             "In case of an error processing a batch, the behavior is determined by the --failure-policy parameter. ",
             arg_group="Models Input",
         )
+        context.argument(
+            "max_models_per_batch",
+            options_list=["--max-models-per-batch", "--mmpb"],
+            help="Indicates the maximum model size per batch when create models using directory contains "
+            "more than 250 models. The default batch size will be 30. This value could be useful when encouter "
+            "DTDLParser error. Reduce the number to prevent models pulling disordered from directory. ",
```
vilit1: The maximum model size per batch when creating more than 250 models. Reduce this number to prevent a DTDLParser error.

default should be shown automatically when you have a default argument in the command params (try running the command with -h)

also not sure what the last sentence (Reduce the number to prevent models pulling disordered from directory) means but we should put that into help if it is important

azext_iot/digitaltwins/commands_models.py
```diff
@@ -13,11 +13,16 @@
 
 def add_models(
     cmd, name_or_hostname, models=None, from_directory=None,
-    resource_group_name=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value
+    resource_group_name=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value,
+    max_models_per_batch=None
```
vilit1: if this is set to a default of 30, then it will show that in -h

azext_iot/digitaltwins/providers/model.py
```diff
@@ -16,7 +16,7 @@
 logger = get_logger(__name__)
 MAX_MODELS_API_LIMIT = 250
 # avagraw - Max number of models the API's dependency resolution can handle when models are created across multiple API calls.
-MAX_MODELS_PER_BATCH = 35
+MAX_MODELS_PER_BATCH = 30
```
vilit1: this discrepancy is very concerning because there is almost a 10 times difference between what the api promises and actually delivers

digimaun: These findings should be captured and discussed with ADT.

Elsie4ever: yeah this is changed upon testing, and the time that the test run increased is around 3-4 seconds. I think the problem is not quite about itself at this point, it might because we sent the requests too fast and the model imported are becoming disordered😥

digimaun: Comment on the "...we sent the requests too fast and the model imported are becoming disordered" , if you are talking about the service side processing.....I would argue the service should not return so quickly if processing the models hasn't been finished, IIRC this add models operation is intended to be synchronous.

.azure-devops/templates/set-testenv-sentinel.yml
```diff
@@ -72,4 +73,5 @@ steps:
       AZEXT_DT_EP_EVENTHUB_POLICY: $(azext_dt_ep_eventhub_policy)
       AZEXT_DT_EP_EVENTHUB_TOPIC: $(azext_dt_ep_eventhub_topic)
       AZEXT_DT_EP_RG: $(azext_dt_ep_rg)
+      AZEXT_DT_MAX_MODEL_PER_BATCH: $(azext_dt_max_model_per_batch)
```
digimaun: We should try to be consistent with our naming, the parameter uses plural models, `--max-models-per-batch`

CONTRIBUTING.md
```diff
@@ -216,6 +216,7 @@ For all resources, if the environmental variable is not provided, a new instance
 | `azext_dt_ep_eventhub_topic`	| Digital Twin Data History and Endpoint Tests	| The Event Hub instance in the Event Hub Namespace to use.	|
 | `azext_dt_ep_eventhub_topic_consumer_group`	| Digital Twin Data History Tests	| The Event Hub consumer group name to use. Defaults to "$Default".	|
 | `azext_dt_ep_rg`	| Digital Twin Data History and Endpoint Tests	| The resource group that contains the endpoint (Event Hub, Event Grid, Service Bus) variables. If not provided, `az_iot_testrg` will be used as the resource group.	|
+| `azext_dt_max_model_per_batch`	| Digital Twin Model Tests	| The maximum value of process models in batches if models to process exceed the API limit. Defaults to 30.	|
```
digimaun: Description suggestion: "The maximum number of models per batch to submit to the DT Instance if the total set of models exceed the single page API limit."

azext_iot/digitaltwins/providers/model.py
digimaun: This can be referenced for the default value of the respective param.

azext_iot/digitaltwins/params.py
digimaun: I like the description suggestion in the first sentence of @vilit1's comment, its clear and concise.

Regarding the last sentence, the client should be sending models in the right order, dependencies of models should be in the same or prior batch. If client side processing has issues, we should fix it - however our prior understanding was a service side processing lag issue.

azext_iot/digitaltwins/params.py
```diff
@@ -456,6 +456,14 @@ def load_digitaltwins_arguments(self, _):
             "In case of an error processing a batch, the behavior is determined by the --failure-policy parameter. ",
             arg_group="Models Input",
         )
+        context.argument(
```
digimaun, line 459: You are able to add a type constraint to the argument definition https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/_params.py#L880

azext_iot/digitaltwins/providers/model.py
```diff
@@ -70,12 +70,17 @@ def __init__(self, cmd, name, rg=None):
...
+            models=None,
+            from_directory=None,
+            failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value,
+            max_models_per_batch=None):
         if not any([models, from_directory]):
             raise RequiredArgumentMissingError("Provide either --models or --from-directory.")
 
         # If both arguments are provided. --models wins.
         payload = []
+        models_per_batch = int(max_models_per_batch or MAX_MODELS_PER_BATCH)
```
digimaun: This or shouldn't be necessary once the parameter has the default value set

digimaun (commented):
Review chunk

azext_iot/digitaltwins/params.py
```diff
@@ -456,6 +456,14 @@ def load_digitaltwins_arguments(self, _):
...
+            "max_models_per_batch",
+            options_list=["--max-models-per-batch", "--mmpb"],
+            type=int,
+            help="The maximum model size per batch when creating more than 250 models."
+            "Reduce this number to prevent a DTDLParser error. ",
```
vilit1, line 465: q - if the number of models is less than 250, will this input be ignored?

Ex: there are 120 models and the mmpb is set to 30 -> will there be one batch of 100 or 4 batches of 30?

Elsie4ever: yeah this 30 will be ignored and not use if the number of models is less than 250😀

azext_iot/tests/digitaltwins/test_dt_model_unit.py
```diff
@@ -186,7 +185,7 @@ def delete_request_callback(request):
                     models=None,
                     from_directory=ontology_directory,
                 )
-                assert len(models_added) == MAX_MODELS_PER_BATCH
+                assert len(models_added) == 30
```
digimaun: I think how it was before was better (referencing the variable rather than static integer)

azext_iot/digitaltwins/commands_models.py
```diff
@@ -13,11 +13,16 @@
 
 def add_models(
     cmd, name_or_hostname, models=None, from_directory=None,
-    resource_group_name=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value
+    resource_group_name=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value,
+    max_models_per_batch=30
```
digimaun: Setting the default here could  still be by variable reference

vilit1: Basically, make a variable to reference the 30 in common.py and reference that here and in tests :)

azext_iot/digitaltwins/params.py
vilit1: Either add a help example to show this or add in "This number will be ignored if the number of models is less than 250". Sort of leaning towards help examples b/c I hate long param descriptions

Also is there a unit test for this

Elsie4ever: should I still add example in help if we hide the parameter to user?😀

azext_iot/tests/digitaltwins/test_dt_model_unit.py
Elsie4ever: thanks will update!

azext_iot/digitaltwins/params.py
Elsie4ever: and good idea I will add a unit test for the non default value, thanks!

HISTORY.rst
```diff
@@ -3,6 +3,14 @@
 Release History
 ===============
 
+unreleased
++++++++++++++++
+
+**Digital Twins updates**
+
+* Addition of new parameter `--max-models-per-batch` for `az dt model create` to let user adjust batch size when directory exceeds
```
digimaun, line 11: Because this parameter is intended to be a hidden stop-gap, I wonder whether we _should_ mention it in release notes or at least say its a hidden temporary measure...

azext_iot/digitaltwins/providers/model.py
```diff
@@ -70,12 +68,17 @@ def __init__(self, cmd, name, rg=None):
         )
         self.model_sdk = self.get_sdk().digital_twin_models
 
-    def add(self, models=None, from_directory=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value):
+    def add(self,
+            models=None,
+            from_directory=None,
+            failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value,
+            max_models_per_batch=None):
```
digimaun: nit, at least we can add type hints to new stuff

azext_iot/digitaltwins/common.py
```diff
@@ -49,6 +49,9 @@
 # Default Event Hub Consumer Group
 DEFAULT_CONSUMER_GROUP = "$Default"
 
+# Models create
+MAX_MODEL_PER_BATCH = 30
```
digimaun: nit, plural models makes more sense in this context `MAX_MODELS_PER_BATCH`

azext_iot/digitaltwins/providers/model.py
```diff
@@ -70,12 +68,17 @@ def __init__(self, cmd, name, rg=None):
...
         if not any([models, from_directory]):
             raise RequiredArgumentMissingError("Provide either --models or --from-directory.")
 
         # If both arguments are provided. --models wins.
         payload = []
+        models_per_batch = int(max_models_per_batch)
```
digimaun: In what case will max_models_per_batch not be an int? i.e. is this cast needed?

azext_iot/tests/digitaltwins/test_dt_model_unit.py
```diff
@@ -131,32 +133,44 @@ def test_add_models(
...
-    @responses.activate
-    def test_large_ontology_error(self, fixture_cmd, fixture_dt_client):
-        models_added = []
-        models_deleted = []
-
-        def post_request_callback(request):
+    def configure_large_ontology_create_response(self, responses, models_added):
+        def post_request_callback_first(request):
             payload = json.loads(request.body)
+            # Check the batch order starts from model with least dependency
```
digimaun, line 139: This comment is no longer accurate , the assertion is now checking if model with least dependency is in the list of model ids rather than the first entry.

azext_iot/tests/digitaltwins/test_dt_model_unit.py
```diff
@@ -226,7 +222,27 @@ def post_request_callback(request):
...
+    @pytest.mark.usefixtures("set_cwd")
+    @responses.activate
+    def test_large_ontology_error_custom_batch_size(self, fixture_cmd, fixture_dt_client):
+        models_added = []
+
+        self.configure_large_ontology_create_response(responses, models_added)
+
+        ontology_directory = "./references/opendigitaltwins-building/Ontology"
+        if os.path.isdir(ontology_directory) and len(os.listdir(ontology_directory)) > 0:
+            with pytest.raises(CLIError):
```
digimaun, line 244: Can we at least add a comment explaining the reason for this CLIError?

digimaun (commented):
Review chunk

HISTORY.rst
Elsie4ever: got it, should I also mention here the error we got and will delete the parameter once fixed?

digimaun: Sure

azext_iot/tests/digitaltwins/test_dt_model_unit.py
digimaun: This feedback item is more about, a question of are we losing test accuracy by the refactor to a less specific assertion?

Elsie4ever: Hi Paymaun, the reason I change the original check to this is due to: I found first model been added to the list is different from windows vs. ubuntu vs. macOS, for windows it's **dtmi:digitaltwins:rec_3_3:core:Agent;1**, however for ubuntu it's **dtmi:digitaltwins:rec_3_3:business:Role;1**, and for macOS it's **dtmi:digitaltwins:rec_3_3:core:Collection;1** they **all** have 0 model dependency, my assumption is the directory is scanned differently windows vs. ubuntu vs. macOS. The reason why we didn't got test failure is due to the assertion code under CliErrors was _never executed_ due to wrong indentation before this pr. Therefore i need to change a way to test this dtmi:digitaltwins:rec_3_3:core:Agent;1, since it's not always the first entry. Or I can change the assertion into `assert model_ids[0] == "dtmi:digitaltwins:rec_3_3:core:Agent;1" or model_ids[0] == "dtmi:digitaltwins:rec_3_3:business:Role;1" or model_ids[0] == "dtmi:digitaltwins:rec_3_3:core:Collection;1"`, but that would be too complicated, what do you think?😀

Elsie4ever: Think of another way to make the assertion specific without checking the hardcoded id: I can get the dependency length of first entry directly from the model definition and make sure it's 0, just updated the code, please let me know if this work, thanks!

azext_iot/digitaltwins/providers/model.py
```diff
@@ -70,12 +68,17 @@ def __init__(self, cmd, name, rg=None):
         )
         self.model_sdk = self.get_sdk().digital_twin_models
 
-    def add(self, models=None, from_directory=None, failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value):
+    def add(self,
+            models=None,
+            from_directory=None,
+            failure_policy=ADTModelCreateFailurePolicy.ROLLBACK.value,
+            max_models_per_batch: int = None):
```
digimaun: I don't think currently a default of `None` works here for `max_models_per_batch`. If no value is provided, then `None` is used for the comparison and further logic on line 121/122. This should either have no default, or it can be typed as `Optional[int]` where if `None` is the value, the logic can add a default.

Elsie4ever: got it, since I used MAX_MODELS_PER_BATCH as default for add_models, so the default is never used here, will remove

digimaun (approved):
Conditional approval given clean IT pipeline run(s)

vilit1:
Undo temp and then I will merge. It seems like something is off with dt endpoint create so I will investigate in a future pr to see if it is a service issue or something we need to fix

Elsie4ever:
new pipeline build https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7909&view=results

azext_iot/tests/helpers.py
```diff
@@ -204,6 +205,9 @@ def assign_role_assignment(
         )
         sleep(wait)
         tries += 1
+    if not output.success():
+        raise CLIInternalError(f"Failed to assign '{assignee}' the role of '{role}' against scope '{scope}'.")
```
c-ryan-k: On second thought (apologies for the run-around), I'm wondering if we should make this a warning?
 Is it better to:
 - fail the whole test run because of an error in the assignment create for a particular resource, or should we 
 - just let the tests continue and log a warning so we can debug later? 
 
 I'm not sure it even matters, just something I wanted to bring up. I'm fine with either. @vilit1 - do you have any preference?

digimaun:
Why is this PR being used to apply feedback from an unrelated PR? A new PR can be made with follow ups.

Elsie4ever:
> Why is this PR being used to apply feedback from an unrelated PR? A new PR can be made with follow ups.

sorry, will create a new one😋

azext_iot/tests/helpers.py
Elsie4ever: ya a warning should be enough for this case, thanks for the feedback!

Elsie4ever:
Pipeline run succeeded https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=7917&view=results

PR #665, opened 2023-05-26
Iot message endpoint update

vilit1:
Originally an ask from the customer to update connection strings for endpoints. I expanded on this idea to create this update command:
1. Like create, update is also subdivided into commands by type. This limits the amount of checking we need to do (we can assume that inputs will be valid for that type) and makes the help clean and clear
2. Allows for changes to any property allowed by the service without deleting associated routes or message enrichments. Possible uses can be:
  a. Change auth type for the endpoint from connection string to identity based (encouraged by Azure)
  b. Remove certain properties from the endpoint (cosmos db partition key names)
  c. Change the source of the endpoint (eventhub A -> eventhub B)
3. If changing authentication type, null out unneeded properties, like connection string for when the endpoint becomes identity based.
4. Add a keyword that would signify "I want you to fetch the connection string for me".

Todo: add unit tests, make sure there are no missed gaps

History:
```
* Add command group `az iot hub endpoint update`, which will allow you to update your endpoint without deleting it.
```

---

### General Guidelines

Intent for Production

Basic expectations


azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+        return endpoint
+
+    hub_mock = mocker.MagicMock()
+    hub_mock.properties.routing.endpoints.event_hubs = [create_mock_endpoint()]
+    hub_mock.properties.routing.endpoints.service_bus_queues = [create_mock_endpoint()]
+    hub_mock.properties.routing.endpoints.service_bus_topics = [create_mock_endpoint()]
+    hub_mock.properties.routing.endpoints.storage_containers = [create_mock_endpoint()]
+    hub_mock.properties.routing.endpoints.cosmos_db_sql_collections = [create_mock_endpoint()]
+
+    def initialize_mock_client(self, *args):
```
vilit1, line 57: this mock client can be reused for other control plane areas

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+get_servicebus_queue_cstring_path = "azext_iot.iothub.providers.message_endpoint.get_servicebus_queue_cstring"
+get_cosmos_db_cstring_path = "azext_iot.iothub.providers.message_endpoint.get_cosmos_db_cstring"
+parse_cosmos_db_cstring_path = "azext_iot.iothub.providers.message_endpoint.parse_cosmos_db_connection_string"
+get_storage_cstring_path = "azext_iot.iothub.providers.message_endpoint.get_storage_cstring"
+
+
+@pytest.fixture()
+def fixture_update_endpoint_ops(mocker):
+    # Fetch connection string
+    arg_check = mocker.patch(arg_check_path)
```
vilit1: Would really like feedback here:
1. is it fine that the requests aren't made (as we aren't matching to a mocked response; if we did so, we would need to mock the entire hub object, not just a small part like here)
2. should we have each test mock its own endpoint list? Or should we keep it as it is now (creating all endpoint types at once and using that for tests)

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+                "resource_group_name": generate_names(),
+            },
+        ]
+    )
+    def test_message_endpoint_update_event_hub(self, mocker, fixture_cmd, fixture_update_endpoint_ops, req):
+        result = subject.message_endpoint_update_event_hub(
+            cmd=fixture_cmd,
+            hub_name=hub_name,
+            endpoint_name=endpoint_name,
+            **req
```
vilit1, line 124: is this lazy implementation ok?

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+        assert req.get("resource_group_name") == resource_group
+        hub_resource = fixture_find_resource.call_args[0][0].client.begin_create_or_update.call_args[0][2]
+        endpoints = hub_resource.properties.routing.endpoints.event_hubs
+        assert len(endpoints) == 1
+        endpoint = endpoints[0]
+
+        assert endpoint.name == endpoint_name
+        mock = mocker.Mock
+
+        # if a prop is not set, it will be a Mock object
```
vilit1, line 139: The idea behind these checks is that if a property of an endpoint is not modified by update, it will show up as a mock object. Are you all ok with this

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+    yield find_resource, arg_check
+
+
+class TestMessageEndpointUpdate:
+    @pytest.mark.parametrize(
+        "req",
+        [
+            {},
+            {
+                "endpoint_account_name": None,
```
vilit1: Should I keep the key/value pairs where the value is None? It helps with understanding the test but does make it so there is more stuff to read

azext_iot/iothub/params.py
```diff
@@ -577,10 +566,39 @@ def load_iothub_arguments(self, _):
...
+
+    with self.argument_context("iot hub message-endpoint update") as context:
         context.argument(
-            "endpoint_account_name",
-            options_list=["--endpoint-account"],
-            help="The account name for the endpoint resource.",
+            "connection_string",
+            options_list=["--connection-string", "-c"],
+            help="Connection string of the routing endpoint. Use `update` if you want the connection string to "
+            "be fetched from the service instead of providing it in the argument.",
```
vilit1: Note that `update` is for "I want the CLI to find the new connection string for me. 

This is difficult since we can get sources for the connection string from different places:
1. the user inputs (which I currently enforce) - similar to create
2. the existing connection string
3. the existing properties of the endpoint (if no connection string)

We could also say if you want to update the connection string, you must provide the connection string yourself (and skip the update)

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+        endpoint_uri: Optional[str] = None,
+        entity_path: Optional[str] = None,
+        database_name: Optional[str] = None,
+        primary_key: Optional[str] = None,
+        secondary_key: Optional[str] = None,
+        partition_key_name: Optional[str] = None,
+        partition_key_template: Optional[str] = None,
+        identity: Optional[str] = None
+    ):
+        # if nothing is provided -> should we block?
```
vilit1, line 242: Question. Currently other update commands do not block if no arguments are provided (as far as I know)

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+        # have the user say the type. Will make args easier (as in we do not need to check for unneeded args)
+        original_endpoint = self._show_by_type(endpoint_name=endpoint_name, endpoint_type=endpoint_type)
+
+        if any([connection_string, primary_key, secondary_key]) and identity:
+            optional_msg = ", --primary-key and/or --secondary-key," if primary_key or secondary_key else ""
+            error_msg = "Please use either --connection-string" + optional_msg + " or --identity."
+            raise MutuallyExclusiveArgumentError(error_msg)
```
vilit1, line 250: Prob make the test check the msg logic for this. primary key and secondary key would only show up in cosmos db...

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+            original_endpoint.authentication_type = AuthenticationType.IdentityBased.value
+            if identity == SYSTEM_ASSIGNED_IDENTITY:
+                original_endpoint.identity = None
+            else:
+                original_endpoint.identity = ManagedIdentity(
+                    user_assigned_identity=identity
+                )
+        elif any([connection_string, primary_key, secondary_key]):
+            original_endpoint.identity = None
+            original_endpoint.authentication_type = AuthenticationType.KeyBased.value
```
vilit1, line 284: Note that we can set auth type to None here and it will default to keybased

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+            if EndpointType.CosmosDBContainer.value != endpoint_type.lower():
+                original_endpoint.endpoint_uri = None
```
vilit1, line 286: Note that this means if a user is updating a non cosmos db endpoint to be keybased, we will ignore the endpoint_uri argument.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+                    endpoint_account_name=endpoint_account_name,
+                    entity_path=entity_path,
+                    endpoint_policy_name=endpoint_policy_name
+                )
+                fetch_connection_string = True
+                if not endpoint_resource_group:
+                    endpoint_resource_group = self.hub_resource.additional_properties["resourcegroup"]
+                if not endpoint_subscription_id:
+                    endpoint_subscription_id = self.hub_resource.additional_properties['subscriptionid']
+            elif EndpointType.CosmosDBContainer.value != endpoint_type.lower():
```
vilit1: Reverse this statement for readability

azext_iot/iothub/params.py
vilit1: Also note that if we do 2 and 3, we would have to decide what we would want to do if a person gives some user input that is used to get the connection string. It should win over the existing properties.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+                    cmd=self.cli,
+                    account_name=endpoint_account_name,
+                    rg=endpoint_resource_group,
+                    sub=endpoint_subscription_id
+                )
+            if container_name:
+                original_endpoint.container_name = container_name
+            if encoding:
+                original_endpoint.encoding = encoding.lower()
+            if file_name_format:
```
vilit1, line 310: was originally thinking of adding the option to revert to default here but decided against since it creates more custom param inputs (like "update" or "")

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +218,226 @@ def create(
...
+                    original_endpoint.primary_key = parsed_cs["AccountKey"]
+                    original_endpoint.secondary_key = parsed_cs["AccountKey"]
+                # parse out endpoint uri from connection string
+                if not endpoint_uri:
+                    original_endpoint.endpoint_uri = parsed_cs["AccountEndpoint"]
+            if database_name:
+                original_endpoint.database_name = database_name
+            if container_name:
+                original_endpoint.container_name = container_name
+            if partition_key_name:
```
vilit1, line 330: Not sure how much the service team will care if we allow the user to null out the partition key template (and keep the partition key name). For create, they want us to set it to a default in this scenario

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
```diff
@@ -143,6 +143,83 @@ def test_iot_eventhub_endpoint_lifecycle(provisioned_event_hub_with_identity_mod
     assert len(eventhub_list) == 3
     assert endpoint_list["eventHubs"] == eventhub_list
 
+    # Update
+    # Keybased -> System
```
vilit1, line 146: this is somewhat minimal testing compared to what the user can do with the endpoints (specifically I skip changing to a different resource (could be shared access policy key for keybased, eventhub, eventhub namespace; full testing will require having a second eventhub namespace and making more eventhubs and policies within one namespace)

azext_iot/iothub/_help.py
```diff
@@ -645,6 +645,146 @@ def load_iothub_help():
               --container {container_name} --encoding json --identity {user_identity_resource_id}
     """
 
+    helps[
+        "iot hub message-endpoint update"
```
vilit1, line 649: Added this as a way to organize params. I do not want the same mess as with the original routing-endpoint create.

azext_iot/iothub/_help.py
```diff
@@ -645,6 +645,146 @@ def load_iothub_help():
...
+        short-summary: Update the properties of an existing endpoint for an IoT Hub.
+    """
+
+    helps[
+        "iot hub message-endpoint update cosmosdb-container"
+    ] = """
+        type: command
+        short-summary: Update the properties of an existing Cosmos DB Container endpoint for an IoT Hub.
+        long-summary: |
+                      If updating the authentication type from a key-based to identity-based Cosmos DB Container endpoint, you must provide the endpoint account (via --endpoint-account). The old primary key and secondary key will be nulled out.
```
vilit1, line 661: we could add some confirmations based on if the user is changing auth types, entire namespaces, etc. But I do not like talkative cli commands so I will not add unless there is great desire.

azext_iot/iothub/params.py
```diff
@@ -577,10 +566,39 @@ def load_iothub_arguments(self, _):
             "{YYYY}. Any one placeholder may be specified at most once, but order and non-placeholder components are "
             "arbitrary. If partition key name is provided, partition key template defaults to {deviceid}-{YYYY}-{MM}",
         )
+
+    for endpoint_type in ["cosmosdb-container", "storage-container"]:
```
vilit1: I could do something similar for the create and update only params

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
```diff
@@ -0,0 +1,938 @@
...
+from azure.mgmt.iothub.models import ManagedIdentity
+
+hub_name = "hubname"
+hub_rg = "hubrg"
+endpoint_name = generate_names()
+resource_not_found_error = "Resource not found."
+generic_response = {generate_names(): generate_names()}
+
+
+path_find_resource = "azext_iot.iothub.providers.discovery.IotHubDiscovery.find_resource"
```
vilit1: Will need to clean this up once we finalize a unit testing style

azext_iot/iothub/_help.py
c-ryan-k: Agree - I think for destructive actions (deleting / removing endpoints and routes) a confirmation is warranted, but for updates it's likely not neccessary.

azext_iot/iothub/params.py
c-ryan-k: Might need to track down which params are in which commands (create / update across the matrix for various endpoint types) - but as long as they're consistent, it doesn't necessarily hurt to put all params under a parent `iot hub message-endpoint` context

azext_iot/iothub/params.py
c-ryan-k: Can we check for `--connection-string` or `-c` being passed as a param but with no argument / `None` instead? Or possibly a second param (though I know this makes the logic pretty wacky) that signifies this exact behavior?

To make sure I'm understanding correctly, this `update` value means "I mean to provide a connection string, but please gather the properties for me instead of making me type it myself" correct?

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Do you mean block, as in.... error? Or block as in simply "do nothing" instead of running an update with the same hub object that comes from the GET?

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: That's a good point, we don't have many tests that check for accuracy of error messages themselves, most are simply for throwing the right exception type.

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: In that case, does it make more sense to set it to `None` after `if identity` before we start doing checks, and only set it if we need to set it to identity-based? 

Otherwise it's probably fine, I remember a similar issue cropping up in CLI core - as long as it's correctly set to identity when we need it to be, defaulting to keybased is probably fine as it's been the de-facto default since the beginning of the feature I believe.

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Would it make sense to warn the user that they're providing input that's being ignored?

I'm not sure if that would break existing workflows (probably not since this command space is new) - but it might be worth letting the user know "this value isn't being used to update to the [endpoint_type] you're trying to update"

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Most of our control-plane updates (checkout CLI core for legacy examples 🙃) use the format:

```
def update():
   model = get_model()
   
   if new_property_value:
      model.old_property = new_property_value
   if other_new_property_value:
      model.other_old_property = other_new_property_value
      
   service.update(model)
```

I think it's totally acceptable to simply "only update the params that are passed in" as it matches our existing patterns.
Future thinking / brainstorming out loud - maybe there's a possibility to add a `--set-default prop1 prop2` type command which takes in a space-separated list of properties to reset to their default values? I can see this being very tricky to scope / troubleshoot though.

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: We can always ask the service folks for their considerations, but in the past we've been committed to letting the user input values as they see fit, stopping them if there's a conflict (like mutually exclusive arguments, invalid values, etc) and letting the service tell us if something went wrong processing their input.

c-ryan-k (commented):
Chunk of review addressing comments

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
c-ryan-k: I think either way is fine - either explicitly set them to `None`, or don't set them, and then check that their values are the same as before (either non-existent or `None`) - am I understanding that correctly?

azext_iot/iothub/providers/message_endpoint.py
vilit1: block as in do nothing/do not call the api

so in the current scenario, the call is still made (like how other update calls function)

azext_iot/iothub/providers/message_endpoint.py
vilit1: I do have a disclaimer in the long summary that describes which props will be nulled out for which type. 

I could also add a warning but it will require more if statements to this if statement mess...

azext_iot/iothub/providers/message_endpoint.py
vilit1: ok will leave it as power to the user

azext_iot/iothub/providers/message_endpoint.py
vilit1: we could do a set default but then it could become confusing as you can do `--set-default encoding --encoding "blah"` or have people trying to set defaults to things that have no defaults... (and have us do more checks)

would be a future consideration but I am not sure how much traction it would get unless we have most params have defaults (which is not true in the current situation)

azext_iot/iothub/providers/message_endpoint.py
vilit1: I am assuming you are saying set auth type to None in this elif statement. Which is reasonable. But will keep it as keybased so it is more consistent with create

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
vilit1: if they are not present, it is the same as if they are None (or the default which is None)

In the tests, I am checking if a param was set by seeing if the mock object has it or not (if it doesnt, trying to call that attribute will cause it to give a mock object)

azext_iot/iothub/params.py
vilit1: let me check again if there is a way to have it as a flag. It would have to be something so that we know if the user wants us to update the connection string or not touch it at all

I was going to have a separate arg but decided it is prob better to keep one arg and I followed the update pattern from identity [system]

Yes, the update value means I want to update the connection string but I want you to get it for me

vilit1: There is a way
image

but the more I think about it, the more I want to keep it as force the user to type in "update" to show that they want us to update the connection string (which would hold more weight if the identity type is changed)

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_unit.py
vilit1: also regarding 1, by doing this we are bypassing the need to check if the sdk supports cosmos db or not (since we catch the call right before the sdk is called)

azext_iot/iothub/params.py
vilit1: After talking offline with Ryan, I decided to just make this only take in the connection string. So the user must provide the connection string (and we would not look it up for them)

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +219,174 @@ def create(
...
+            endpoint_list = endpoints.storage_containers
+        # note that cosmos db support will be determined by if the user can call the command.
+        if endpoint_type.lower() == EndpointType.CosmosDBContainer.value:
+            endpoint_list = endpoints.cosmos_db_sql_collections
+
+        for endpoint in endpoint_list:
+            if endpoint.name.lower() == endpoint_name.lower():
+                return endpoint
+
+        raise ResourceNotFoundError(f"{endpoint_type} endpoint {endpoint_name} not found in IoT Hub {self.hub_resource.name}.")
```
c-ryan-k: This seems like a useful bit of functionality that could be utilized elsewhere - but this chunk of code requires you to specify an `endpoint_name` - is it possible to extend this to return `endpoint_list` if there's no `endpoint_name` but return `endpoint` if a specific name is provided?

Or is this currently only used as a helper for "select endpoint by type and name"?

vilit1: this is a helper specifically designed for getting the necessary result for update. If anything, I can try to combine it with the show functionality (so if no endpoint type - check all) but would not want to combine this with list

c-ryan-k: Ok, the only thing I'd want to make sure of then (in this case) is that `endpoint_name` is definitely `not None`

azext_iot/iothub/providers/message_endpoint.py
c-ryan-k: Gotcha, yeah looks like in `az iot hub update` we still do call the GET -> PUT even if no parameters have changed.

I don't think this is an issue, especially if it's similar to how other commands currently function.

azext_iot/iothub/providers/message_endpoint.py
```diff
@@ -224,6 +219,174 @@ def create(
...
+            original_endpoint.authentication_type = AuthenticationType.IdentityBased.value
+            if identity == SYSTEM_ASSIGNED_IDENTITY:
+                original_endpoint.identity = None
+            else:
+                original_endpoint.identity = ManagedIdentity(
+                    user_assigned_identity=identity
+                )
+        elif any([connection_string, primary_key, secondary_key]):
+            if original_endpoint.identity:
+                logger.warning(NULL_WARNING.format("The managed identity argument"))
```
c-ryan-k: Would this be the identity `argument` or `property`?

vilit1: should be property

azext_iot/iothub/providers/message_endpoint.py
vilit1: the update command that calls it (and show in a bit) will have an endpoint_name unless there is some logic I missed

vilit1: updated the show

azext_iot/tests/iothub/message_endpoint/test_iothub_message_endpoint_int.py
vilit1: Will say this is ok for now since unit testing exists for this command

vilit1:
Things are more or less working https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=8628&view=results

c-ryan-k (approved):
Test coverage looks good - we might wanna do some pre-release smoke tests to make sure we're not missing a weird endpoint state / input corner case, but so far it looks like you've got everything covered 👍

vilit1:
https://dev.azure.com/azureiotdevxp/aziotcli/_build/results?buildId=8671&view=logs&j=6c5ef33c-c99f-53f8-99b2-716ee366dbc8&t=64c2748c-227f-5b4c-ee99-9c56b6eaac78 tests still pass so will merge

PR #748, opened 2025-07-31
feat: move `az iot hub devicestream` to extension

vilit1:
Moving the devicestream command group to extension so we don't need to worry about mixing api versions in azure-cli

Note that to get this to pop up - you will need to have a version of the azure cli that does not have this command

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/iothub/commands_device_stream.py
+++ b/azext_iot/iothub/commands_device_stream.py
@@ -13,7 +13,6 @@ def show_device_stream(
     hub_name: str,
     resource_group_name: Optional[str] = None,
 ) -> List[dict]:
-    print("in extension")
     return DeviceStreamProvider(
         cmd=cmd, hub_name=hub_name, rg=resource_group_name
     ).show()
--- a/azext_iot/iothub/commands_device_stream.py
+++ b/azext_iot/iothub/commands_device_stream.py
@@ -1,0 +1,19 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azext_iot.iothub.providers.device_stream import DeviceStreamProvider
+from typing import Optional, List
+
+
+def show_device_stream(
+    cmd,
+    hub_name: str,
+    resource_group_name: Optional[str] = None,
+) -> List[dict]:
+    print("in extension")
+    return DeviceStreamProvider(
+        cmd=cmd, hub_name=hub_name, rg=resource_group_name
+    ).show()
--- a/azext_iot/iothub/commands_device_stream.py
+++ b/azext_iot/iothub/commands_device_stream.py
@@ -5,14 +5,14 @@
 # --------------------------------------------------------------------------------------------
 
 from azext_iot.iothub.providers.device_stream import DeviceStreamProvider
-from typing import Optional, List
+from typing import Optional
 
 
 def show_device_stream(
     cmd,
     hub_name: str,
     resource_group_name: Optional[str] = None,
-) -> List[dict]:
+) -> Optional[dict]:
     return DeviceStreamProvider(
         cmd=cmd, hub_name=hub_name, rg=resource_group_name
     ).show()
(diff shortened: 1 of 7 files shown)


azext_iot/iothub/commands_device_stream.py
```diff
@@ -0,0 +1,19 @@
...
+from azext_iot.iothub.providers.device_stream import DeviceStreamProvider
+from typing import Optional, List
+
+
+def show_device_stream(
+    cmd,
+    hub_name: str,
+    resource_group_name: Optional[str] = None,
+) -> List[dict]:
+    print("in extension")
```
vilit1: remove this

digimaun:
> Note that to get this to pop up - you will need to have a version of the azure cli that does not have this command

In prior dynamics of cli extensions, if same command exists in both core and extension the extension command overrides the core version.

azext_iot/iothub/providers/device_stream.py
```diff
@@ -0,0 +1,36 @@
...
+    def __init__(
+        self,
+        cmd,
+        hub_name: str,
+        rg: Optional[str] = None,
+    ):
+        super(DeviceStreamProvider, self).__init__(cmd, hub_name, rg)
+        self.cli = EmbeddedCLI(cli_ctx=self.cmd.cli_ctx)
+
+    def show(self) -> dict:
```
digimaun: As the code is written return type would be Optional[dict] due to `.get("deviceStreams")`

azext_iot/tests/iothub/device_stream/test_iothub_device_stream_int.py
```diff
@@ -0,0 +1,25 @@
...
+# --------------------------------------------------------------------------------------------
+
+import pytest
+from azext_iot.common.embedded_cli import EmbeddedCLI
+
+cli = EmbeddedCLI()
+
+
+@pytest.mark.hub_infrastructure(location="northeurope")
+def test_device_stream(provisioned_only_iot_hubs_module):
```
digimaun: If cli core hub sdk gets incremented to use ga api version, is the goal of this test to assert None == None?

azext_iot/iothub/commands_device_stream.py
```diff
@@ -0,0 +1,18 @@
...
+
+from azext_iot.iothub.providers.device_stream import DeviceStreamProvider
+from typing import Optional, List
+
+
+def show_device_stream(
+    cmd,
+    hub_name: str,
+    resource_group_name: Optional[str] = None,
+) -> List[dict]:
```
digimaun: return type

azext_iot/iothub/providers/device_stream.py
```diff
@@ -0,0 +1,36 @@
...
+        result = self.cli.invoke(
+            f"resource show -n {self.target['name']} -g {self.target['resourcegroup']} "
+            f"--api-version {DEVICE_STREAMS_API_VERSION} --resource-type {HUB_PROVIDER}",
+            capture_stderr=True
```
digimaun, line 31: What is the intended behavior of the command if an API error happens (looking at capture_stderr=True)?

vilit1: This was the old cli - which I do want to rewrite a bit but it would be a bigger task

The capture_stderr=True results in the error just being actually raised rather than just printed out in console. By having it raised, the overall error should present itself better (rather than be a json decode error)

azext_iot/tests/iothub/device_stream/test_iothub_device_stream_int.py
vilit1: yes - I figured having a None test is still good...

I will add a fixture to create this via az resource directly to avoid this issue

azext_iot/iothub/providers/device_stream.py
vilit1: copilot did fight me hard for this implementation

vilit1:
image

azext_iot/iothub/providers/device_stream.py
digimaun: Ok. It's been a long while since I worked on the code base and forgot the capture_stderr dynamic, a more appropriate name for it to me would be something like "raise_stderr".

digimaun (approved):

PR #767, opened 2026-01-07
feat: replace uamqp with PyAMQP

ketkimnaik:
Remove uampq dependency and use pyAmqp from azure-eventhub SDK.

Reference Doc: https://microsoft-my.sharepoint.com/:w:/p/ketkinaik/IQDaBom9nBPTS5pS3WBwZv3lAZj5FHHICZxBnh5LhqD7qAY?e=52Xmfc

Performed the Integration tests locally:
test_iot_messaging_int.py -> done
test_iothub_c2d_messages_int.py -> done
test_iot_central_int.py -> done
test_iot_central_devices_int.py -> 2 failing irrelevant ones

**test_iothub_c2d_messages_int.py => all tests passed**

image

**test_iot_central_devices_int.py => 2 are failing not related to my changes**

image

image

**test_iot_central_int.py => all are passing**

image

image

**test_iot_messaging_int.py => failures were transient. They passed without any changes.**

image

image

image

image

Ran all the commands which are affected due to uampq changes and got the results:

**az iot hub monitor-events -n mytesthub123456 --output table -p all**

image

**az iot device c2d-message send -n mytesthub123456 -d simDevice --data "Test msg" --ack full --output json**

image

image

**az iot central diagnostics monitor-events --app-id 96646c52-78a1-440f-843c-dc8a1f8ae24a --output yaml**

image

**az iot hub monitor-feedback -n mytesthub123456 -d simDevice --output yaml**

image


azext_iot/_help.py
```diff
@@ -83,8 +82,7 @@
     type: command
     short-summary: Monitor feedback sent by devices to acknowledge cloud-to-device (C2D) messages.
     long-summary: |
-                  This command relies on and may install dependent Cython package (uamqp) upon first execution.
-                  https://github.com/Azure/azure-uamqp-python
+                  This command uses PyAMQP for AMQP-based communication.
```
digimaun: It is more accurate to say This command uses azure-eventhub's AMQP implementation.

azext_iot/_help.py
```diff
@@ -34,8 +34,7 @@
     type: command
     short-summary: Monitor device telemetry & messages sent to an IoT Hub.
     long-summary: |
-                  This command relies on and may install dependent Cython package (uamqp) upon first execution.
-                  https://github.com/Azure/azure-uamqp-python
+                  This command uses PyAMQP and azure-eventhub for AMQP-based communication.
```
digimaun: This sentence makes it sound like PyAMQP is a stand-alone released product, which it is not. The reason this implementation detail was here in the first place was because uamqp was a binary distribution and handled as a special case, so it was added as context for the user.

Would either update the sentence to indicate usage of azure-eventhub's AMQP implementation or omit it.

azext_iot/common/_azure.py
```diff
@@ -28,7 +28,18 @@ def _parse_connection_string(cs, validate=None, cstring_type="entity"):
...
+    decomposed = _parse_connection_string(cs, validate, "IoT Hub")
+
+    # Validate SharedAccessKey is valid Base64
+    import base64
+    try:
+        key = decomposed.get("SharedAccessKey")
+        if key:
+            base64.b64decode(key, validate=True)
+    except Exception as e:
+        raise ValueError(f"IoT Hub connection string has invalid SharedAccessKey (not valid Base64): {e}")
```
digimaun: This can be added to _parse_connection_string() to benefit other connection string types.

azext_iot/common/base_discovery.py
```diff
@@ -270,7 +270,7 @@ def get_target(
         """
         cstring = kwargs.get("login")
         if cstring:
-            return self.get_target_by_cstring(connection_string=cstring)
+            return self.get_target_by_cstring(connection_string=cstring, **kwargs)
```
digimaun: What kwargs are being plumbed here ?

azext_iot/iothub/providers/discovery.py
```diff
@@ -41,9 +41,22 @@ def _initialize_client(self):
...
-    @classmethod
-    def get_target_by_cstring(cls, connection_string: str) -> Dict[str, str]:
-        return IotHubTarget.from_connection_string(cstring=connection_string).as_dict()
+    def get_target_by_cstring(self, connection_string: str, **kwargs) -> Dict[str, str]:
+        target = IotHubTarget.from_connection_string(cstring=connection_string).as_dict()
+
+        # When using connection string, provide minimal info needed for _discover_eventhub_endpoint fallback
+        # We need to find the resource to get resourcegroup and subscription for the fallback to work
+        try:
+            resource = self.find_resource(resource_name=target["name"])
```
digimaun: cstring based target construction is intended to be built with no login context, this addition is mixing the scope and does not belong here

azext_iot/monitor/builders/hub_target_builder.py
```diff
@@ -40,72 +36,109 @@ def build_iot_hub_target(self, target):
...
-            result["events"]["address"] = redirect["address"]
-            return redirect, result
-        finally:
-            await receive_client.close_async()
-
     async def _build_iot_hub_target_async(self, target):
-        endpoint = AmqpBuilder.build_iothub_amqp_endpoint_from_target(target)
+        # If events metadata not provided, attempt to discover it
         if "events" not in target:
-            _, link_redirect_events = await self._evaluate_redirect(endpoint)
```
digimaun, line 78: In the existing implementation, the details of the default eventhub backing an iothub could be discovered from an iot hub connection string via amqp link redirect. Here it seems az login is required.

To put another way `az iot hub monitor-events` should work without `az login` through a connection string. Has this been tested?

azext_iot/monitor/builders/hub_target_builder.py
```diff
@@ -40,72 +36,109 @@ def build_iot_hub_target(self, target):
...
+        except Exception as e:
+            raise CLIInternalError(
+                f"Unable to query partitions for '{target['entity'].split('.')[0]}': {e}"
+            )
 
         raise CLIInternalError(
             f"Unable to determine partitions for '{target['entity'].split('.')[0]}'."
         )
+
+    async def _discover_eventhub_endpoint(self, target):
```
digimaun: This functionality is already provided via [IotHubDiscovery](https://github.com/Azure/azure-iot-cli-extension/blob/dev/azext_iot/iothub/providers/discovery.py#L66), why the duplicate functionality ?

azext_iot/tests/utility/test_monitor_parsers_unit.py
```diff
@@ -28,6 +28,57 @@ def _encode_app_props(app_props: dict):
...
+    """
+    Helper function to create EventData object that mimics structure expected by parsers.
+    The parsers expect:
+    - message.system_properties: dict with annotations (device_id, interface, etc.) and content-type/encoding
+    - message.properties: dict with application properties
+    - message.body: bytes or generator of payload
+    - message.get_data(): generator that yields body (for Issue class compatibility)
+    """
+    # Create a mock EventData-like object
+    class MockEventData:
```
digimaun: What is the reason for creating MockEventData here rather than testing with the real objects?

azext_iot/monitor/telemetry.py
```diff
@@ -100,81 +101,90 @@ async def _initiate_event_monitor(
...
+            f"SharedAccessKey={target.key};"
+            f"EntityPath={target.path}"
+        )
+        consumer_client = EventHubConsumerClient.from_connection_string(
+            connection_str,
+            consumer_group=target.consumer_group,
+            eventhub_name=target.path,
+        )
+    elif target.sas_credential:
+        # IoT Central: Use AzureSasCredential (secure token handling)
```
digimaun: What is the reason sas credential is only used for IoT Central?

azext_iot/_help.py
```diff
@@ -34,8 +34,7 @@
     type: command
```
digimaun, line 34: There needs to be a parameter update to deprecate and hide uamqp related parameter(s) where there is at least 1.

azext_iot/monitor/parsers/common_parser.py
```diff
@@ -175,25 +187,56 @@ def _parse_content_type(
...
         try:
-            return unicode_binary_map(message.application_properties)
+            return unicode_binary_map(message.properties or {})
         except Exception:
             details = strings.invalid_application_properties()
             self._add_issue(severity=Severity.warning, details=details)
             return {}
 
-    def _parse_payload(self, message: Message, content_type):
+    def _parse_payload(self, message: EventData, content_type):
```
digimaun, line 216: I see similar code here, in `central_parser` and `issue`. Message parsing can be encapsulated in a single function and re-used across modules.

azext_iot/_help.py
ketkimnaik: Done

azext_iot/_help.py
ketkimnaik: Done.

azext_iot/common/_azure.py
ketkimnaik: Done.

azext_iot/common/base_discovery.py
ketkimnaik: Yes, it is not needed. I tried different approach earlier for that I added but it is not needed now.

azext_iot/iothub/providers/discovery.py
ketkimnaik: Fixed.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: Fixed - implemented AMQP link redirect. Tested working with --login connection string, no az login needed.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: Agree. I removed it. Earlier I added because I was trying multiple approaches.

azext_iot/tests/utility/test_monitor_parsers_unit.py
ketkimnaik: Unit test isolation - we're testing the parser logic, not azure-eventhub's EventData implementation. Because of that I used mock so it will give us control over the data structure without external dependencies.

azext_iot/monitor/telemetry.py
ketkimnaik: IoT Hub uses connection string because:
1. AzureSasCredential with manually generated tokens fails with "CBS Put token error: com.microsoft:auth-failed" 
2. EventHubSharedKeyCredential causes async error: "object AccessToken can't be used in 'await' expression"
3. Connection string approach works reliably - SDK handles async token generation and CBS authentication internally

Both IoT Hub and Central use credential-based authentication, just different patterns required by the azure-eventhub SDK for async context.

azext_iot/_help.py
ketkimnaik: Done

azext_iot/monitor/parsers/common_parser.py
ketkimnaik: Done.

azext_iot/monitor/builders/hub_target_builder.py
```diff
@@ -40,72 +36,185 @@ def build_iot_hub_target(self, target):
...
+                                path = path.split("/", 1)[1] if "/" in path else path
+                            if path.endswith("/$management"):
+                                path = path.replace("/$management", "")
+
+                            result = {
+                                "endpoint": hostname_redirect,
+                                "path": path
+                            }
+                except Exception as e:
+                    import logging
```
digimaun: Import logging at the module level not scoped per exception. Also please following existing patterns of importing logger from knack to maximize compatibility with azure cli

azext_iot/monitor/builders/_common.py
```diff
@@ -4,16 +4,15 @@
...
 # --------------------------------------------------------------------------------------------
 
-import uamqp
 import urllib
-
+from azure.eventhub.aio import EventHubConsumerClient
+from azure.core.credentials import AzureSasCredential
 from azext_iot.monitor.models.target import Target
 
 DEBUG = False
```
digimaun: If DEBUG is not used it can be deleted. In the past it was a dev centric option to expand amqp connection logs.

azext_iot/monitor/builders/_common.py
```diff
@@ -23,47 +22,30 @@ async def convert_token_to_target(tokens) -> Target:
     url = urllib.parse.urlparse(raw_url)
     hostname = url.hostname
 
-    meta_data = await _query_meta_data_internal(hostname, path, sas_token, token_expiry)
+    # Create secure credential object
```
digimaun: nit, this comment and the next ('Get partition count...') don't really add value

azext_iot/monitor/builders/hub_target_builder.py
digimaun: Please confirm this was after an `az logout` to be sure no user account context was involved.

azext_iot/tests/utility/test_monitor_parsers_unit.py
digimaun: The custom MockEventData makes assumptions about the structure (e.g., system_properties, annotations, get_data()) that could drift if EventData's interface changes and we wouldn't catch it until runtime.

Interestingly, the Azure SDK's own unit tests [test_event_data.py](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/eventhub/azure-eventhub/tests/unittest/test_event_data.py) construct real AMQP Message objects and use EventData._from_message(). Maybe not be worth it here.

A smaller improvement: unittest.mock.Mock(spec=EventData) would at least catch accidental access to non-existent attributes that the real EventData doesn't have.

azext_iot/monitor/builders/hub_target_builder.py
digimaun: Can we please test the implementation with free sku hubs as well? With and without --login. Free sku was handled specially in prior implementation.

azext_iot/monitor/builders/hub_target_builder.py
```diff
@@ -40,72 +36,185 @@ def build_iot_hub_target(self, target):
...
+                        partitions=list(amqp_partition_ids),
+                        policy=policy,
+                        key=key
+                    )
+        except Exception as e:
+            raise CLIInternalError(
+                f"Unable to query partitions for '{target['entity'].split('.')[0]}': {e}"
+            )
 
         raise CLIInternalError(
```
digimaun: When does can this later exception be raised? Can client.get_partition_ids() return no partitions?

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: I did az logout and then ran and it worked :)

digimaun: Nice, free sku as well?

ketkimnaik: I tested free sku scenario as well and it is working.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: Done.

azext_iot/monitor/builders/_common.py
ketkimnaik: Done.

azext_iot/monitor/builders/_common.py
ketkimnaik: Yes, removed them.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: The second exception is raised if client.get_partition_ids() succeeds but returns an empty/falsy value (empty list, None). This could theoretically happen with a malformed Event Hub configuration, though the Azure SDK typically throws an exception instead. We could simplify by removing the falsy check since an empty partition list would indicate a broken hub anyway.

azext_iot/tests/utility/test_monitor_parsers_unit.py
ketkimnaik: Good point! Updated to use Mock(spec=EventData) instead of the custom class.

azext_iot/monitor/telemetry.py
```diff
@@ -100,81 +101,95 @@ async def _initiate_event_monitor(
...
+    elif target.sas_credential:
+        # IoT Central: Use pre-generated SAS token credential
+        consumer_client = EventHubConsumerClient(
+            fully_qualified_namespace=target.hostname,
+            eventhub_name=target.path,
+            consumer_group=target.consumer_group,
+            credential=target.sas_credential,
+        )
+    else:
+        raise ValueError("Target must have either (policy, key) or sas_credential")
```
digimaun: This exception is raised to the user correct? If so, what is the expected user action to resolve?

azext_iot/_help.py
```diff
@@ -34,8 +34,7 @@
...
     short-summary: Monitor device telemetry & messages sent to an IoT Hub.
     long-summary: |
-                  This command relies on and may install dependent Cython package (uamqp) upon first execution.
-                  https://github.com/Azure/azure-uamqp-python
+                  This command uses azure-eventhub's AMQP implementation.
 
                   Note: The event will be displayed even if the message body is non-unicode decodable, in
                   this case the event payload portion will be displayed as {{non-decodable payload}} with
```
digimaun, line 40: Double checking if this behavior is retained in this changeset

```
  Note: The event will be displayed even if the message body is non-unicode decodable, in
  this case the event payload portion will be displayed as {{non-decodable payload}} with
  the rest of the event properties that are available.
```

azext_iot/monitor/builders/hub_target_builder.py
digimaun: If the sdk raises a suitable exception, then we don't need to handle it. If there are simplification opportunities where we don't lose something of value we should take it.

azext_iot/monitor/builders/hub_target_builder.py
```diff
@@ -40,72 +36,185 @@ def build_iot_hub_target(self, target):
...
+        """
+        Discover Event Hub endpoint using AMQP link redirect.
+        This allows discovery with just a connection string, without Azure login.
+
+        When connecting to IoT Hub's management endpoint, it redirects to the
+        Event Hub-compatible endpoint, revealing the actual endpoint and path.
+        """
+        def _sync_redirect():
+            try:
+                from azure.eventhub._pyamqp import ReceiveClient as PyAMQPReceiveClient
```
digimaun, line 115: A bit surprising the azure client doesn't have a higher level means of handling link redirect

azext_iot/monitor/telemetry.py
ketkimnaik: This shouldn't be raised to users under normal circumstances - it's a defensive check for an invalid internal state. The Target is constructed by our own builders (hub_target_builder, central_target_builder) which always populate either (policy, key) or sas_credential. If this is raised, it indicates a bug in our target construction logic, not user error. I am changing it to CLIInternalError so user don't have to take any action.

azext_iot/_help.py
ketkimnaik: Yes, this behavior is retained. The _parse_payload method in common_parser.py still uses unicode_decode(data=data, default=NON_DECODABLE_PAYLOAD which displays {non-decodable payload} for non-unicode decodable content, exactly as documented in the help text.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: Yeah, we have to use _pyamqp (internal API). The high-level EventHubConsumerClient doesn't expose redirect handling since Event Hubs don't typically use this pattern. IoT Hub's Event Hub-compatible endpoint uses AMQP link redirects for discovery, which is why we need the lower-level PyAMQP client. This allows connection-string-only discovery without Azure login, matching the prior uAMQP behavior.

azext_iot/monitor/builders/hub_target_builder.py
ketkimnaik: Done.

digimaun (approved):

PR #769, opened 2026-01-21
fix: handle redacted SharedAccessKey in connection string parsing

ketkimnaik:
Fix for sharedaccesskey redacted version.

Ran integration tests and they are passing.

### DIFF
--- a/azext_iot/common/_azure.py
+++ b/azext_iot/common/_azure.py
@@ -23,15 +23,15 @@ def _parse_connection_string(cs, validate=None, cstring_type="entity"):
                         cstring_type, k
                     )
                 )
-
-    # Validate SharedAccessKey is valid Base64 if present
-    import base64
-    try:
-        key = decomposed.get("SharedAccessKey")
-        if key:
-            base64.b64decode(key, validate=True)
-    except Exception as e:
-        raise ValueError(f"{cstring_type} connection string has invalid SharedAccessKey (not valid Base64): {e}")
+        # Validate SharedAccessKey is valid Base64 if present and validation is requested
+        if "SharedAccessKey" in validate:
+            import base64
+            try:
+                key = decomposed.get("SharedAccessKey")
+                if key:
+                    base64.b64decode(key, validate=True)
+            except Exception as e:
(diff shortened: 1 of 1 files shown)


azext_iot/common/_azure.py
```diff
@@ -23,15 +23,17 @@ def _parse_connection_string(cs, validate=None, cstring_type="entity"):
...
-            base64.b64decode(key, validate=True)
-    except Exception as e:
-        raise ValueError(f"{cstring_type} connection string has invalid SharedAccessKey (not valid Base64): {e}")
+        # Validate SharedAccessKey is valid Base64 if present and validation is requested
+        # Skip validation for redacted values (e.g., "****" or "***")
+        if "SharedAccessKey" in validate:
+            import base64
+            key = decomposed.get("SharedAccessKey")
+            # Only validate if key exists and doesn't look like a redacted value
+            if key and not (set(key) <= {'*'}):
```
digimaun: Why only handle this char specially ?

azext_iot/common/_azure.py
```diff
@@ -23,15 +23,17 @@ def _parse_connection_string(cs, validate=None, cstring_type="entity"):
...
-
-    # Validate SharedAccessKey is valid Base64 if present
-    import base64
-    try:
-        key = decomposed.get("SharedAccessKey")
-        if key:
-            base64.b64decode(key, validate=True)
-    except Exception as e:
-        raise ValueError(f"{cstring_type} connection string has invalid SharedAccessKey (not valid Base64): {e}")
+        # Validate SharedAccessKey is valid Base64 if present and validation is requested
```
digimaun: Can we remove this validation for now?

This wasn't a necessary addition to the uamqp replacement work and errors will be raised for invalid SharedAccessKey input.

The direction its going in now with specific char changing behavior is not optimal.

ketkimnaik: I added it as defensive check. I am removing it because it doesn't seem to be manadatory. In the backgroud I ran all the integration tests responsible for c2d commands.

digimaun (approved):

PR #795, opened 2026-03-13
feat: add hostname-type param for IoT Hub

cheatsheet1999:
This PR is expected to meet below requirements:

- For any CLI command which returns a hostname (or a hostname as part of a connection string), add an optional flag –hostnameType which takes in Classic, Device, Service to determine the hostname that CLI should return. 

- Portal/CLI will be able to detect GW version by checking the [gatewayVersion](https://github.com/Azure/azure-rest-api-specs-pr/pull/25736/files) IoT Hub property, and display the new hostnames only for GWv2 hubs.

Sample command to test, with mocked outputs as reference
---
Hub connection-string (allowed: classic, service):

1. `az iot hub connection-string show -n tls13hubv2` (default to show classic host name)
2. `az iot hub connection-string show -n tls13hubv2 --hostname-type classic` - shows "classic" hostname, which is what we have today

3. `az iot hub connection-string show -n tls13hubv2 --hostname-type service` 

---
Device connection-string (allowed: classic, device):

1. `az iot hub device-identity connection-string show -n tls13hubv2 -d test-cs-device`
2. `az iot hub device-identity connection-string show -n tls13hubv2 -d test-cs-device --hostname-type classic`

3. `az iot hub device-identity connection-string show -n tls13hubv2 -d test-cs-device --hostname-type device`, 

---
Module connection-string (allowed: classic, device):

1. `az iot hub module-identity connection-string show -n tls13hubv2 -d test-cs-device -m test-module`
2. `az iot hub module-identity connection-string show -n tls13hubv2 -d test-cs-device -m test-module
  --hostname-type classic` - With old hostname  

3. `az iot hub module-identity connection-string show -n tls13hubv2 -d test-cs-device -m test-module
  --hostname-type device` - With device hostname


cheatsheet1999:
WIP: I'll need to remove direct API call once python-sdk update schema.

azext_iot/_params.py
```diff
@@ -349,6 +350,34 @@ def load_arguments(self, _):
...
             help="Flag indicating the connection string returned is for the default EventHub endpoint. Default: false.",
         )
+        context.argument(
+            "hostname_type",
+            options_list=["--hostname-type", "--ht"],
+            arg_type=get_enum_type(HostnameType),
+            default=HostnameType.service.value,
+            help="Type of hostname to use in the connection string. "
+            "'classic' uses the default hostname, "
+            "'device' uses the TLS 1.3 device hostname, "
```
digimaun: For `az iot hub connection-string show` what is the reason we expose the option for device?

azext_iot/_params.py
```diff
@@ -349,6 +350,34 @@ def load_arguments(self, _):
             options_list=["--default-eventhub", "--eh"],
             help="Flag indicating the connection string returned is for the default EventHub endpoint. Default: false.",
         )
+        context.argument(
+            "hostname_type",
+            options_list=["--hostname-type", "--ht"],
+            arg_type=get_enum_type(HostnameType),
+            default=HostnameType.service.value,
+            help="Type of hostname to use in the connection string. "
+            "'classic' uses the default hostname, "
```
digimaun, line 359: If I configure my hub for TLS 1.3, is the classic hostname still supported? If so, what is the use case for that (why would I want to use the non tls1.3 hostname?

azext_iot/_params.py
cheatsheet1999: The spec says " --hostnameType {Classic|Device|Service} to any CLI command that returns a hostname or connection string." 

I followed that consistently across all 3 connection string commands. Happy to restrict this command to just Classic and Service if the you feel Device isn't applicable here. Thanks!

azext_iot/_params.py
cheatsheet1999: Yes, the classic hostname is still supported. I agree with you that if a hub is configured for TLS1.3, there's no strong use case for classic hostname-type. I believe classic is there because we need backward compatability. The spec states "Rollback supported by re-setting endpoint type." If we want to remove Classic as an option, we should clarify with the PM.

azext_iot/_params.py
digimaun: Ok, I'm not sure what spec you are referring to, but we need to think through the scenarios and use cases and see what makes sense for the commands.

azext_iot/_params.py
digimaun: The thought is that `az iot hub connection-string show` should automatically choose the proper hostname based on whether TLS 1.3 is enabled or not. This should work if there is no compelling reason the classic hostname should be used when TLS 1.3 is enabled.

azext_iot/_params.py
cheatsheet1999: I was referring to [this](https://microsoft-my.sharepoint-df.com/:w:/p/jesusbar/cQqcNyLl8mdFSbbVm4vmDqFqEgUCpadFwq3InKMz5QZn6EhDgw?isSPOFile=1&ovuser=72f988bf-86f1-41af-91ab-2d7cd011db47%2Cyuelinzhao%40microsoft.com&wdExp=TEAMS-TREATMENT&web=1&clickparams=eyJBcHBOYW1lIjoiVGVhbXMtRGVza3RvcCIsIkFwcFZlcnNpb24iOiI0OS8yNjAyMjcwNDIwOSJ9) spec. 

Section **1.1.2** mentiones _"For any CLI command that returns a hostname or connection string, add: --hostnameType {Classic|Device|Service}"_ 

Thank you for your review!

azext_iot/_params.py
cheatsheet1999: I agree that's a cleaner design. 

On the IoT CLI side we can definitely do that, auto-detect whether the hub supports TLS 1.3 (by checking if the service/device hostnames are available) and use them by default, falling back to classic for GWv1 hubs. 

The --hostname-type flag would remain as an optional override for users who need a specific hostname. 

Should we propose this design instead?

digimaun: Optional override if the classic hostname is still valid after configuring TLS 1.3, made a comment about it in the doc you linked

zzcodework: I believe we can't remove the Classic endpoint because there're order and constrained devices which don't support TLS 1.3, +1 on default to TLS 1.3 endpoint and add optional override.

azext_iot/_params.py
zzcodework: How many commands need to be updated because of the gateway version factor introduced?

azext_iot/_params.py
cheatsheet1999: My current implementation defaults to the TLS 1.3 endpoint (service for service-side commands, device for device provisioning commands) and keeps --hostname-type classic as an optional override for backward compatibility with constrained devices. This aligns with both your feedback. Thanks for the review.

azext_iot/_params.py
digimaun: As you know we have dedicated commands for obtaining hub/device/module connection string. At first glance, it doesn't seem appropriate to expose the new device endpoint for `az iot hub connection-string show` because its scope is not devices. And vice versa why expose the new service endpoint for `az iot hub device-identity connection-string show` (or module cstring)?

cheatsheet1999: I had the same feeling when I started implementing this, so I see what you mean. I’ll update it. Thanks Paymaun.

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
     return SasTokenAuthentication(uri, policy, key, duration)
 
 
-def _build_device_or_module_connection_string(entity, key_type="primary"):
+def _fetch_tls_hostnames(cmd, hub_name, resource_group_name, hub_location=None):
```
cheatsheet1999, line 2399: Hi @digimaun, given that the control plane SDK won’t be ready until the public review, I’d really appreciate your guidance on the best approach for now. Currently, we’re using a direct API call here.

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
...
+        mgmt_host = f"{hub_location}.management.azure.com"
+
+    url = (
+        f"https://{mgmt_host}/subscriptions/{sub_id}"
+        f"/resourceGroups/{resource_group_name}"
+        f"/providers/Microsoft.Devices/IotHubs/{hub_name}"
+    )
+    req = HttpRequest(method="GET", url=url, params={"api-version": "2026-03-01-preview"})
+    resp = client.iot_hub_resource._client.send_request(req)
+    data = resp.json()
```
HangyiWang: just a question: would `.json()` raise exception if not 200?

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
...
+    mgmt_host = "management.azure.com"
+    if hub_location and "euap" in hub_location.lower():
+        mgmt_host = f"{hub_location}.management.azure.com"
+
+    url = (
+        f"https://{mgmt_host}/subscriptions/{sub_id}"
+        f"/resourceGroups/{resource_group_name}"
+        f"/providers/Microsoft.Devices/IotHubs/{hub_name}"
+    )
+    req = HttpRequest(method="GET", url=url, params={"api-version": "2026-03-01-preview"})
```
HangyiWang: probably should put `2026-03-01-preview` as a constant somewhere

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
...
+    }
+
+
+def _resolve_hostname_by_type(cmd, hub, hostname_type):
+    """Resolve the appropriate hostname based on the hostname type."""
+    if hostname_type == HostnameType.classic.value:
+        return hub.properties.host_name
+
+    tls_info = _fetch_tls_hostnames(
+        cmd, hub.name, hub.additional_properties.get("resourcegroup"), hub.location
```
HangyiWang: for `hub.additional_properties.get("resourcegroup")`, would `resourcegroup` be case sensitive?

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
...
+    from azure.cli.core.commands.client_factory import get_mgmt_service_client
+    from azure.cli.core.profiles import ResourceType
+    from azure.core.rest import HttpRequest
+
+    client = get_mgmt_service_client(cmd.cli_ctx, ResourceType.MGMT_IOTHUB)
+    sub_id = client._config.subscription_id
+
+    # Use regional management endpoint for Canary regions, otherwise standard
+    mgmt_host = "management.azure.com"
+    if hub_location and "euap" in hub_location.lower():
```
HangyiWang: nit: could be safer to use `hub_location.lower().endswith("euap")` or just hardcode 2 canary region names somewhere to compare them

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2395,59 @@ def _iot_build_sas_token(
...
+    """Resolve the appropriate hostname based on the hostname type."""
+    if hostname_type == HostnameType.classic.value:
+        return hub.properties.host_name
+
+    tls_info = _fetch_tls_hostnames(
+        cmd, hub.name, hub.additional_properties.get("resourcegroup"), hub.location
+    )
+
+    if tls_info.get("gatewayVersion") != "V2":
+        logger.warning(
```
HangyiWang: is this a user facing error or just internal debug warning?

azext_iot/operations/hub.py
cheatsheet1999: `.json()` doesn't raise exception on non-200. However, the behavior is still correct - because if the API returns an error, props.get("deviceHostName") returns None and the caller falls back to classic hostname.

But this is a valid question, I can add an explicit status check, like 
```
if resp.status_code >= 400:
        return {"deviceHostName": None, "serviceHostName": None, "gatewayVersion": None}
```
 for clarity if you prefer.

azext_iot/operations/hub.py
cheatsheet1999: I agree, thanks!

azext_iot/operations/hub.py
cheatsheet1999: Nope. The SDK returns lowercase "resourcegroup". This is also consistent with how the rest of the codebase uses it.

azext_iot/operations/hub.py
cheatsheet1999: Good point, I agree! Will go ahead with 
```
if hub_location and hub_location.lower().endswith("euap"):
```

azext_iot/operations/hub.py
cheatsheet1999: It's user facing. I intentionally warn the user that their hub doesn't support TLS 1.3 endpoints and fall back to classic. 

This way it's non-breaking - existing GWv1 hubs still work, users just see a warning.

Any concerns here?

azext_iot/operations/hub.py
HangyiWang: no just go with what you think is best, I just had a concern about the untracked exception

azext_iot/operations/hub.py
HangyiWang: cool

azext_iot/operations/hub.py
HangyiWang: no concern — just thinking about what info we should show users, like action items or errors, and what they should do next etc. all good.

azext_iot/operations/hub.py
cheatsheet1999: Cool, I think it's better to have a check here. Will update, thanks.

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2396,61 @@ def _iot_build_sas_token(
...
+            "IoT Hub '%s' is not a GWv2 hub. Falling back to classic hostname. "
+            "The 'device' and 'service' hostname types are only available on GWv2 IoT Hubs.",
+            hub.name,
+        )
+        return hub.properties.host_name
+
+    hostname_key = "deviceHostName" if hostname_type == HostnameType.device.value else "serviceHostName"
```
HangyiWang: just clarify - so by default we're using `serviceHostName` as key?

HangyiWang:
LGTM thanks! @cheatsheet1999 next time you can cleanup the PR description default template if that's no applied to you and you can add more descriptive info in that area.

azext_iot/operations/hub.py
cheatsheet1999: Yes, thanks for the call out, this is expected.

--- update to clarify:
This line only runs when hostname_type is **not** classic (classic returns early at line 2434). So by the time we reach here, it's either device or service:

So I think it's not "defaulting to service", it's just the else branch for the only other option. The classic case never reaches this line.

azext_iot/operations/hub.py
cheatsheet1999: I’m leaning toward using the direct API call approach for the private review, then focusing on the SDK refactor. Once the SDK work is complete and the Hub/DPS APIs are deployed to all regions, I would switch from the direct API calls to the SDK approach for public preview. 

What do you think? @digimaun

azext_iot/operations/hub.py
```diff
@@ -2443,8 +2501,13 @@ def iot_get_device_connection_string(
         login=login,
         auth_type_dataplane=auth_type_dataplane,
     )
+    hostname_override = None
+    if hostname_type != HostnameType.classic.value:
+        discovery = IotHubDiscovery(cmd)
+        hub = discovery.find_resource(hub_name_or_hostname, resource_group_name)
```
digimaun: This is a redundant usage of discovery.find_resource() since iot_device_show() already uses it via discovery.get_target().

digimaun: Then `_resolve_hostname_by_type` calls `_fetch_tls_hostnames` which makes the call to get the hub resource with the new preview API ... this can/should be done more efficiently.

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2396,61 @@ def _iot_build_sas_token(
...
+    if hub_location and hub_location.lower().endswith("euap"):
+        mgmt_host = f"{hub_location}.management.azure.com"
+
+    url = (
+        f"https://{mgmt_host}/subscriptions/{sub_id}"
+        f"/resourceGroups/{resource_group_name}"
+        f"/providers/Microsoft.Devices/IotHubs/{hub_name}"
+    )
+    req = HttpRequest(method="GET", url=url, params={"api-version": IOTHUB_PREVIEW_API_VERSION})
+    resp = client.iot_hub_resource._client.send_request(req)
```
digimaun: A downside of calling the private _client.send_request is likely missing built-in retries, error handling and pipeline policies.

digimaun: If logging policy isn't configured may show bearer token in --debug logs. I would validate whether this is the case.

digimaun: If `_client.send_request()` raises an exception (network errors, whatever else) that would raise a stack trace to the user ... right?

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2396,61 @@ def _iot_build_sas_token(
...
+    }
+
+
+def _resolve_hostname_by_type(cmd, hub, hostname_type):
+    """Resolve the appropriate hostname based on the hostname type."""
+    if hostname_type == HostnameType.classic.value:
+        return hub.properties.host_name
+
+    tls_info = _fetch_tls_hostnames(
+        cmd, hub.name, hub.additional_properties.get("resourcegroup"), hub.location
```
digimaun: `hub.additional_properties.get("resourcegroup")` implies it can return None, but `_fetch_tls_hostnames()` doesn't account for resource_group_name=None ... it plugs it into the URL as is. If 'resourcegroup' is always available and not None it could be directly accessed.

azext_iot/operations/hub.py
cheatsheet1999: Thanks. For the second comment, ﻿I Validated, ran the command with --debug and confirmed the Authorization header is redacted (REDACTED /*****) in the output, ﻿so no token leakage.
 
For the first one, retries and error handling - agreed, I've wrapped the call in try/except so failures gracefully fall back to classic hostname with a warning, rather than showing a stack trace to the user. Moving forward, I'll change to SDK approach once that part of work is done

cheatsheet1999: Thanks for catching this issue - I ﻿wrapped the call in try/except. On any exception (network errors, timeouts, etc.), it logs a warning and gracefully falls back to classic hostname rather than surfacing a stack trace to the user.

azext_iot/operations/hub.py
cheatsheet1999: Thanks for catching this - I overlooked it. I’ll change it to use direct access. At this point, the hub has already been resolved

azext_iot/operations/hub.py
cheatsheet1999: Thanks for pointing this out, ﻿I  refactored to eliminate the redundant find_resource(). Now we call discovery.get_target() once at the top of the function, then reuse the target dict.

﻿This reduces the calls from 3 (get_target + find_resource + preview API) to 2 (get_target + preview API). The preview API call is still needed since the standard API doesn't return TLS 1.3 hostnames.

azext_iot/operations/hub.py
```diff
@@ -2458,19 +2524,25 @@ def iot_get_module_connection_string(
...
         login=login,
-        auth_type_dataplane=auth_type_dataplane,
+        auth_type=auth_type_dataplane,
     )
+    module = _iot_device_module_show(target, device_id, module_id)
+    hostname_override = None
+    if hostname_type != HostnameType.classic.value:
+        hostname_override = _resolve_hostname_by_type(
+            cmd, target["entity"], target["name"], target["resourcegroup"],
+            hostname_type, target.get("location")
```
digimaun: does _resolve_hostname_by_type behaves well without and with --login ?

digimaun: curious if you try --login when hostname_type != classic  , when using --login  `az logout` first

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2396,63 @@ def _iot_build_sas_token(
...
+        data = resp.json()
+        props = data.get("properties", {})
+        gateway_version = props.get("iotHubDetails", {}).get("gatewayVersion")
+        return {
+            "deviceHostName": props.get("deviceHostName"),
+            "serviceHostName": props.get("serviceHostName"),
+            "gatewayVersion": gateway_version,
+        }
+    except Exception as e:
+        logger.warning("Failed to fetch TLS 1.3 hostnames for '%s': %s", hub_name, str(e))
```
digimaun, line 2429: Is warning most appropriate? If a user does `--hostname-type device` and it doesn't work isn't an error more accurate?

azext_iot/operations/hub.py
cheatsheet1999: ﻿without --login (normal -n/-g): Works correctly, resolves to device/service, hostname as expected.

with ﻿--login + --hostname-type device: The --login path builds the target from the connection string directly, which doesn't include resourcegroup. This crashes with KeyError: 'resourcegroup

Thanks for catching this issue, I didn't think about this scanario. Let me figure out something here.

azext_iot/operations/hub.py
cheatsheet1999: I agree. ﻿ 

I was thinking about graceful degradation, like, if something fails, still give the user a result rather than crashing. But that's the wrong mindset here.

cheatsheet1999: ﻿ 1. _fetch_tls_hostnames (internal, called in _build_target for transport): keeps the warning + returns empty. CLI falls back to classic for transport, the command still works. 
﻿ 
2. _resolve_hostname_by_type (user-facing, called when --hostname-type is non-classic): raises InvalidArgumentValueError if the requested TLS hostname is unavailable. User gets a clear error

So: warning for internal graceful degradation, error for explicit user requests that can't be fulfilled.

azext_iot/operations/hub.py
```diff
@@ -2394,7 +2396,66 @@ def _iot_build_sas_token(
...
+        return empty
+
+
+def _transform_hostname(hostname, hostname_type):
+    """Transform a hostname to the requested type via string manipulation."""
+    hub_name = hostname.split(".")[0]
+    hostname_map = {
+        HostnameType.classic.value: f"{hub_name}.azure-devices.net",
```
digimaun: This is not going to work for government clouds if/when they different domains. Consider extracting the domain suffix from the original hostname rather than hardcoding it.

azext_iot/iothub/providers/discovery.py
```diff
@@ -72,22 +72,33 @@ def _build_target(
         # but that will be better served aligning with vNext pattern for Iot Hub
 
         include_events = kwargs.get("include_events", False)
+        classic_hostname = resource.properties.host_name
+        rg = getattr(resource, 'resourcegroup', None) or resource.additional_properties.get("resourcegroup")
+
+        # TODO: Remove _fetch_tls_hostnames once SDK uses 2026-03-01-preview —
+        # resource will have serviceHostName natively from client.get().
+        from azext_iot.operations.hub import _fetch_tls_hostnames
+        tls_info = _fetch_tls_hostnames(self.cmd, resource.name, rg, resource.location)
```
digimaun, line 81: _build_target is a common handler for many operations...this change will introduce an additional ARM call to the preview API/performance hit for all those operations.

azext_iot/tests/iothub/tls13/test_tls13_unit.py
```diff
@@ -0,0 +1,142 @@
...
+        from azext_iot.operations.hub import _resolve_hostname_by_type
+        target = self._make_target(service_hostname="testhub.service.azure-devices.net")
+        result = _resolve_hostname_by_type(target, HostnameType.service.value)
+        assert result == "testhub.service.azure-devices.net"
+
+    def test_missing_device_hostname_raises_error(self):
+        import pytest
+        from azext_iot.operations.hub import _resolve_hostname_by_type
+        target = self._make_target()
+        with pytest.raises(Exception):
```
digimaun: `pytest.raises(InvalidArgumentValueError)` ?

azext_iot/common/shared.py
```diff
@@ -75,6 +75,16 @@ class KeyType(Enum):
     secondary = "secondary"
 
 
+class HostnameType(Enum):
+    """
+    Type of hostname to use in connection strings.
+    """
+
+    classic = "classic"
```
digimaun: The member names should be all caps moving forward

CLASSIC, DEVICE, SERVICE -- we are treating them like constants (of course the values can be whatever casing they need to be.

azext_iot/operations/hub.py
cheatsheet1999: Thanks for letting me know.

azext_iot/iothub/providers/discovery.py
cheatsheet1999: I understand that, ﻿you're right that this adds an extra ARM call for all operations going through _build_target.

﻿This is needed to meet a requirement 
image


we need to resolve the service hostname for transport on every dataplane command. Once the SDK moves to
 2026-03-01-preview, client.get() will return serviceHostName natively and this extra call goes away entirely
 (tracked via the TODO).

Is this okay as a temporary approach for the private preview?

digimaun: alright, please make sure this is tracked/won't get lost

cheatsheet1999: I'll track it, this will be fixed/removed before public review.

azext_iot/tests/iothub/tls13/test_tls13_unit.py
cheatsheet1999: Thanks for catching this.

digimaun (approved):

PR #797, opened 2026-03-16
chore: remove LRO workarounds for policy revokeIssuer and activateBYOR

HangyiWang:
The `revoke_issuer` and `activate_byor` methods previously contained try/except workarounds to handle a backend issue where the ADR returned 200 OK responses without a status field in the LRO operation status body. 
This caused the Azure SDK's ARMPolling to fail with OperationFailed, even though the operation had succeeded. 
The ADR backend has since been fixed — the operation status endpoint now returns `{"status": "Succeeded", "percentComplete": 100.0, ...}` in the final 200 response, which ARMPolling handles correctly. 
image

### DIFF
--- a/azext_iot/adr/providers/policy.py
+++ b/azext_iot/adr/providers/policy.py
@@ -8,7 +8,6 @@ from typing import Dict, Optional
 
 from azure.cli.core.azclierror import AzureResponseError, ResourceNotFoundError
 from azure.core.exceptions import HttpResponseError
-from knack.log import get_logger
 from rich.console import Console
 
 from azext_iot.adr.common import (
@@ -20,7 +19,6 @@ from azext_iot.adr.providers.base import ADRProvider
 from azext_iot.common.utility import wait_for_terminal_state
 
 console = Console()
-logger = get_logger(__name__)
 
 
 class PolicyProvider(ADRProvider):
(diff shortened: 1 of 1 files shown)


digimaun (approved):

cheatsheet1999 (approved):

zzcodework (approved):

azext_iot/adr/providers/policy.py
```diff
@@ -171,21 +169,7 @@ def revoke_issuer(self, policy_name: str, namespace_name: str, resource_group_na
                 namespace_name=namespace_name,
                 policy_name=policy_name,
             )
-            try:
-                wait_for_terminal_state(poller, **kwargs)
-            except HttpResponseError as e:
-                # The backend returns 200 OK with an empty body when the LRO completes,
-                # but ARMPolling expects a "status" or "provisioningState" field in the
```
zzcodework, line 178: I assume the new api version added the status field and fixed this backend service issue, but curious how the previous api version was able to pass review and checks, since the status field is a standard and required (i assume) property?

HangyiWang: The API version didn't change. It's still `2026-03-01-preview`. The backend team fixed the response on their service side — they started including `["status": "Succeeded"]` in the operation status response without bumping the API version. This is a bug fix to conform to an existing contract, not a new API version.

The ARM LRO status field requirement is enforced at runtime by the ARM polling client (azure-core SDK), not at API spec review time. The API spec (Swagger) likely defined the operation status response correctly, but the actual service implementation didn't populate the status field. API spec reviews check the spec, not the runtime behavior — so the gap between spec and implementation went unnoticed.

For resource operations ARM LRO polling responses on the `/operationStatuses/` endpoint, the validation is less strict — the response goes through to the client, and it's the client-side ARMPolling that fails when it can't find status field.

HangyiWang: Issue clarifications: when the CLI called `revokeIssuer` or `activateBYOR` on the ADR backend, the operation itself succeeded, but the LRO operation status endpoint (GET `.../operationStatuses/...`) returned a 200 OK response without a "status" field in the body. The Azure SDK's ARMPolling client expects `{"status": "Succeeded"}` in the 200 body to know the operation is done — without it, it throws `OperationFailed` even though nothing actually failed.

Now ADR backend put the `status` in the response.

PR #799, opened 2026-03-23
fix: send non-nullable disable property in device revoke request

HangyiWang:
**Problem:**
The [Swagger spec](https://github.com/Azure/azure-rest-api-specs/blob/e10841fe5bd4efff452e04ca5f4a2685161b4a2d/specification/deviceregistry/resource-manager/Microsoft.DeviceRegistry/preview/2026-03-01-preview/deviceregistry.json#L5867 ) defines `DeviceCredentialsRevokeRequest.disable` as a non-nullable `boolean` (not required). When the `--disable` flag was omitted by the user, the CLI sent `{"disable": null}` in the request body. 

While the ADR backend accepted the initial POST (202), the `null` value caused the long-running operation to fail during polling with:

```
HttpResponsePayloadAPISpecValidationFailed: Missing required property: result. Paths in payload: '$.result'
```
**Fix:**
Only include `disable` in the request body when explicitly set by the user. When `--disable` is omitted, send `{}` instead of `{"disable": null}`.

The ADR backend also should validate the request payload against the [Swagger spec](https://github.com/Azure/azure-rest-api-specs/blob/e10841fe5bd4efff452e04ca5f4a2685161b4a2d/specification/deviceregistry/resource-manager/Microsoft.DeviceRegistry/preview/2026-03-01-preview/deviceregistry.json#L5867 ) upfront and reject `disable: null` with a clear 400 error on the initial POST, rather than accepting the request and failing silently during the async operation.

### DIFF
--- a/azext_iot/adr/providers/device.py
+++ b/azext_iot/adr/providers/device.py
@@ -86,13 +86,13 @@ class DeviceProvider(ADRProvider):
         device_name: str,
         namespace_name: str,
         resource_group_name: str,
-        disable: Optional[bool] = None,
+        disable: bool = False,
         **kwargs,
     ):
         """Revoke credentials for a device in the namespace."""
         body = {}
-        if disable is not None:
-            body["disable"] = disable
+        if disable:
+            body["disable"] = True
         with console.status(
             f"Revoking credentials for device '{device_name}' in namespace {namespace_name}..."
         ):
--- a/azext_iot/adr/providers/device.py
+++ b/azext_iot/adr/providers/device.py
@@ -90,6 +90,9 @@ class DeviceProvider(ADRProvider):
         **kwargs,
     ):
         """Revoke credentials for a device in the namespace."""
+        body = {}
+        if disable is not None:
+            body["disable"] = disable
         with console.status(
             f"Revoking credentials for device '{device_name}' in namespace {namespace_name}..."
         ):
@@ -97,6 +100,6 @@ class DeviceProvider(ADRProvider):
                 resource_group_name=resource_group_name,
                 namespace_name=namespace_name,
                 device_name=device_name,
-                body={"disable": disable},
+                body=body,
             )
             return wait_for_terminal_state(poller, **kwargs)
--- a/azext_iot/adr/commands_device.py
+++ b/azext_iot/adr/commands_device.py
@@ -55,7 +55,7 @@ def adr_device_revoke(
     device_name: str,
     namespace_name: str,
     resource_group_name: str,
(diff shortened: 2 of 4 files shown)


azext_iot/adr/providers/device.py
```diff
@@ -90,13 +90,16 @@ def revoke(
         **kwargs,
     ):
         """Revoke credentials for a device in the namespace."""
+        body = {}
+        if disable is not None:
```
digimaun: if disable is a flag, this can be a truthy condition to be more pythonic

```
if disable:
  body["disable"] = disable
```
or 
```
if disable:
  body["disable"] = True
```

cheatsheet1999 (approved):
Nice! This makes sense given the swagger defines `disable` as non-nullable. Looks good.

azext_iot/adr/providers/device.py
cheatsheet1999: Also wanted to check on this, would `disable=False` still be sent with this approach?

cheatsheet1999: Got it. Just thinking out loud: since `disable=False` is a valid value for the API, would we want to still send `"disable": false` in that case,  instead of omitting the field?

Otherwise it seems we wouldn't be able to distinguish between "not provided" and "explicitly false".

digimaun: You are right, leave it as is.

HangyiWang: @cheatsheet1999  So `disable=false` should be sent in the device update command or explicit device disable during revoke, not device revoke by default. According to the Swagger spec, it’s optional, so ideally no request body should be sent when invoking the device revoke command.

HangyiWang: @digimaun Let me update it. We shouldn't keep 3 values including `None` for the disable flag.

Elsie4ever (approved):

azext_iot/adr/providers/device.py
HangyiWang: @digimaun changed to pythonic way. The --disable flag is either present (sends {"disable": true}) or absent (sends {}). No false value is ever sent, which aligns with the Swagger — disable is a boolean property in the request body, and we only include it when the user wants to disable.

digimaun (approved):

azext_iot/adr/providers/device.py
zzcodework: ADR API spec allows nullable, but their service code is having issues if the property is omitted, as a workaround, the CLI must always send a value.

PR #802, opened 2026-03-27
chore: adjust CMS device integration test

HangyiWang:
- Switch test region from canary to Prod
- Change `--attributes` parameter type from `Dict` to `str` to support clearing with empty string
- Fix device integration test assertions to access fields under ARM `properties` envelope
- Run revoke before clear-policy to avoid re-assignment
- Remove version/thumbprint assertions from revoke (tested; not changed by backend)

image

### DIFF
--- a/azext_iot/adr/providers/device.py
+++ b/azext_iot/adr/providers/device.py
@@ -57,10 +57,10 @@ class DeviceProvider(ADRProvider):
         if attributes is not None:
             if isinstance(attributes, str):
                 if attributes == "":
-                    parsed_attributes = None
+                    attributes = None
                 else:
-                    parsed_attributes = cast(Dict[str, Any], shell_safe_json_parse(attributes))
-            inner_props["attributes"] = parsed_attributes
+                    attributes = shell_safe_json_parse(attributes)
+            inner_props["attributes"] = attributes
         if policy_resource_id is not None:
             if policy_resource_id == "":
                 inner_props["policy"] = None
--- a/azext_iot/adr/providers/device.py
+++ b/azext_iot/adr/providers/device.py
@@ -4,7 +4,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-from typing import Any, Dict, Optional, cast
+from typing import Dict, Optional
 
 from knack.log import get_logger
 from rich.console import Console
--- a/azext_iot/adr/providers/device.py
+++ b/azext_iot/adr/providers/device.py
@@ -4,7 +4,7 @@
 # Licensed under the MIT License. See License.txt in the project root for license information.
 # --------------------------------------------------------------------------------------------
 
-from typing import Dict, Optional
+from typing import Any, Dict, Optional, cast
 
 from knack.log import get_logger
 from rich.console import Console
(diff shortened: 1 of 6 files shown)


azext_iot/adr/providers/device.py
digimaun: I guess here its Optional[str] as well, but then we can add a type annotation on the assignment from `shell_safe_json_parse` to Dict[str, any] e.g.  on line 62

`attributes: Dict[str, any] = shell_safe_json_parse(attributes)`

HangyiWang: Sure fixed.

azext_iot/adr/providers/device.py
```diff
@@ -56,7 +56,10 @@ def update(
             inner_props["operatingSystemVersion"] = operating_system_version
         if attributes is not None:
             if isinstance(attributes, str):
-                attributes = shell_safe_json_parse(attributes)
+                if attributes == "":
+                    attributes = None
+                else:
+                    attributes = shell_safe_json_parse(attributes)
```
digimaun, line 62: The type annotation got reverted here? Alternatively/in addition shell_safe_json_parse could have the proper return annotation

azext_iot/tests/adr/test_adr_device_int.py
```diff
@@ -298,85 +266,44 @@ def revoke_and_check(revoke_args, expected_enabled, label=""):
...
-                        assert d.get("enabled") is expected_enabled, (
+                        assert dp.get("enabled") is expected_enabled, (
                             f"[{call_label}] expected enabled={expected_enabled}, "
-                            f"got {d.get('enabled')}"
-                        )
-
-                    cur_version = d.get("version")
-                    cur_transition = d.get("lastTransitionTime")
-                    if prev_version is not None and cur_version is not None:
-                        assert cur_version > prev_version, (
```
digimaun, line 311: This seems like a good assertion, why did we need to drop it ?

azext_iot/tests/adr/test_adr_device_int.py
```diff
@@ -298,85 +266,44 @@ def revoke_and_check(revoke_args, expected_enabled, label=""):
...
-                            hub_dev.get("authentication", {})
-                            .get("x509Thumbprint", {})
-                            .get("primaryThumbprint")
-                        )
-                        if new_hub_thumbprint != prev_hub_thumbprint:
-                            break
-                        if attempt < max_attempts:
-                            _log(
-                                LogKind.WARN,
-                                "[%s] Hub thumbprint unchanged (attempt %d/%d), "
```
digimaun, line 345: We not tracking if the device thumbprint changes anymore?

azext_iot/tests/adr/test_adr_device_int.py
HangyiWang: I think so but the `version` and `lastTransitionTime` would be unchanged somehow from backend this time. They remain the same after the revoke.

azext_iot/tests/adr/test_adr_device_int.py
HangyiWang: Yeah we can't - the thumbprint is unchanged now from my testing.  Probably need to report to backend.

peolivei2 (approved):

azext_iot/adr/providers/device.py
HangyiWang: Image pylance complains the annotation and the shell_safe_json_parse can only have `Any` return annotation IMO.

digimaun: What about

```
from typing import cast

attributes = cast(Dict[str, Any], shell_safe_json_parse(attributes))
```

HangyiWang: sure but need a separate variable for the type.

digimaun (approved):

digimaun (approved):

digimaun (approved):

PR #805, opened 2026-03-30
fix(monitor): wire HTTP_PROXY/HTTPS_PROXY env vars to EventHub consumer client

ketkimnaik:
1. Added proxy support — monitor-events now reads HTTPS_PROXY/HTTP_PROXY env vars and passes them to the EventHub consumer client, consistent with other commands in the extension.

2. Fixed AMQP WebSocket transport bug — when a proxy is configured, automatically switches to AmqpOverWebsocket and correctly formats proxy_hostname with the URL scheme (e.g. http://host), fixing the EventHubError('host:port\nhost:port') bug in pyamqp.

3. Added aiohttp dependency + login discovery fallback — aiohttp is required for async WebSocket transport; also improved --login target discovery to hydrate Event Hub metadata from ARM when the connection string path lacks it.

Testing:
Successfully ran integration test, unit tests and live commands.

image

---

### General Guidelines

Intent for Production

Basic expectations

### DIFF
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -432,6 +432,16 @@ def load_arguments(self, _):
             help="Number of telemetry messages to capture before the monitor is terminated. "
             "If not specified, monitor keeps running until meeting the timeout threshold of not receiving messages from hub.",
         )
+        context.argument(
+            "transport",
+            options_list=["--transport", "--tr"],
+            arg_type=get_enum_type(["amqp", "amqp_ws"]),
+            help="Underlying transport protocol for the Event Hub client. "
+            "'amqp_ws' (AMQP over WebSocket) is required when routing through an HTTP proxy "
+            "and may also be preferred in environments where port 5671 is blocked. "
+            "Defaults to 'amqp'. When a proxy is configured via HTTPS_PROXY or HTTP_PROXY "
+            "environment variables, 'amqp_ws' is used automatically.",
+        )
 
     with self.argument_context("iot hub monitor-feedback") as context:
         context.argument(
--- a/azext_iot/_params.py
+++ b/azext_iot/_params.py
@@ -36,6 +36,7 @@ from azext_iot.common.shared import (
 )
 from azext_iot._validators import mode2_iot_login_handler, process_top
 from azext_iot.assets.user_messages import info_param_properties_device
+from azext_iot.monitor.models.enum import Transport
 
 
 dps_auth_type_dataplane_param_type = CLIArgumentType(
@@ -435,7 +436,7 @@ def load_arguments(self, _):
         context.argument(
             "transport",
             options_list=["--transport", "--tr"],
-            arg_type=get_enum_type(["amqp", "amqp_ws"]),
+            arg_type=get_enum_type(Transport),
             help="Underlying transport protocol for the Event Hub client. "
             "'amqp_ws' (AMQP over WebSocket) is required when routing through an HTTP proxy "
             "and may also be preferred in environments where port 5671 is blocked. "
--- a/azext_iot/central/providers/monitor_provider.py
+++ b/azext_iot/central/providers/monitor_provider.py
@@ -31,6 +31,7 @@ class MonitorProvider:
         consumer_group: str,
         central_handler_args: CentralHandlerArguments,
         central_dns_suffix: str,
+        transport: str = None,
     ):
         central_device_provider = CentralDeviceProvider(
             cmd=cmd, app_id=app_id, token=token, api_version=ApiVersion.ga.value
@@ -44,6 +45,7 @@ class MonitorProvider:
             token=token,
             consumer_group=consumer_group,
             central_dns_suffix=central_dns_suffix,
+            transport=transport,
         )
         self._handler = self._build_handler(
             central_device_provider=central_device_provider,
@@ -81,11 +83,12 @@ class MonitorProvider:
         token: str,
         consumer_group: str,
         central_dns_suffix: str,
+        transport: str = None,
     ):
         from azext_iot.monitor.builders import central_target_builder
 
         targets = central_target_builder.build_central_event_hub_targets(
-            cmd, app_id, token, central_dns_suffix
+            cmd, app_id, token, central_dns_suffix, transport=transport
         )
         [target.add_consumer_group(consumer_group) for target in targets]
 
--- a/azext_iot/central/providers/monitor_provider.py
+++ b/azext_iot/central/providers/monitor_provider.py
@@ -47,6 +47,7 @@ class MonitorProvider:
             central_dns_suffix=central_dns_suffix,
             transport=transport,
         )
+        self._transport = transport
         self._handler = self._build_handler(
             central_device_provider=central_device_provider,
             central_template_provider=central_template_provider,
@@ -63,6 +64,7 @@ class MonitorProvider:
             on_start_string=self._handler.generate_startup_string("Monitoring"),
             on_message_received=self._handler.parse_message,
             timeout=telemetry_args.timeout,
+            transport=self._transport,
         )
 
     def start_validate_messages(self, telemetry_args: TelemetryArguments):
@@ -74,6 +76,7 @@ class MonitorProvider:
             on_start_string=self._handler.generate_startup_string("Validating"),
             on_message_received=self._handler.validate_message,
             timeout=telemetry_args.timeout,
+            transport=self._transport,
         )
 
     def _build_targets(
--- a/azext_iot/iothub/providers/discovery.py
+++ b/azext_iot/iothub/providers/discovery.py
@@ -43,54 +43,31 @@ class IotHubDiscovery(BaseDiscovery):
 
     @classmethod
     def get_target_by_cstring(cls, connection_string: str) -> Dict[str, str]:
-        return IotHubTarget.from_connection_string(cstring=connection_string).as_dict()
-
-    def get_target(
-        self, resource_name: str, resource_group_name: str = None, **kwargs
-    ) -> Dict[str, str]:
-        target = super().get_target(
-            resource_name=resource_name,
-            resource_group_name=resource_group_name,
-            **kwargs,
-        )
-
-        if not (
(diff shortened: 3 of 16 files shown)


azext_iot/monitor/utility.py
```diff
@@ -75,3 +77,33 @@ def extract_message_body(message) -> bytes:
...
+    if not proxy_value:
+        return None
+
+    parsed = urlparse(proxy_value if "://" in proxy_value else f"http://{proxy_value}")
+    if not parsed.hostname or not parsed.port:
+        return None
+
+    proxy_scheme = parsed.scheme or "http"
+    settings = {
+        "proxy_hostname": f"{proxy_scheme}://{parsed.hostname}",
```
digimaun, line 110: We should add a comment explaining why the bare hostname doesn't work like [this example](https://github.com/Azure/azure-sdk-for-python/blob/main/sdk/eventhub/azure-eventhub/samples/async_samples/proxy_async.py#L22) , linking to github issue(s) for reference

setup.py
```diff
@@ -58,6 +58,7 @@
     "treelib~=1.6",
     "packaging>=23.2",
     "azure-eventhub~=5.15.0",
+    "aiohttp>=3.9,<4.0",
```
digimaun, line 63: It should be documented here that it is only needed for azure-eventhub amqp over websocket. Lets also see if azure-eventhub supports extras to install aiohttp as part of the azure-eventhub install.

peolivei2 (approved):

azext_iot/iothub/providers/discovery.py
```diff
@@ -45,6 +45,53 @@ def _make_kwargs(self, **kwargs) -> Dict[str, Any]:
     def get_target_by_cstring(cls, connection_string: str) -> Dict[str, str]:
         return IotHubTarget.from_connection_string(cstring=connection_string).as_dict()
 
+    def get_target(
```
digimaun: The `--login` flag exists so users can operate with just a connection string, without an Azure login/session. `self.find_resource()` hits ARM, requiring the user to be authenticated contradicting the purpose of `--login`. I don't see a compelling reason to keep this code at this time.

digimaun: On the proxy support with cstring via --login, thinking of the following options. Either:
- Raise an error (and document in the command) if the user tries to use proxy with --login that it is not supported.
- For monitor-events specifically, attempt to parse the cstring as an IoT Hub one as before...if that fails, attempt to parse as an event hub cstring and use with the eventhub consumer client

What do you think?

digimaun:
What if we added amqp over websocket support as a user option rather than only triggered for proxy? Maybe an add of `--transport` with defaults to `amqp` and there is an `amqp_ws` option.

setup.py
ketkimnaik: It is not part of azure-eventhub extras. I have added comment for it.

azext_iot/monitor/utility.py
ketkimnaik: Done

azext_iot/iothub/providers/discovery.py
ketkimnaik: Agree on removing get_target — the ARM call contradicts --login semantics. For proxy + --login, Option B makes more sense: detect if the --login value is already an Event Hub connection string (contains servicebus.windows.net + EntityPath) and use it directly, bypassing the AMQP redirect. This allows proxy to work end-to-end. The AMQP redirect path (_evaluate_redirect) uses raw pyamqp which doesn't support proxy, so Option A would leave proxy + --login with no viable path.

ketkimnaik:
> What if we added amqp over websocket support as a user option rather than only triggered for proxy? Maybe an add of `--transport` with defaults to `amqp` and there is an `amqp_ws` option.

I have added transport option in monitoring events command.

azext_iot/monitor/utility.py
```diff
@@ -75,3 +77,34 @@ def extract_message_body(message) -> bytes:
...
+        or os.environ.get("HTTP_PROXY")
+        or os.environ.get("http_proxy")
+    )
+
+    if not proxy_value:
+        return None
+
+    parsed = urlparse(proxy_value if "://" in proxy_value else f"http://{proxy_value}")
+    if not parsed.hostname or not parsed.port:
+        return None
```
digimaun, line 105: If there is no port this function returns None and proxy is not used? Is that expected? At least a warning would be helpful

azext_iot/iothub/providers/discovery.py
```diff
@@ -43,6 +43,30 @@ def _make_kwargs(self, **kwargs) -> Dict[str, Any]:
 
     @classmethod
     def get_target_by_cstring(cls, connection_string: str) -> Dict[str, str]:
+        # If this is an Event Hub connection string (e.g. the IoT Hub built-in endpoint
+        # connection string from the Azure portal), parse it directly.  An EH connection
+        # string uses "Endpoint=" and "EntityPath=" rather than "HostName=", so
+        # IotHubTarget.from_connection_string would raise a ValueError on it.
+        if "EntityPath=" in connection_string and "servicebus.windows.net" in connection_string:
```
digimaun, line 51: Does this parsing belong here? Looks like the event hub cstring is being parsed multiple times here and _build_from_eh_connection_string and there is parse_event_hub_connection_string as well.

ketkimnaik: Consolidated EH CS parsing into _build_from_eh_connection_string using the shared parse_event_hub_connection_string utility — no more duplicate parsing.

azext_iot/monitor/utility.py
ketkimnaik: Added logger.warning(...) when proxy env var is set but hostname/port can't be parsed, so the user is informed rather than silently ignored.

azext_iot/monitor/builders/_common.py
```diff
@@ -29,13 +31,19 @@ async def convert_token_to_target(tokens) -> Target:
 
 async def _query_partition_count(hostname, path, credential):
```
digimaun: I think we need to plumb transport through here, otherwise this functionality will use amqp ws only when proxy is set

azext_iot/iothub/providers/discovery.py
digimaun: I still some dupe logic with EventTargetBuilder._build_from_eh_connection_string, what if this part was changed to something like

```
if "EntityPath=" in connection_string and "servicebus.windows.net" in connection_string:
    return {
        "cs": connection_string,
        "entity": "eventhub",  # placeholder, because EventTargetBuilder._build_from_eh_connection_string handles parsing and Target construction
        "policy": "",
        "primarykey": "",
    }
```

digimaun: if "name" is needed can plug that in the dict as well (double check)

azext_iot/monitor/builders/_common.py
ketkimnaik: Done

azext_iot/iothub/providers/discovery.py
ketkimnaik: Done — EH CS path now returns a minimal placeholder (as suggested); all parsing deferred to EventTargetBuilder. "name" included as noted.

azext_iot/operations/hub.py
```diff
@@ -2670,6 +2670,7 @@ def iot_hub_monitor_events(
     content_type=None,
     device_query=None,
     message_count: Optional[int] = None,
+    transport=None,
```
digimaun, line 2673: transport should be added to central.commands_monitor.monitor_events

azext_iot/_params.py
```diff
@@ -432,6 +432,16 @@ def load_arguments(self, _):
             help="Number of telemetry messages to capture before the monitor is terminated. "
             "If not specified, monitor keeps running until meeting the timeout threshold of not receiving messages from hub.",
         )
+        context.argument(
+            "transport",
+            options_list=["--transport", "--tr"],
+            arg_type=get_enum_type(["amqp", "amqp_ws"]),
```
digimaun: Please make a common enum type for this instead of hard coded strings, and we should reference those values

azext_iot/central/providers/monitor_provider.py
digimaun, line 58: transport plumbing here

digimaun, line 70: transport plumbing needed?

ketkimnaik: Yes — validate-messages also accepts --transport and connects to Event Hub partitions the same way, so dropping it here would silently ignore the transport setting.

ketkimnaik: Done

azext_iot/_params.py
ketkimnaik: Done

azext_iot/operations/hub.py
ketkimnaik: Done

azext_iot/monitor/models/enum.py
```diff
@@ -5,10 +5,15 @@
...
 
 
 class Severity(IntEnum):
     info = 1
     warning = 2
     error = 3
+
+
+class Transport(str, Enum):
+    amqp = "amqp"
```
digimaun: for new enums please have the members be all caps e.g. AMQP, AMQP_WS

azext_iot/central/providers/monitor_provider.py
digimaun: Then the command handler should include also transport, not only at the provider level

azext_iot/monitor/models/enum.py
ketkimnaik: I followed the existing values pattern. I will replace it :)

azext_iot/central/providers/monitor_provider.py
ketkimnaik: Done

digimaun (approved):
LGTM

PR #809, opened 2026-04-07
feat: add --hostname-type param for DPS

cheatsheet1999:
﻿Add --hostname-type (auto/classic/device) to DPS enrollment and enrollment-group commands (show/list/create/update) and registration commands (show/list). Transforms IoT Hub hostnames in output to the requested TLS 1.3 endpoint type.

 - Default is `auto`, which resolves to `device` for DPS (device provisioning)
 - `classic` actively transforms to the default hostname
 - `device` transforms to the TLS 1.3 device hostname
 - String transformation via _transform_hostname 
 - Fix IotDpsSku and IotHubSku - They used to act as a string instead of the actual value when creating hub/dps

### - Why adding rule_exclusions on hostname_type
﻿`--hostname-type` is defined at the parent `iot dps enrollment` scope with default=auto , which applies to all sub-commands including update. 
The CLI linter rule says "no defaults on update commands." But removing the default would make update output different from show/list/create for the same enrollment, hence the exclusion

﻿### - Why auto resolves to device for DPS?
 DPS enrollments are for device provisioning. Design doc says "device for device provisioning commands."
 
### - Why string transform instead of reading from API? 
DPS dataplane API doesn't return deviceHostName/serviceHostName on enrollments. Design doc confirms: "no SDK changes required for TLS 1.3

### ﻿- Why no service option for DPS? 
 DPS enrollments are device provisioning. Service hostname is for service-side operations, not applicable to DPS.

### DIFF
--- a/azext_iot/core/custom.py
+++ b/azext_iot/core/custom.py
@@ -99,12 +99,7 @@ def _enum_to_str(value):
     if isinstance(value, Enum):
         return value.value
     s = str(value)
-    return s.split(".")[-1] if "." in s else s
-
-
-def _strip_none(d):
-    """Remove None values from a dict for modelless SDK serialization."""
-    return {k: v for k, v in d.items() if v is not None}
+    return s.rsplit(".", maxsplit=1)[-1] if "." in s else s
 
 
 def iot_dps_create(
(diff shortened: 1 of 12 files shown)


azext_iot/core/custom.py
```diff
@@ -91,6 +91,17 @@ def iot_dps_get(client, dps_name, resource_group_name=None):
     return client.iot_dps_resource.get(provisioning_service_name=dps_name, resource_group_name=resource_group_name)
 
 
+def _enum_to_str(value):
```
cheatsheet1999: Just realized it's better to just use Enum.value, I'll remove this.

cheatsheet1999:
Converting to draft, needs update based on dev design session we discussed today

PR #811, opened 2026-04-10
fix: resolve code scanning security alerts

HangyiWang:
https://github.com/Azure/azure-iot-cli-extension/security/code-scanning

- Added `is_eventhub_connection_string()` helper in `common/utility.py` that parses the `Endpoint=` hostname via `urlparse` and validates it ends with `.servicebus.windows.net`, replacing loose `in` substring checks.
- Updated `iothub/providers/discovery.py` and `monitor/builders/hub_target_builder.py` to use the new helper.

- Added `permissions: {}` at the top level of 6 workflow files to enforce least-privilege token permissions.
- Suppressed `CKV_GHA_7` with inline comments — inputs are typed booleans controlling flow, not injected into build output.

- Fixed `central/commands_api_token.py`: replaced bare `try/except/pass` with safe `get_enum_keys` pattern.
- Fixed `central/commands_device.py`: replaced bare `pass` with `logger.debug()` for skipped devices.
- Suppressed 7 legitimate cleanup/fallback patterns with `# nosec B110`.

- Added missing `__init__.py` files in 11 directories to fix `test_package_init` test.


azext_iot/common/utility.py
```diff
@@ -574,6 +574,25 @@ def find_between(s, start, end):
...
+def is_eventhub_connection_string(cs: str) -> bool:
+    """Check if a connection string is an Event Hub connection string.
+
+    Validates that the string contains 'EntityPath=' and that the
+    Endpoint hostname ends with '.servicebus.windows.net'.
+    """
+    if "EntityPath=" not in cs:
+        return False
+    try:
+        # Extract the Endpoint value: Endpoint=sb://<host>/;...
```
digimaun: _azure.py already has _parse_connection_string() / parse_event_hub_connection_string() which uses split("=", 1) and handles edge cases better, can we not re-use these.

digimaun: also _parse_connection_string does a case insensitive check

.github/workflows/stage_release.yml
```diff
@@ -1,11 +1,17 @@
 name: "[auto] Draft Github Release"
+permissions: {}
 
 on:
   workflow_call:
 
 jobs:
   create_draft_release:
-    runs-on: [self-hosted, 1ES.Pool=iotupx-iot-cli-github-hosted-pool, 1ES.ImageOverride=Ubuntu20.04Compliant]
+    runs-on:
```
digimaun, line 9: More consistent with code base
```
runs-on:
  - self-hosted
  - 1ES.Pool=iotupx-iot-cli-github-hosted-pool
  - 1ES.ImageOverride=Ubuntu20.04Compliant
```

.github/workflows/trigger_ado_int_tests.yml
```diff
@@ -1,4 +1,6 @@
+# checkov:skip=CKV_GHA_7:workflow_dispatch inputs are typed booleans controlling test selection, not injected into build output
 name: Run Azure DevOps Integration Tests
```
digimaun, line 2: Unlike the other workflow files in this PR, none of the three jobs here (call-ado-pipeline, wait-for-completion, CancelPipeline) define job-level permissions. They don't check out code so they likely don't need contents: read, but worth confirming none of the steps (e.g., az pipelines run, the curl cancel call) rely on a GITHUB_TOKEN with specific scopes. If they don't, adding an explicit permissions: {} at the job level would be more intentional and consistent with the pattern in the other workflow files.

.github/workflows/conventional_pr.yml
```diff
@@ -1,12 +1,8 @@
...
-    types: [
-        opened,
-        edited,
-        ready_for_review,
-        reopened,
-    ]
+    types: [opened, edited, ready_for_review, reopened]
 jobs:
   verify_title:
     runs-on: ubuntu-latest
```
digimaun, line 9: nit, similar the comment in trigger_ado_int_tests.yml, adding permissions: {} at the verify_title job level would be more intentional and consistent with the other changes

.github/workflows/release_workflow.yml
```diff
@@ -1,5 +1,7 @@
+# checkov:skip=CKV_GHA_7:workflow_dispatch inputs are typed booleans controlling release flow, not injected into build output
```
digimaun: Does this suppression work if its not at a resource scope level, and placed at the top of the file?

azext_iot/common/utility.py
HangyiWang: ok i see

.github/workflows/stage_release.yml
HangyiWang: updated

.github/workflows/trigger_ado_int_tests.yml
HangyiWang: kk added

.github/workflows/conventional_pr.yml
HangyiWang: added

digimaun (approved):

PR #813, opened 2026-04-14
feat: Support MI auth for DPS linked-hub

cheatsheet1999:
﻿ Description:

  Adds Managed Identity support for `az iot dps linked-hub create` and updates hostname defaults across
  connection string commands for TLS 1.3.

  Changes

  ### DPS linked-hub MI support:

   - New params on linked-hub create: `--authentication-type` (SystemAssigned/UserAssigned),
 ` --user-assigned-identity`, `--hostname-type` (device/classic, default device)
   - MI path sets hostName directly (no connection string), resolves from hub properties
   - Key-based path now defaults to device hostname when using `--hub-name`
   - Validates MI is enabled on DPS before linking
   - Rejects service hostname in --connection-string (DPS doesn't support it)

### Hub connection string hostname:

   - auto hostname type now resolves to device hostname (was service) for hub connection strings
   - All connection string commands support all four types: **auto, classic, device, service** (Service hostname is for hub configuration/management operations, like creating devices, managing twins)

  Testing

   - 1757 unit tests pass (8 new for MI validation and hostname resolution)
   - E2E verified on live resources: key-based (device/classic), MI (SystemAssigned/UserAssigned), service
  rejection, error cases


azext_iot/core/custom.py
```diff
@@ -78,6 +79,13 @@ def _get_resource_group_from_hub(hub):
     return hub["resourcegroup"]
 
 
+def _resolve_linked_hub_hostname(hub, hostname_type="device"):
+    """Resolve IoT Hub hostname for DPS linked hub based on hostname type."""
+    if hostname_type == "device":
+        return hub["properties"].get("deviceHostName") or hub["properties"]["hostName"]
```
HangyiWang, line 87: why `.get("deviceHostName")` and `["hostName"]`? why not use `get` or `[]` both

azext_iot/core/custom.py
```diff
@@ -327,40 +335,107 @@ def iot_dps_linked_hub_create(
     connection_string=None,
     location=None,
     resource_group_name=None,
+    authentication_type=None,
+    user_assigned_identity=None,
+    hostname_type="device",
```
HangyiWang, line 342: previous codes mentioned it will fallback to `classic` so why not use `classic` as default

azext_iot/core/_params.py
```diff
@@ -134,6 +134,26 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='Location of the IoT hub.',
                    arg_group='IoT Hub Identifier',
                    deprecate_info=c.deprecate(hide=True))
+        c.argument('authentication_type',
+                   options_list=['--authentication-type', '--auth-type'],
+                   arg_type=get_enum_type(IotHubAuthenticationType),
+                   help='Authentication type for the linked IoT Hub. '
+                   "'KeyBased' uses a connection string. "
+                   "'SystemAssigned' uses the DPS system-assigned managed identity. "
```
HangyiWang, line 142: nit: i know it would be long but i think `SystemAssignedManagedIdentity` is better `SystemAssigned`, same for `UserAssigned` - a bit weird omiting `ManagedIdentity` suffix since we have another `keyBased` type

azext_iot/core/custom.py
```diff
@@ -327,40 +335,107 @@ def iot_dps_linked_hub_create(
...
+            )
+        if authentication_type == IotHubAuthenticationType.USER_ASSIGNED.value and "UserAssigned" not in identity_type:
+            raise InvalidArgumentValueError(
+                f"User-assigned managed identity is not configured on DPS '{dps_name}'. "
+                "Please assign a user identity before linking with UserAssigned authentication."
+            )
 
-    resource_group_name = _ensure_dps_resource_group_name(client, resource_group_name, dps_name)
+        linked_hub_entry = {
+            "location": location,
```
HangyiWang: nit   `location: location or hub["location"]` can remove line 364 365

azext_iot/core/_params.py
HangyiWang: and how did we get `keyBased` type for connection string type - not sth like `connectionStr`

azext_iot/core/custom.py
cheatsheet1999: Good question. Because `deviceHostName` might not exist in the dict (e.b. if a hub is v1). `hub["properties"]["deviceHostName"]` would raise a KeyError if the key is absent, crashing instead of falling back to hostName. `.get()` returns None when the key is missing, which is falsy, so the or chain works. ﻿

azext_iot/core/custom.py
cheatsheet1999: ﻿The default is "device" per the TLS 1.3 spec: "Newly linked hubs default to the device endpoint." Also confirmed with PM on this topic.

The fallback to classic happens inside the `_resolve_linked_hub_hostname()`.  If the hub doesn't have deviceHostName (not a GWv2 hub), it automatically falls back to the classic hostname. If we defaulted to "classic", TLS 1.3 capable hubs would never use the device endpoint unless the user explicitly specified it, which from my opinion, defeats the purpose of the upgrade.

azext_iot/operations/hub.py
```diff
@@ -2397,6 +2397,10 @@ def _iot_build_sas_token(
 def _transform_hostname(hostname, hostname_type):
     """Transform a hostname to the requested type via string manipulation."""
     parts = hostname.split(".")
+    if len(parts) < 2:
```
HangyiWang: why `< 2`? `>= 2` is fine?

azext_iot/core/_params.py
cheatsheet1999: Yeah. It gives me a lot of headaches when it comes to naming... Let's chat on this :)

azext_iot/core/custom.py
```diff
@@ -327,40 +335,107 @@ def iot_dps_linked_hub_create(
...
+        if not any([connection_string, hub_name]):
+            raise RequiredArgumentMissingError("Please provide the IoT Hub name or connection string.")
+        if not connection_string:
+            hub_client = iot_hub_service_factory(cmd.cli_ctx)
+            hub = iot_hub_get(cmd, hub_client, hub_name=hub_name, resource_group_name=hub_resource_group)
+            host_name = _resolve_linked_hub_hostname(hub, hostname_type)
+            if not location:
+                location = hub["location"]
+            # Build connection string with resolved hostname
+            policies = iot_hub_policy_get(hub_client, hub_name, "iothubowner",
```
HangyiWang: nit make `iothubowner` constant?

azext_iot/core/custom.py
```diff
@@ -327,40 +335,107 @@ def iot_dps_linked_hub_create(
...
+                                         _get_resource_group_from_hub(hub))
+            connection_string = "HostName={};SharedAccessKeyName={};SharedAccessKey={}".format(
```
HangyiWang: nit probably a format function for connection string since it's fixed format for all use cases

azext_iot/core/custom.py
```diff
@@ -327,40 +335,107 @@ def iot_dps_linked_hub_create(
...
+                host_name, policies["keyName"], policies["primaryKey"]
+            )
+        else:
+            if ".service.azure-devices" in connection_string.lower():
```
HangyiWang, line 406: nit `.service.azure-devices` can be constant

azext_iot/operations/hub.py
cheatsheet1999: The `< 2` check is a safety guard to prevent an IndexError on parts[1]. 

Actualy by re-thinking this line I feel it's redandunt, because the input always comes from the SDK (e.g., hub.azure-devices.net) so it's always valid. I'll remove. Thanks.

azext_iot/core/custom.py
cheatsheet1999: Yes!

azext_iot/core/_params.py
cheatsheet1999: ﻿The values `KeyBased`, `SystemAssigned`, and `UserAssigned` come from the DPS ARM API contract, these are the exact values the API accepts for the `authenticationType` field on the linked hub entry. We can't rename them unfortunately.

azext_iot/core/custom.py
cheatsheet1999: ﻿ Good suggestion. Since this check is only used in this one place, I'll keep it inline for now to avoid
 adding a constant for a single-use string. If we add more service hostname validations elsewhere, I'll
 extract it then

azext_iot/core/custom.py
```diff
@@ -327,40 +337,104 @@ def iot_dps_linked_hub_create(
...
-                hub_name = re.search(r"hostname=(.[^\;\.]+)?", connection_string, re.IGNORECASE).group(1)
-            except AttributeError:
-                raise InvalidArgumentValueError("Please provide a valid IoT Hub connection string.")
+            raise RequiredArgumentMissingError(
+                "Please provide --hub-name for managed identity authentication."
+            )
+        if authentication_type == IotHubAuthenticationType.USER_ASSIGNED.value and not user_assigned_identity:
+            raise RequiredArgumentMissingError(
+                "--user-assigned-identity is required when --authentication-type is UserAssigned."
+            )
```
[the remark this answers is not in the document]
cheatsheet1999: I agree this would be a more clear experience

azext_iot/core/custom.py
```diff
@@ -327,40 +337,104 @@ def iot_dps_linked_hub_create(
...
-            location = iot_hub_get(cmd, hub_client, hub_name=hub_name, resource_group_name=hub_resource_group)["location"]
-        except CLIError:
-            raise RequiredArgumentMissingError("Please provide the IoT Hub location.")
+        hub = iot_hub_get(cmd, hub_client, hub_name=hub_name, resource_group_name=hub_resource_group)
+        host_name = _resolve_linked_hub_hostname(hub, hostname_type)
+
+        # Validate MI is enabled on DPS
+        resource_group_name = _ensure_dps_resource_group_name(client, resource_group_name, dps_name)
+        dps = iot_dps_get(client, dps_name, resource_group_name)
+        identity_type = dps["identity"]["type"]
```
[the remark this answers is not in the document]
cheatsheet1999: Applied suggested change

azext_iot/_params.py
```diff
@@ -353,11 +353,12 @@ def load_arguments(self, _):
...
             options_list=["--hostname-type", "--ht"],
-            arg_type=get_enum_type(["auto", "classic", "service"]),
+            arg_type=get_enum_type(HostnameType),
             default=HostnameType.AUTO.value,
             help="Type of hostname to use in the connection string. "
-            "'auto' uses the TLS 1.3 service hostname on GWv2 hubs, classic otherwise. "
+            "'auto' uses the TLS 1.3 device hostname on GWv2 hubs, classic otherwise. "
             "'classic' always uses the default hostname. "
+            "'device' uses the TLS 1.3 device hostname (errors if not GWv2). "
             "'service' uses the TLS 1.3 service hostname (errors if not GWv2).",
```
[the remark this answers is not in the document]
cheatsheet1999: Add GWv2 validation

ketkimnaik (approved):
LGTM

HangyiWang (approved):

PR #816, opened 2026-04-20
ci: correct region in workflow title when auto scheduled 

HangyiWang:
image


cheatsheet1999:
Just to confirm, the issue here was that when triggered by schedule, there are no user inputs, so the region would always fall back to westus regardless of the actual target region?

cheatsheet1999 (approved):

HangyiWang:
> Just to confirm, the issue here was that when triggered by schedule, there are no user inputs, so the region would always fall back to westus regardless of the actual target region?

You're correct.

PR #818, opened 2026-04-22
ci: fix Python 3.10 cryptography import error and split scheduler workflow

HangyiWang:
Integration tests on the `preview` branch fail on Python 3.10 with:

```
ImportError: undefined symbol: PyType_GetName
```
https://github.com/Azure/azure-iot-cli-extension/actions/runs/24769446572/job/72473288383

`az extension add` installs dependencies (including `cryptography`) using az CLI's bundled Python 3.13. The resulting native `.so` uses `PyType_GetName` (added in Python 3.11), which crashes when loaded under Python 3.10. The `preview` branch triggers this because `_factory.py` eagerly imports `azure.identity` at module load time, which chains into `cryptography`.

Add `cryptography` to the `--force-reinstall` list alongside `rpds-py` in the "Build and install extension" step, so binary-incompatible native modules are rebuilt for the actual test Python version.

Split the cron scheduler into a separate workflow (`int_test_schedule.yml`) so that dispatcher-only runs no longer appear as ghost entries under "Integration Tests". The main workflow now only contains actual test runs.

### DIFF
--- a/.github/workflows/int_test.yml
+++ b/.github/workflows/int_test.yml
@@ -1,16 +1,8 @@
 name: Integration Tests
 run-name: >-
-  ${{ github.event_name == 'schedule'
-    && 'Integration Tests (scheduled)'
-    || format('Integration Tests ({0} | {1})', inputs['test-services'] || 'auto', inputs.regions || 'westus') }}
+  Integration Tests (${{ inputs['test-services'] || 'auto' }} | ${{ inputs.regions || 'westus' }})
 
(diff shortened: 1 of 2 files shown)


cheatsheet1999:
LGTM. The cryptography force-reinstall makes sense, same root cause as rpds-py. Splitting the scheduler is a nice cleanup for the Actions UI

cheatsheet1999 (approved):

PR #821, opened 2026-04-28
fix: tweak ADR BYOR int tests

HangyiWang:
- All changes are due to temporary backend behavior — TODO(BYOR) markers added for easy tracking.
- No production code changes; test-only.
- All other BYOR tests (create, activate, sync) continue to pass.


cheatsheet1999 (approved):
No concerns, maybe worth to create an item to track

PR #822, opened 2026-05-04
feat: add linked-hub validation and mixed endpoint warnings

cheatsheet1999:
## No breaking changes to existing user workflows since last bug bash

Add validation and warnings for DPS linked-hub TLS 1.3 endpoint selection:
 - Change --hostname-type default from "device" to "auto" (same behavior: device on GWv2, classic
fallback on V1)
 - Block --hostname-type device when hub does not expose TLS 1.3 CNAME (GWv1)
 - Warn when DPS dynamic allocation references hubs with mixed hostname types (device + classic)

### DIFF
--- a/azext_iot/core/custom.py
+++ b/azext_iot/core/custom.py
@@ -82,11 +82,45 @@ def _get_resource_group_from_hub(hub):
     return hub["resourcegroup"]
 
 
-def _resolve_linked_hub_hostname(hub, hostname_type="device"):
+def _resolve_linked_hub_hostname(hub, hostname_type="auto"):
     """Resolve IoT Hub hostname for DPS linked hub based on hostname type."""
-    if hostname_type == "device":
-        return hub["properties"].get("deviceHostName") or hub["properties"]["hostName"]
-    return hub["properties"]["hostName"]
+    if hostname_type == "classic":
+        return hub["properties"]["hostName"]
+    device_hostname = hub["properties"].get("deviceHostName")
+    if hostname_type == "device" and not device_hostname:
+        hub_name = hub.get("name", "unknown")
+        raise InvalidArgumentValueError(
+            f"The device hostname is not available for IoT Hub '{hub_name}'. "
+            "This hostname type is only supported on GWv2 IoT Hubs. "
+            "Use '--hostname-type classic' or '--hostname-type auto' instead."
+        )
+    # "auto" or "device" with available deviceHostName
+    return device_hostname or hub["properties"]["hostName"]
+
+
+def _warn_mixed_endpoint_types(linked_hubs):
+    """Warn if DPS dynamic allocation references hubs with mixed hostname types."""
+    types = set()
+    for hub in linked_hubs:
+        # Only check hubs participating in allocation
+        if hub.get("applyAllocationPolicy") is False:
+            continue
+        hostname = hub.get("name", "") or hub.get("hostName", "")
+        if not hostname:
+            cs = hub.get("connectionString", "")
+            for part in cs.split(";"):
+                if part.lower().startswith("hostname="):
+                    hostname = part.split("=", 1)[1]
+                    break
+        if ".device." in hostname:
+            types.add("device")
+        elif hostname:
+            types.add("classic")
+    if len(types) > 1:
+        logger.warning(
+            "DPS has linked hubs with mixed hostname types (device and classic). "
+            "This may cause inconsistent behavior during device provisioning."
+        )
 
 
 # CUSTOM METHODS FOR DPS
@@ -340,7 +374,7 @@ def iot_dps_linked_hub_create(
     resource_group_name=None,
     authentication_type=None,
     user_assigned_identity=None,
-    hostname_type="device",
+    hostname_type="auto",
     apply_allocation_policy=None,
     allocation_weight=None,
     no_wait=False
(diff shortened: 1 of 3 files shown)


azext_iot/core/custom.py
```diff
@@ -82,11 +82,45 @@ def _get_resource_group_from_hub(hub):
...
 
-def _resolve_linked_hub_hostname(hub, hostname_type="device"):
+def _resolve_linked_hub_hostname(hub, hostname_type="auto"):
     """Resolve IoT Hub hostname for DPS linked hub based on hostname type."""
-    if hostname_type == "device":
-        return hub["properties"].get("deviceHostName") or hub["properties"]["hostName"]
-    return hub["properties"]["hostName"]
+    if hostname_type == "classic":
+        return hub["properties"]["hostName"]
+    device_hostname = hub["properties"].get("deviceHostName")
```
HangyiWang, line 89: nit: line 88 uses `hub["properties"]["hostName"]` while line 89 uses `hub["properties"].get("deviceHostName")`, why `get` here?

azext_iot/core/custom.py
```diff
@@ -82,11 +82,45 @@ def _get_resource_group_from_hub(hub):
     return hub["resourcegroup"]
 
 
-def _resolve_linked_hub_hostname(hub, hostname_type="device"):
+def _resolve_linked_hub_hostname(hub, hostname_type="auto"):
```
HangyiWang, line 85: should we have a guard like
``` 
if hostname_type not in ("auto", "classic", "device"):
    raise InvalidArgumentValueError(
        "Invalid hostname type. Allowed values are: auto, classic, device."
    )
```

azext_iot/core/custom.py
```diff
@@ -82,11 +82,45 @@ def _get_resource_group_from_hub(hub):
...
+        if hub.get("applyAllocationPolicy") is False:
+            continue
+        hostname = hub.get("name", "") or hub.get("hostName", "")
+        if not hostname:
+            cs = hub.get("connectionString", "")
+            for part in cs.split(";"):
+                if part.lower().startswith("hostname="):
+                    hostname = part.split("=", 1)[1]
+                    break
+        if ".device." in hostname:
```
HangyiWang: This looks a bit fragile to me. do we have a more reliable way to distinguish the hostname type, instead of inferring it from the hostname string?

azext_iot/core/custom.py
cheatsheet1999: Yes, hostName is always present on every hub (classic hostname). deviceHostName only exists on GWv2 hubs - on V1 it's None/missing. So we use .get() for deviceHostName (might not exist) and [] for hostName (always exists, error if missing).

azext_iot/core/custom.py
```diff
@@ -82,11 +82,45 @@ def _get_resource_group_from_hub(hub):
...
+
+
+def _warn_mixed_endpoint_types(linked_hubs):
+    """Warn if DPS dynamic allocation references hubs with mixed hostname types."""
+    types = set()
+    for hub in linked_hubs:
+        # Only check hubs participating in allocation
+        if hub.get("applyAllocationPolicy") is False:
+            continue
+        hostname = hub.get("name", "") or hub.get("hostName", "")
```
HangyiWang: if you get hostname from `hub.get("name", "")`, it could potentially have nothing to do with the type. you can only infer type from `hub.get("hostName", "")`

azext_iot/core/custom.py
cheatsheet1999: I believe the CLI arg_type=get_enum_type(["auto", "device", "classic"]) in _params.py (line 150) already validates the input before it reaches this function. The user can't pass an invalid value. Adding the guard is redundant but harmless

azext_iot/core/custom.py
HangyiWang: Sounds good.

azext_iot/core/custom.py
cheatsheet1999: Yeah, I had the same concern, so I understand your point. However, there’s no explicit type field for linked hubs in the DPS API. That said, the hostname string is the only available indicator. 

I’m leaning on the assumption that the hostname format (hub.device.azure-devices.net) is part of the Azure IoT Hub contract, so I believe it won’t be too fragile.

I can make the check more precise by validating the second segment (hostname.split('.')[1] == 'device'), if you prefer.

ketkimnaik: It would be great if you make it more precise.

azext_iot/core/custom.py
cheatsheet1999: From the API response I saw image

name is always the hostname, it's how DPS identifies the linked hub, but your point is valid, I'll update the priority order to 1) hostname (MI path), then extract from conenctionString (Key-based path), name as last fallback

HangyiWang: Cool, that's right - I'm thinking if `name` has no `device.azure-devices.net` suffix.

azext_iot/core/custom.py
HangyiWang: Okay if this is already validated elsewhere and there’s no additional benefit here, I’d prefer to avoid adding duplicate checks.

azext_iot/core/custom.py
HangyiWang: Yeah, unfortunate that the API doesn’t expose an explicit field. Just want to make sure this string-based check is safe for edge cases.

HangyiWang (approved):
LGTM make sure we covered edge cases.

azext_iot/core/custom.py
cheatsheet1999: Updated, thanks!

PR #839, opened 2026-06-02
feat: add integration tests dashboard with ADO Wiki pipeline

ketkimnaik:
## Summary
Adds an Azure IoT CLI Dashboard that displays integration test workflow run status.

### What's included
- `.azure-pipelines/dashboard.yml` — ADO pipeline (scheduled every 6h + manual trigger)
- `docs/dashboard/generate_wiki.py` — Generates markdown tables for ADO Wiki
- `docs/dashboard/update_wiki.py` — Pushes content to ADO Wiki via REST API
- `.github/workflows/dashboard.yml` — GitHub Actions workflow summary (fallback)
- `docs/dashboard/generate.py` — Generates HTML tables for workflow summary

### Dashboard shows
- **azure-iot-cli-extension** dev + preview branches (side by side)
- **azure-iot-ops-cli-extension** dev branch

### Auth
Uses `iot-dashboard-bot` GitHub App (no PAT required). Private key stored in ADO pipeline secrets.

### Access
ADO Wiki page — gated to project members, stable URL.


digimaun:
I want to step back from the implementation details and raise a concern about the overall approach.

This is internal maintainer tooling, the audience is our team and the output is an internal ADO Wiki, but it's being added to a public repo. That means we'd be publishing internal details in open source, including ADO org and project, and the internal wiki path.

The App credential only needs to be consumed by the ADO pipeline. The private key can sit in Key Vault, and the ADO job can use the installation token and call the GitHub API on its own, which generate_wiki.py already does. So there's no need for the GitHub Actions workflow or for any of this to live in the public repo.

The design also leans on cross repo triggering, where the ops repo dispatches into this repo to refresh the dashboard. That couples the two repos' CI lifecycles and needs a cross repo token to do it. An ADO job that pulls the last runs from both repos on a schedule removes that coupling entirely, and generate_wiki.py already does the pulling.

generate.py and generate_wiki.py also carry near identical fetch and render logic that has already diverged, so a change in one can be missed in the other.

We should consider moving the whole thing into ADO as a single pipeline and generator, keep the App credential, and pull GitHub stats from there. That keeps our internal CI topology out of public view, drops the duplicated generators, removes the cross repo trigger, and avoids maintaining a second pipeline.

ketkimnaik:
Closing: dashboard tooling moved entirely to ADO per review feedback. Internal maintainer tooling should not live in a public repo — avoids exposing ADO org/project/wiki paths and removes cross-repo trigger coupling.

PR #840, opened 2026-06-04
feat: adr ns

HangyiWang:
## Scope and stacked dependency

This is the cloud ADR feature PR against `preview`. Azure/azure-iot-cli-extension#891 builds on this branch and separately upgrades the general Hub/DPS management and data-plane clients, including certificate-issuing device registration. Review those child-PR changes separately from the ADR functionality inherited here.

## Main changes

- Expand the ADR namespace surface to 109 commands: namespace/identity management, Hub/DPS/Software Updates links, certificate authorities and policies, Registry Devices, groups, jobs/runs, reports, and Software Updates.
- Restore the 17 `registry-device` CRUD/wait, authentication, attribute, and capability commands against the modeless SDK. Ordinary authentication output redacts symmetric keys; `auth show-keys` is the explicit secret-retrieval command.
- Make namespace-side link commands canonical. Validate targets, topology, selected identities, and DPS-before-Hub ordering; preserve existing endpoint settings and unrelated identities. Link RBAC preflight reuses existing grants and creates only missing service assignments when the caller is authorized.
- Retire composite link-delete and resource-side namespace-binding options. Standalone resource deletion remains available. Cloud-only ADR does not expose AIO custom-location asset/discovered-resource workflows.
- Add synchronous modeless ADR and Software Updates clients, writable request projections, target-subscription credential binding, and actionable service-error diagnostics.
- Extend unit/native-CLI/SDK-wire coverage and owned-resource integration scenarios, with explicit cleanup evidence and isolated authentication phases.

## API and routing boundaries

| Surface | API version |
|---|---|
| ADR management and Update Instance management | `2026-11-02-preview` |
| Software Updates data plane | `2026-11-02-preview` |
| ADR Hub target and identity-safety reads | `2026-10-01-preview` |
| ADR DPS target and identity-safety reads | `2026-06-01-preview` |

Management uses the Central US EUAP ARM endpoint with AzureCloud-compatible authority/audience validation. Software Updates data uses the service-derived endpoint. General Hub/DPS management SDK defaults in **this PR** remain `2026-05-01-preview` / `2026-08-31`; their upgrades belong to Azure/azure-iot-cli-extension#891.

Resource mutations use bounded resource-state polling; asynchronous POST actions follow their authenticated `Location` URL. These are different polling paths, and service-side polling authorization failures must not be reported as successful resource completion.

## Review follow-up

- Recover failed links without forcing a previously absent inbound identity; direct recovery guidance to supported link-update commands, not retired unlink commands.
- Measure ADR wait budgets with a monotonic deadline and capped sleeps. An in-flight transport request is not made interruptible by the wait loop.
- Distinguish unavailable DPS enrichment from an authoritative empty classic-Hub list, without masking programming errors or mandatory cloud-routing validation.
- Apply real deprecation metadata to legacy endpoint-label aliases while preserving scripts.
- Clarify workflow-first help, managed-identity spellings, new-Hub defaults, and synchronous group-delete `--no-wait` behavior.
- Register local `calculate-hash` as an ordinary command and remove stale linter exemptions and workflow descriptions.

## Validation

- Final `9c9de498` qualification: **8,455 unique unit cases** from the complete earlier cohort plus whole affected-file replays. The feedback refresh passed **126 related cases**; all previous coverage for the changed monitor module was discarded and replaced with fresh data. Every other production/prerequisite file was verified unchanged. This is not a fresh full-suite invocation at the current head.
- Entire-PR introduced/changed handwritten runtime, including multiline executable origins: **4,242/4,242 statements (100.00%)** and **1,301/1,302 branch outcomes (99.92%)**. Generated SDK is measured separately, not used to inflate this result. No coverage exclusions or branch suppressions were added; the remaining structural RBAC loop arc stays in the denominator.
- Flake8 and pylint passed. Independent review findings were resolved, including mandatory cloud validation outside optional enrichment error handling.
- CI-only follow-up `e182258c` selects Python before installing Azure CLI, binding pip to that interpreter so pylint resolves CLI exception ancestry. Production runtime is unchanged; the added pipeline-order regression passed.
- Runtime qualification at `9c9de498` had **43/43 successful PR checks**; the separate live workflows completed with the failures described below. The subsequent documentation-only cleanup at `f2d8c18c` shortens CONTRIBUTING.md, removes two unused RST guides and their docstring reference, and does not change runtime or test logic.
- Full remote workflow [35480849103](https://github.com/Azure/azure-iot-cli-extension/actions/runs/35480849103), on `db84f654` / Python 3.13 / Central US EUAP, completed with failures: ADR **12 passed / 13 failed / 6 fixture-dependent skips**; DPS **62/62 passed**; HubControl **27 passed / 1 failed**; HubData **45 passed / 3 failed**. Required Hub nodes all ran; no filtering or new failure-suppressing skips were introduced.
- ADR failures comprise CA action-poll authorization, stale parent/child deletion state, subscription collection `502 ProviderError`, UpdateInstance operation-lock inconsistency, and one SU-link authorization failure after fresh grants. RegistryDevice functional assertions completed; its case failed during namespace cleanup.
- Hub failures include system-identity Storage authorization and unfiltered job-list rejection. The inherited feedback false-success defect is now fixed: AMQP errors fail unmet requested-message waits, and mixed-device batches retain later matching records. Normal cancellation, completed waits, and non-wait behavior are preserved. Offline regressions and strict review passed; no claim is made that this resolves the underlying transport/feedback loss.
- Complete HubData follow-up on `9c9de498`: [run 35493919545](https://github.com/Azure/azure-iot-cli-extension/actions/runs/35493919545), finished 2026-09-20 09:03 UTC. Entra: **39 passed / 3 failed**; SAS: **6/6 passed**; total **45 passed / 3 failed**, with all required nodes executed and no skips or setup/teardown failures. The feedback scenario passed. Remaining failures are system-identity and user-assigned-identity Storage export authorization (`400 BlobContainerValidationError`) and unfiltered job listing (`400 ArgumentInvalid`). All phase cleanup receipts are complete. Unchanged services were not redundantly rerun.
- HubData cleanup completed. Separate read-only HubControl reconciliation confirmed all previously uncertain non-deployment roots absent; six deployment records remain. Original failed cohort and cleanup receipts are preserved.


zzcodework:

Reviewing `full diff` (head `7b31d0c`), first automated pass on this PR.

This is a large change (128 files, ~98k additions), so I scoped the review to the hand-written surface under `azext_iot/adr/**` plus `HISTORY.rst` / `linter_exclusions.yml` / `tox.ini`, and skipped the ~90k lines of generated `azext_iot/sdk/deviceupdate/**` code. Overall the structure is good: the provider/command split is consistent, the LRO workaround in `providers/base.py` is unusually well documented (including how to revert it), and the `HISTORY.rst` "Intentionally unsupported" section is genuinely useful. Findings below.

**Blocking**

- `azext_iot/adr/providers/link.py`, `_endpoint_update_body` — this drops `provisioning` from the endpoint body on update. The docstring right above it states the backend rejects a sparse delta and requires the endpoint's full identity to be re-sent, i.e. the endpoint map value is replace-semantics. Under that contract, `az iot adr ns link hub update` will silently reset the `availability` / `allocationWeight` that `_build_hub_endpoint_body` wrote at `hub add` time, because the update body only carries `endpointType`, `resourceId` and `inboundCallerIdentity`. The comment says provisioning is "intentionally omitted because established links only support inbound-identity rotation" — but omitting a field from a replace body is not the same as not changing it. Unless the service merges at the endpoint level (worth confirming), please carry the existing block over:

  ```python
  current_provisioning = existing.get("provisioning")
  if current_provisioning is not None:
      body["provisioning"] = current_provisioning
  ```

  This affects `hub_update` only (DPS/SU endpoints have no provisioning block), and I did not find a unit test that asserts `availability`/`allocationWeight` survive a `hub update` — worth adding one either way, since a test would pin down which semantics the service actually has.

**Suggestions**

- `azext_iot/adr/providers/certificate_policy.py` — all five call sites use the pattern `try: poller = ... except HttpResponseError as e: self._handle_parent_not_found(e, ...)`, which is only correct because `_raise_if_parent_not_found` in `providers/base.py` always raises (either `ResourceNotFoundError` or a re-`raise error`). Nothing in the helper's signature or return type communicates that. If it is ever changed to return on the non-matching path, `create`/`update`/`delete` become `UnboundLocalError` on `poller`, and `show`/`list` silently return `None` — an error turning into an empty result is the worst failure mode here. Suggest either annotating it `-> NoReturn` or making the call sites `raise self._parent_not_found_error(...)` so the control flow is explicit to readers and linters.

- `azext_iot/adr/providers/base.py`, `_poll_provisioning_state` — the loop treats `state is None` as success and returns the body on the first GET. For PUT/PATCH on a resource type that does not project `provisioningState` that is the intent, but for DELETE the first GET normally returns the resource that is still being torn down; if that body has no `provisioningState` the delete is reported complete while the resource still exists. Consider gating that branch on `not is_delete`, so a DELETE only completes on a 404 (or an explicit terminal state).

- `azext_iot/adr/providers/base.py`, `_poll_provisioning_state` / `_poll_location` — 5xx responses are swallowed and retried silently for `LRO_POLL_RETRIES`. That is the right behavior, but the eventual timeout message only reports the last `provisioningState`, which is `None` in exactly the case where every poll 5xx'd. Including the last HTTP status code would make a "service was down the whole time" timeout distinguishable from a genuinely stuck provisioning.

- `azext_iot/adr/providers/link.py`, `_side_get_dps_resource` — on any failure (commonly missing RBAC on the DPS, which the docstring calls out) this returns `{}`, so `dps_show` emits `"brownfieldHubs": []`. A caller cannot distinguish "no hubs registered on this DPS" from "we could not read the DPS". Since `brownfieldHubs` is documented in `_help.py` and asserted by tests, consider omitting the key (or emitting `null`) when the side-GET failed, so `[]` stays authoritative.

- `azext_iot/adr/providers/namespace.py`, `identity_remove` — `userAssignedIdentities` is emitted whenever `existing_ids` is non-empty, including the `None` tombstones, even when `_managed_identity_type` resolves to `"None"` (last UAMI removed and no SAMI). ARM generally expects `userAssignedIdentities` to be absent when `type` is `None`. Worth confirming against a real PATCH rather than only the mocked unit test.

**Nits**

- `azext_iot/adr/common.py`, `validate_iso8601_datetime` — `raise InvalidArgumentValueError(...)` inside `except Exception` without `from` discards the parse error; and the bare `Exception` also swallows a `TypeError` if a non-string ever reaches it, reporting it as a bad ISO 8601 value.

- `azext_iot/adr/providers/certificate_authority.py`, `create` — the trailing `else: raise ArgumentUsageError("--type must be either Root or ICA.")` and the `--issuer-type must be either Microsoft or External` branch are unreachable: `params.py` registers both with `get_enum_type(...)`, so knack rejects bad values first. Harmless as defense-in-depth, but it is a second source of truth that will drift if the enums gain a member.

- `azext_iot/adr/providers/certificate_policy.py`, `create` — `no_wait` is popped from `kwargs` and then passed straight back as `no_wait=no_wait`; the pop is redundant here (unlike `update`/`delete`, which branch on it).

Summary: one likely-real behavior bug in `hub update` (silently clearing hub `provisioning` settings), plus a few robustness items around the temporary LRO polling workaround and the "helper always raises" pattern. The command surface removals (`ns credential`, `ns policy`, `ns device`) are called out in `HISTORY.rst`, which is the right place for them.

_This is an automated review and may be incomplete; the generated SDK under `azext_iot/sdk/deviceupdate/**` and the test files were not reviewed line by line._

HangyiWang:
Restored the pre-rollback ADR namespace behavior and hardened it in commits `4654f2c` and `3f6a300`.

Highlights:
- restored automatic, atomic Hub/DPS/SU RBAC preflight (including ADU first-party assignment), target/identity validation, scoped waits, safe link persistence, linked-identity protection, and all five SU catalog/operation-status commands
- added cross-subscription/cross-tenant RBAC handling and protected outbound-identity changes on linked namespaces
- removed unsafe `--no-wait` from destructive two-phase link deletes and hardened cleanup ordering/failure reporting
- separated ordinary Hub/DPS management to the cloud ARM endpoint while retaining ADR target lookups on the canary endpoint; pinned the live-compatible Hub API
- expanded integration coverage, including self-provisioned disposable SU and the five restored SU discovery/status commands

Validation:
- 3,352 repository unit tests passed locally
- 842 ADR/factory tests passed; changed ADR/factory production files are 100% covered except namespace provider at 99.64%
- flake8 clean; pylint 10.00/10
- live namespace CRUD + identity passed in centraluseuap
- live bundled and SU RBAC preflights reached the expected unauthorized boundary on the local non-Owner account and verified zero endpoint mutations + cleanup
- CI lint/unit passed on Python 3.13: https://github.com/Azure/azure-iot-cli-extension/actions/runs/34360075229
- CI ADR live execution was blocked before tests by OIDC federation (`AADSTS700213` for this branch ref); the integration code itself did not run in CI.

HangyiWang:
Follow-up validation:
- command-table linter now passes on `2eb274d`
- CI Python 3.13 lint/unit passed in run 34360075229
- ADR live CI did not execute because Azure Login rejected the branch OIDC subject (`AADSTS700213`); local centraluseuap namespace CRUD/identity tests passed, and bundled + SU RBAC preflights reached the expected unauthorized boundary with zero endpoint mutations
- one Azure DevOps `build_and_publish_azure_cli_test_sdk` check is still red with only a generic `Bash exited with code 1` annotation; the same check passed on the prior push and the build/style/unit matrices are otherwise passing.

HangyiWang:
Resolved the remaining Azure DevOps SDK-build failure in `d2c12a8`. Root cause: installing `azure-cli-testsdk` generated a local `build/` directory that shadowed the installed `build` frontend, causing `python -m build` to fail with `No module named build.__main__`. The pipeline now removes that generated directory before invoking the build frontend. Build 13590 confirms `build_and_publish_azure_cli_test_sdk` passes.

zzcodework:

Reviewing `changes since 7b31d0c` (head `3d8bd39`)

**Scope note:** the increment since my last pass is itself 226 commits / 300 files, so this is not an exhaustive re-review. What I actually read this round: the new wait subsystem (`azext_iot/adr/providers/wait.py`, `azext_iot/adr/commands_wait.py`), the new link helpers (`link_preflight.py`, `link_persistence.py`, `endpoints.py`, `device_class.py`), the packaging/changelog convention files, and the DPS registration provider. The ADR provider rewrites (`link.py` +522/-382, `namespace.py`, `software_update*.py`, `rbac.py`) and the regenerated SDKs were not reviewed line by line. Copilot has an open inline comment on `azext_iot/dps/providers/device_registration.py` — see the note at the end rather than a restatement.

**Blocking**

None found in the areas I reviewed.

**Suggestions**

1. `azext_iot/adr/providers/wait.py` — `wait_for_resource`: a *permanent* 404 is polled for the full timeout on the default (non-explicit) path. In the `except (ClientException, HttpResponseError, CLIResourceNotFoundError)` handler, a not-found result only short-circuits when `deleted` is set, and only re-raises under `explicit and updated and not any((created, exists, custom))`. When the caller passes none of `--created/--updated/--deleted/--exists/--custom`, `explicit` is `False`, so neither branch fires and the loop just records `last_observation = "resource not found"` and sleeps. A typo'd name or wrong resource group therefore spins for `DEFAULT_WAIT_TIMEOUT` (3600 s) before surfacing `Wait operation timed out after 3600 seconds. Last observation: resource not found.` Tolerating a 404 is right *after* the resource has been observed once (create races), but the "never seen it at all" case is almost always a user error. Tracking whether any poll ever returned the resource and failing fast with the underlying `ResourceNotFoundError` when it never did would keep the create-race tolerance and cut an hour off the common mistake.

2. `azext_iot/adr/providers/wait.py` — `_explicit_condition` does not receive `deleted`, so `--deleted` aborts on a `Failed` provisioningState instead of waiting for the resource to disappear. `wait_for_resource` sets `explicit = any((created, updated, deleted, exists, custom))`, so `--deleted` takes the explicit path and each poll calls `_explicit_condition(resource, created=False, updated=False, exists=False, custom=None)`. That function raises `AzureResponseError("The operation failed.")` whenever `provisioningState` casefolds to `failed` — but a resource sitting in `Failed` is exactly the state you often want to wait out a delete on, and a delete of a previously-failed resource is a normal cleanup flow. The only terminal signal that matters for `--deleted` is the 404 already handled in the exception branch; short-circuiting `_explicit_condition` (or passing `deleted` through and returning `False` early) would avoid erroring out on a resource that is on its way out.

**Nits**

3. `azext_iot/adr/providers/link_preflight.py` — `validate_target_state`'s Standard-Hub check is a prefix match: `sku_name.upper().startswith("S")`. That is correct for today's IoT Hub SKUs (accepts S1/S2/S3, rejects F1 and B1/B2/B3), but it is a looser contract than the error message ("Use a Standard S-tier Hub such as S1") implies. An explicit membership test against the known S-tier names would express the intent and not silently accept a future SKU that merely starts with "S".

**Checked and clean** (recording so these are not re-raised):

- Version/changelog convention is satisfied: `azext_iot/constants.py` `VERSION` moves `0.33.0b1` → `1.0.0b1` and `HISTORY.rst` carries a matching `1.0.0b1 (Preview)` heading with substantial entries. This is the convention finding that recurs in this repo, and it is handled here.
- The recurring `azure-core` / `azure-mgmt-core` dependency-floor problem does not apply: `setup.py` declares `azure-core>=1.31.0,<1.40.0` and `azure-mgmt-core>=1.5.0,<2.0.0`, which cover `get_arm_endpoints` (mgmt-core 1.5.0) and `settings.current.azure_cloud` (core 1.31.0) used by the regenerated clients.
- `azext_iot/adr/providers/link_preflight.py` — `get_target` resolves the client with `subscription_id=parsed["subscription_id"]` from the target resource ID rather than the ambient subscription, and maps 404 to a `ResourceNotFoundError` that names the group and subscription it actually looked in. Good cross-subscription behaviour.
- `azext_iot/adr/providers/wait.py` — the failure paths correctly call `progress.stop()` on exception and `progress.end()` on the success/timeout paths, and `sleeper(min(interval, remaining))` keeps the last sleep inside the deadline rather than overshooting by up to `interval`.
- `CredScanSuppressions.json` additions are scoped to test fixture files with `$credential$` / obviously synthetic placeholders and carry justifications — no production path suppressed.

**Cross-branch note on the open Copilot comment:** Copilot's inline finding on `azext_iot/dps/providers/device_registration.py` (supplying only `--certificate-file-path` or only `--key-file-path` falls into the X.509 branch with the other value `None`) is real on this branch — the guard is `elif certificate_file or key_file:` with no pair check. The sibling PR #891 branch (`users/hangyiwang/adr-ignite-preview`) already carries the fix: `_validate_explicit_auth_inputs` plus `if not certificate_file or not key_file: raise RequiredArgumentMissingError(CERTIFICATE_FILE_ERROR)` in `_validate_attestation_params`. Worth porting rather than re-deriving, and worth checking which branch lands first so the fix is not lost in the merge.

---
This is an automated review and may be incomplete — treat the "checked and clean" list as the boundary of what was actually verified rather than as sign-off on the whole increment.

PR #843, opened 2026-06-15
test: 99% unit test coverage

HangyiWang:
## Changes

Closes the unit-test gaps called out in the coverage-by-service post, one service at a time. 

This is primarily **new unit tests**, plus a few small **logic/bug fixes** that the tests surfaced. 

No public command behavior changes.

| Service | Before | After |
|---------|--------|-------|
| ADR     | 94%    | **100%** |
| DPS     | 59%    | **99%**  |
| IoT Hub | 66%    | **99%**  |
| ADU     | 36%    | **99%**  |

### 1. Tests (the bulk of this PR)

- New unit-test files across all four services (commands, providers, loaders, registration/command-map, utilities).
- Focused on real logic: connection-string builders, SAS/hostname routing, validators, result transforms, twin patch-building.

### 2. Logic / business-code changes (3 total, all small)

- **ADU** – `azext_iot/deviceupdate/providers/base.py`: fixed user-assigned identity built as a **set** `{id, {}}` instead of a **dict** `{id: {}}` — the original was the wrong type and would raise `TypeError` on the unhashable `{}` (a `# pylint: disable=unhashable-member` was masking it).
- **ADR** – `azext_iot/adr/providers/namespace.py`: removed dead code (`properties = {}` followed by `if properties:`, which could never run).
- **IoT Hub** – `azext_iot/operations/hub.py`: added `# pragma: no cover` to genuinely unit-untestable code (live AMQP/MQTT monitoring, blob export/import, and defensive `except CloudError` / `except (AttributeError, TypeError)` handlers). No behavior change.

### 3. Other fixes

- **IoT Hub** – `azext_iot/iothub/providers/helpers/state_strings.py`: fixed a message template that used two placeholders (`{0}/{1}`) but is called with a single device arg, which could `IndexError` at format time.

## For reviewers (where to focus / what to skip)

- **Skim safely:** the new `test_*_unit.py` files — additive, no production impact.
- **Look closely (≈4 lines total):** the three source changes above — `base.py` identity fix, `namespace.py` dead-code removal, `state_strings.py` template fix. These are the only changes that touch runtime behavior.
- **The `# pragma: no cover` markers** in `hub.py` are deliberate — they exclude live-I/O and defensive branches that only integration tests can exercise; they make the coverage number reflect what unit tests can honestly cover.

## Notes

- The 3 bugs above were found *because* of writing these tests — a good signal the uplift has value beyond the percentage.
- Once these land, the gaps in the team post are addressed enough to consider adding the **PR build gate** to enforce thresholds going forward.

### DIFF
--- a/azext_iot/deviceupdate/providers/base.py
+++ b/azext_iot/deviceupdate/providers/base.py
@@ -155,7 +155,7 @@ class DeviceUpdateAccountManager(DeviceUpdateClientHandler):
             else:
                 return DeviceUpdateMgmtModels.ManagedServiceIdentity(
                     type=DeviceUpdateMgmtModels.ManagedServiceIdentityType.USER_ASSIGNED,
-                    user_assigned_identities={assign_identity[0], {}},  # pylint: disable=unhashable-member
+                    user_assigned_identities={assign_identity[0]: {}},
                 )
         else:
             target_identity_type = DeviceUpdateMgmtModels.ManagedServiceIdentityType.USER_ASSIGNED
--- a/azext_iot/operations/hub.py
+++ b/azext_iot/operations/hub.py
@@ -495,7 +495,7 @@ def _update_device_key(target, device, auth_method, pk, sk, etag=None):
             device=device,
             custom_headers=headers,
         )
-    except CloudError as e:
+    except CloudError as e:  # pragma: no cover
         handle_service_exception(e)
 
 
@@ -1032,7 +1032,7 @@ def iot_device_module_key_regenerate(
                     "If-Match": '"{}"'.format(etag if etag else "*")
                 },
             )
-        except CloudError as e:
+        except CloudError as e:  # pragma: no cover
             handle_service_exception(e)
 
     if renew_key_type in [RenewKeyType.primary.value, RenewKeyType.secondary.value]:
@@ -1296,7 +1296,7 @@ def _iot_device_module_twin_update(target, device_id, module_id, parameters, eta
         )
     except CloudError as e:
         handle_service_exception(e)
-    except (AttributeError, TypeError) as err:
+    except (AttributeError, TypeError) as err:  # pragma: no cover
         raise CLIInternalError(err)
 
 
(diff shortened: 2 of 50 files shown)


azext_iot/deviceupdate/providers/base.py
```diff
@@ -155,7 +155,7 @@ def assemble_account_auth(
             else:
                 return DeviceUpdateMgmtModels.ManagedServiceIdentity(
                     type=DeviceUpdateMgmtModels.ManagedServiceIdentityType.USER_ASSIGNED,
-                    user_assigned_identities={assign_identity[0], {}},  # pylint: disable=unhashable-member
+                    user_assigned_identities={assign_identity[0]: {}},
```
cheatsheet1999, line 158: If I'm understanding this correctly, assigning a single user‑assigned identity to a Device Update account was previously broken at runtime,  so that path raised a TypeError every time it ran? If so, should we call this out in HISTORY.rst and bump the patch version for the bug fix? Just a thought - non blocking,

HangyiWang: Good call - hold your stamp, let me check first. Thanks.

azext_iot/operations/hub.py
```diff
@@ -2642,7 +2642,7 @@ def _create_export_import_job_properties(
         if exists(input_blob_container_uri):
             input_blob_container_uri = read_file_content(input_blob_container_uri)
         job_properties.input_blob_container_uri = input_blob_container_uri
-    else:
+    else:  # pragma: no cover
```
cheatsheet1999: I agree `# pragma: no cover `is right for genuinely untestable code (live streams, throttle, etc.). But this branch seems easy to cover with one tiny test, like call with job_type="fake", assert it raises ClientRequestError, so we could test instead of exclude. 

I think this PR is already in good shape, we can leave it as‑is and tweak this in a follow‑up sometimes later.

cheatsheet1999 (dismissed):

azext_iot/deviceupdate/providers/base.py
HangyiWang: You're right, I updated the HISTORY.rst but no need to bump version.

azext_iot/operations/hub.py
HangyiWang: Yeah sure, sounds good - I just removed this line and added a quick test. Thanks.

cheatsheet1999 (approved):

PR #854, opened 2026-07-10
fix: preserve scheme-prefixed endpoint names in iot hub state export

cheatsheet1999:
## Description
`az iot hub state export` was dropping routing endpoints when the backing resource (Event Hub / Service Bus namespace, or Cosmos DB / Storage account) had a name that began with a URI-scheme character.

## Why it happened
The host name was pulled out of the endpoint URI with `str.strip("sb://")` / `str.strip("https://")`. `str.strip(chars)` strips any leading/trailing characters contained in `chars` (a *set*), it does **not** strip a prefix. Names starting with `s`, `b`, `h`, `t`, or `p` therefore lost those characters - e.g. `sb://scyhbus...` produced `cyhbus` instead of `scyhbus`. The follow-up lookup for the corrupted name failed and the endpoint was removed from the exported ARM template. This affected all name-/key-based endpoint types: Event Hub, Service Bus queue, Service Bus topic (`sb://`), and Cosmos DB / Storage (`https://`).

## Fix
Replaced the string stripping with `urllib.parse.urlparse`:

urlparse is Python's standard-library URL parser and returns the host correctly regardless of scheme. This is not a new dependency, urlparse is already used the same way in other files. 

The eight previously duplicated parsing sites now go through one  `_endpoint_host_name()` helper, which is what let the same defect spread across every endpoint type.
 

## Tests

- Unit ( test_hub_state_unit.py ) -  TestEndpointHostNameParsing , parametrized over both schemes, scheme-character prefixes, port stripping, and mixed case. 
-  Integration ( test_hub_state_int.py ) - two tests that provision a Service Bus namespace / Cosmos DB (covers both https and sb prefixes), add a routing endpoint, run export, and assert the endpoint survives with its name intact. Both verified against live Azure.

image
image

### Service bus topic
image

Before:
image
Notice the first s being dropped -> error

After:
image

### Cosmos DB
Before:
image

After:
image

### DIFF
--- a/azext_iot/iothub/providers/state.py
+++ b/azext_iot/iothub/providers/state.py
@@ -50,6 +50,12 @@ logger = get_logger(__name__)
 cli = EmbeddedCLI()
 
 
+def _endpoint_host_name(endpoint_uri: str) -> str:
+    """Return the resource host name from a routing endpoint URI, dropping the scheme."""
+    from urllib.parse import urlparse
+    return (urlparse(endpoint_uri).hostname or "").split(".")[0]
+
+
 class StateProvider(IoTHubProvider):
     def __init__(
         self,
@@ -643,7 +649,7 @@ class StateProvider(IoTHubProvider):
         # Cosmos Db
         cosmos_endpoints = []
         for ep in endpoints.get("cosmosDBSqlContainers", []):
-            account_name = ep["endpointUri"].strip("https://").split(".")[0]
+            account_name = _endpoint_host_name(ep["endpointUri"])
             if ep.get("primaryKey") or ep.get("secondaryKey"):
                 try:
                     cosmos_keys = cli.invoke(
@@ -698,7 +704,7 @@ class StateProvider(IoTHubProvider):
         for ep in endpoints["eventHubs"]:
             if ep.get("connectionString"):
                 endpoint_props = parse_iot_hub_message_endpoint_connection_string(ep["connectionString"])
-                namespace = endpoint_props["Endpoint"].strip("sb://").split(".")[0]
+                namespace = _endpoint_host_name(endpoint_props["Endpoint"])
                 try:
                     ep["connectionString"] = cli.invoke(
                         "eventhubs eventhub authorization-rule keys list --namespace-name {} --resource-group {} "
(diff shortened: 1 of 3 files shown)


azext_iot/iothub/providers/state.py
```diff
@@ -50,6 +50,12 @@
 cli = EmbeddedCLI()
 
 
+def _endpoint_host_name(endpoint_uri: str) -> str:
+    """Return the resource host name from a routing endpoint URI, dropping the scheme."""
+    from urllib.parse import urlparse
+    return (urlparse(endpoint_uri).hostname or "").split(".")[0]
```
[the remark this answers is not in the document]
cheatsheet1999: code is self-explanatory.

HangyiWang (dismissed):

azext_iot/iothub/providers/state.py
khalilCodeX, line 57: nit: the comment says hostname, but description says resource name. i think you want to add something like extract backing recourse name...

khalilCodeX, line 57: Should we validate endpoint_uri here, or is an empty / invalid URI impossible by contract?

khalilCodeX (approved):
i have added a few comments. take a look at them but the code is sound overall.

azext_iot/iothub/providers/state.py
cheatsheet1999: Renamed the helper to _endpoint_resource_name and updated the docstring so the name, description, and behavior all agree. It returns the backing resource name, thanks!

cheatsheet1999: These URIs always come from the hub's control-plane routing config, where a valid endpoint URI is guaranteed, so I kept the helper lean. It also degrades safely if that ever changed,  urlparse("").hostname is None , so it returns  ""  rather than raising. Happy to add an explicit guard if you'd prefer the defensiveness.

cheatsheet1999:
@khalilCodeX, could you take another look before I merge? Thank you,

HangyiWang (approved):

cheatsheet1999:
Synced offline, merging into dev.

PR #856, opened 2026-07-11
feat: remove --hostname-type option from linked-hub update

cheatsheet1999:
Removes  `--hostname-type`  from  `az iot dps linked-hub update`  - endpoint hostname is now fixed at creation  (matching the portal);  `create` keeps  `--hostname-type`  and  `update`  keeps its auth/allocation options. Bumped to  0.31.0b3 .

image

image
image

### DIFF
--- a/azext_iot/core/_params.py
+++ b/azext_iot/core/_params.py
@@ -170,13 +170,6 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
                    help='IoT Hub short name. Preferred over --linked-hub; resolves to the correct '
                    'entry regardless of classic/device hostname.',
                    arg_group='Linked Hub Identifier')
-        c.argument('hostname_type',
-                   options_list=['--hostname-type', '--ht'],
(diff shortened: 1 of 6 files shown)


HangyiWang (approved):

azext_iot/core/_params.py
peolivei2, line 185: The update still accepts a connection string. Does the backend validate that the connection string is for the same hostname that was set at creation?

peolivei2 (approved):

azext_iot/core/_params.py
cheatsheet1999: Good question. Yes, the backend validates this. An update with a hostname that does not match the connection string is rejected with the following error:

ERROR: (400309) hostName does not match the host name value specified in the connection string.

image

In the first command, I tried to link the entry using a classic connection string (stored the connection string in an env variable so it would not be exposed in the terminal here). The request failed with a service-side error, and the entry remained unchanged afterward. It still uses the device hostname.

PR #865, opened 2026-07-20
fix: add required permission for int test for release workflow

HangyiWang:


ketkimnaik (approved):

cheatsheet1999:
I’m curious what changed here. The integration tests were passing without `id-token: write` before, were they previously authenticating to Azure through a different mechanism, or is this permission only needed for a new workflow path?

HangyiWang:
> I’m curious what changed here. The integration tests were passing without `id-token: write` before, were they previously authenticating to Azure through a different mechanism, or is this permission only needed for a new workflow path?

So initially there was no permissions field at all. Later for security reasons, the scope was narrowed to `contents:read`. However, this integration test workflow also requires the additional `id-token:write` permission.

PR #870, opened 2026-07-30
fix: add preview flag to az iot adr ns commands

HangyiWang:


zzcodework:

Reviewing the **full diff** (head `6a0c951e`, base `preview`).

Small, correct change — `is_preview=True` on the four `iot adr ns` command groups is the right mechanism (group-level flag propagates the preview warning to every command in the group), and the reformatting to multi-line `with self.command_group(...)` keeps line length in check.

**Blocking**

_None identified._

**Suggestions**

1. `azext_iot/tests/adr/test_adr_registration_unit.py::test_load_adr_commands` — the assertion `[command_call.args[0] for ...] == expected_preview_groups` pins both the exact set **and the exact order** of every group registered by `load_adr_commands`. Any future group added to `command_map.py` breaks this test for a reason unrelated to preview status, and the failure message will just be a list diff. Asserting the mapping instead is more durable and states the actual intent:

   ```python
   registered = {c.args[0]: c.kwargs.get("is_preview") for c in loader.command_group.call_args_list}
   assert set(registered) == set(expected_preview_groups)
   assert all(registered[g] is True for g in expected_preview_groups)
   ```

2. No `HISTORY.rst` entry. Marking a command surface as preview is customer-visible (it adds the "is in preview and under development" warning to every invocation), so it belongs in the unreleased changelog section.

**Nits**

3. Worth confirming in the PR description whether the parent `iot adr` group (registered outside `azext_iot/adr/command_map.py`) also needs the preview flag, or whether only the `ns` subtree is preview by design — a reader of this diff cannot tell.

**Summary** — change itself is correct and low risk; only asks are loosening the brittle ordering assertion in the new test and adding a changelog entry.

cheatsheet1999 (approved):

cheatsheet1999 (approved):

PR #873, opened 2026-07-31
fix: route data-plane operations to the correct GWv2 endpoint

cheatsheet1999:
## Summary

Routes IoT Hub data-plane operations to the correct GWv2 endpoint. TLS 1.3-capable (GWv2) hubs expose three CNAMEs (classic / device / service); previously the SDK factory always used the classic `entity` hostname, so device-bound traffic did not use the device endpoint.

## Validation
- Two success runs 
  - https://github.com/Azure/azure-iot-cli-extension/actions/runs/30592758675
  -  https://github.com/Azure/azure-iot-cli-extension/actions/runs/30580024507


zzcodework:

Reviewing the **full diff** (head `3537fb7d`, base `preview`).

Nice, focused fix — the split-hostname derivation is centralized, the `target.get("serviceHostName") or target["entity"]` fallback keeps classic hubs on the existing path, and unit coverage was added for both the split and classic cases. A few things worth a second look.

**Blocking**

_None identified._

**Suggestions**

1. `azext_iot/_factory.py` — silent semantic change for the service SDK's SAS uri. Previously both `_get_iothub_device_sdk` and `_get_iothub_service_sdk` used `self.sas_uri`, which `__init__` rewrites to `"{entity}/devices/{device_id}"` whenever `SdkResolver` is constructed with `device_id`. After this change the service SDK always uses the bare hostname, so a resolver built with a `device_id` now issues a **hub-scoped** rather than device-scoped SAS token for service calls. I checked the only non-test call site that passes `device_id` (`azext_iot/iothub/providers/device_messaging.py`, `self.device_resolver = SdkResolver(target=self.target, device_id=device_id)`) and it only pulls `SdkType.device_sdk`, so there is no live regression — but the change is invisible in the diff. Worth an explicit note in the PR description, and consider dropping/renaming `self.sas_uri` + `self.endpoint` now that only `_get_dps_service_sdk` still reads them; leaving them as a half-used attribute is how this drifts next time.

2. The `_service_hostname(target)` helper is copy-pasted verbatim into `azext_iot/monitor/event.py` and `azext_iot/monitor/builders/hub_target_builder.py`, and re-implemented inline in `_factory.py` (`self.target.get("deviceHostName") or self.target["entity"]`) and again in `device_messaging.py::_device_hostname`. That is four copies of the same fallback rule. Since `azext_iot/iothub/common.py` now owns `transform_iot_hub_hostname`, that is the natural home for `resolve_service_hostname(target)` / `resolve_device_hostname(target)` too.

3. `azext_iot/iothub/common.py::transform_iot_hub_hostname` — the function moved out of `operations/hub.py` but lost its use of the `HostnameType` enum, so the map keys are now bare literals (`"classic"`, `"device"`, `"service"`) while callers still pass `HostnameType.X.value`. The values happen to match today (`common/shared.py::HostnameType`), so this is not a bug, but it removes the compile-time link between producer and consumer. Same for the `("device", "service")` literal tuple in `derive_iot_hub_hostnames`. Recommend importing `HostnameType` in `common.py` and keying off it as the original did.

4. `azext_iot/iothub/common.py::transform_iot_hub_hostname` indexes `parts[1]` unguarded, so a dotless hostname raises `IndexError` rather than a CLI error. Pre-existing behaviour, but the function is now a shared public helper in `common.py` rather than a private one in `hub.py`, so a `len(parts) < 2 -> return hostname` guard is cheap insurance. (`derive_iot_hub_hostnames` is correctly guarded with `len(parts) > 2`.)

5. `azext_iot/iothub/models/iothub_target.py::as_dict` gates on `if self.service_hostname:` but sets **both** `serviceHostName` and `deviceHostName` inside the branch. They are always assigned together by `derive_iot_hub_hostnames`, so it works, but keying one field's presence off the other's truthiness is fragile — assert on both, or return the pair as a small dict from the helper.

6. No `HISTORY.rst` entry. This is a customer-visible data-plane routing fix; per repo convention it should land in the unreleased section so the behaviour change on GWv2 hubs is discoverable.

**Nits**

7. The storage-account changes (`--allow-shared-key-access true` in `azext_iot/tests/deviceupdate/conftest.py` and `azext_iot/tests/helpers.py`) are unrelated test-infra fixes bundled into a data-plane routing PR. Fine to keep if you call them out in the description, but they make the PR harder to bisect later.

8. `derive_iot_hub_hostnames` returns `(None, None)` for classic hostnames, so a GWv2 hub whose ARM `properties.hostName` is still reported in classic form gets no split routing. If that can happen, it is worth a comment stating the assumption (routing is driven purely by the shape of the hostname the service hands back).

**Summary** — logic looks correct and the fallback path preserves classic-hub behaviour; the main asks are the undocumented service-SDK SAS-scope change, de-duplicating the four copies of the hostname fallback, and a `HISTORY.rst` entry.

cheatsheet1999:
> Reviewing the **full diff** (head `3537fb7d`, base `preview`).
> 
> Nice, focused fix — the split-hostname derivation is centralized, the `target.get("serviceHostName") or target["entity"]` fallback keeps classic hubs on the existing path, and unit coverage was added for both the split and classic cases. A few things worth a second look.
> 
> **Blocking**
> 
> _None identified._
> 
> **Suggestions**
> 
> 1. `azext_iot/_factory.py` — silent semantic change for the service SDK's SAS uri. Previously both `_get_iothub_device_sdk` and `_get_iothub_service_sdk` used `self.sas_uri`, which `__init__` rewrites to `"{entity}/devices/{device_id}"` whenever `SdkResolver` is constructed with `device_id`. After this change the service SDK always uses the bare hostname, so a resolver built with a `device_id` now issues a **hub-scoped** rather than device-scoped SAS token for service calls. I checked the only non-test call site that passes `device_id` (`azext_iot/iothub/providers/device_messaging.py`, `self.device_resolver = SdkResolver(target=self.target, device_id=device_id)`) and it only pulls `SdkType.device_sdk`, so there is no live regression — but the change is invisible in the diff. Worth an explicit note in the PR description, and consider dropping/renaming `self.sas_uri` + `self.endpoint` now that only `_get_dps_service_sdk` still reads them; leaving them as a half-used attribute is how this drifts next time.
> 2. The `_service_hostname(target)` helper is copy-pasted verbatim into `azext_iot/monitor/event.py` and `azext_iot/monitor/builders/hub_target_builder.py`, and re-implemented inline in `_factory.py` (`self.target.get("deviceHostName") or self.target["entity"]`) and again in `device_messaging.py::_device_hostname`. That is four copies of the same fallback rule. Since `azext_iot/iothub/common.py` now owns `transform_iot_hub_hostname`, that is the natural home for `resolve_service_hostname(target)` / `resolve_device_hostname(target)` too.
> 3. `azext_iot/iothub/common.py::transform_iot_hub_hostname` — the function moved out of `operations/hub.py` but lost its use of the `HostnameType` enum, so the map keys are now bare literals (`"classic"`, `"device"`, `"service"`) while callers still pass `HostnameType.X.value`. The values happen to match today (`common/shared.py::HostnameType`), so this is not a bug, but it removes the compile-time link between producer and consumer. Same for the `("device", "service")` literal tuple in `derive_iot_hub_hostnames`. Recommend importing `HostnameType` in `common.py` and keying off it as the original did.
> 4. `azext_iot/iothub/common.py::transform_iot_hub_hostname` indexes `parts[1]` unguarded, so a dotless hostname raises `IndexError` rather than a CLI error. Pre-existing behaviour, but the function is now a shared public helper in `common.py` rather than a private one in `hub.py`, so a `len(parts) < 2 -> return hostname` guard is cheap insurance. (`derive_iot_hub_hostnames` is correctly guarded with `len(parts) > 2`.)
> 5. `azext_iot/iothub/models/iothub_target.py::as_dict` gates on `if self.service_hostname:` but sets **both** `serviceHostName` and `deviceHostName` inside the branch. They are always assigned together by `derive_iot_hub_hostnames`, so it works, but keying one field's presence off the other's truthiness is fragile — assert on both, or return the pair as a small dict from the helper.
> 6. No `HISTORY.rst` entry. This is a customer-visible data-plane routing fix; per repo convention it should land in the unreleased section so the behaviour change on GWv2 hubs is discoverable.
> 
> **Nits**
> 
> 7. The storage-account changes (`--allow-shared-key-access true` in `azext_iot/tests/deviceupdate/conftest.py` and `azext_iot/tests/helpers.py`) are unrelated test-infra fixes bundled into a data-plane routing PR. Fine to keep if you call them out in the description, but they make the PR harder to bisect later.
> 8. `derive_iot_hub_hostnames` returns `(None, None)` for classic hostnames, so a GWv2 hub whose ARM `properties.hostName` is still reported in classic form gets no split routing. If that can happen, it is worth a comment stating the assumption (routing is driven purely by the shape of the hostname the service hands back).
> 
> **Summary** — logic looks correct and the fallback path preserves classic-hub behaviour; the main asks are the undocumented service-SDK SAS-scope change, de-duplicating the four copies of the hostname fallback, and a `HISTORY.rst` entry.
> 
> _This is an automated review and may be incomplete; please verify findings before acting on them._

It looks like there are no blocking issues, and all the suggested changes are cosmetic. I’ll create a child PBI to track them as follow-up work, since the latest commit is stable and passes both the integration tests and the bug bash.

HangyiWang (approved):

PR #875, opened 2026-08-03
feat: radar ui

HangyiWang:


thirdpartynotice
```diff
@@ -1,188 +0,0 @@
-Third Party Notices for Azure IoT CLI Extension
```
digimaun, line 1: Why delete the 3rd party notice?

HangyiWang: 😆 you're so quick - it's deleted by accident, i will restore

cheatsheet1999:
Could you add a short description to this PR when you get a chance? An 11k-line diff looks a bit intimidating 🙂

HangyiWang:
> Could you add a short description to this PR when you get a chance? An 11k-line diff looks a bit intimidating 🙂

🤣 You guys are really active with PR reviews. I'll add more details and explanations here, but this is just a draft PoC and is still far from ready. The main idea is to explore an interesting terminal-based UI experience for ADR linking.

PR #876, opened 2026-08-05
fix(iot edge devices create): validate device Id used as file system path

zzcodework:
## Summary

Fixes a path traversal issue in `az iot edge devices create`, found during a security review of the repo.

Device Ids from `--cfg` config files and `--device` arguments were used verbatim as file system path segments when building device bundles:

```python
device_cert_output_directory = bundle_output_directory.joinpath(device_id)
if exists(device_cert_output_directory):
    rmtree(device_cert_output_directory)   # attacker-chosen path
```

`PurePath.joinpath` honors both relative traversal and absolute-path override, so a config entry such as `deviceId: "C:/Users/<user>/.ssh"` (or `../../..`) triggered `rmtree()` on that path. This happened *before* the IoT Hub `create_or_update_identity` call that would have rejected the illegal device Id, so the service was not a mitigating control. The same unvalidated value also drove certificate/`config.toml`/`install.sh` writes and the `tar_directory()` output path.

The trust boundary here is the config file, not the CLI invocation — nested edge config files are designed as shareable fleet-definition artifacts (see `docs/samples/sample_devices_config.yaml`), so an operator can be handed one authored elsewhere.

## Changes

- **`edge_device_config.py`** — new `validate_edge_device_id()` enforcing the IoT Hub allowed character set, the 128-character limit, and rejecting path separators / `.` / `..`. Wired into both parse paths (`_process_edge_config_device` and `process_edge_devices_config_args`).
- **`device_identity.py`** — new `_resolve_device_bundle_directory()` asserts the joined path resolves within the `--out` directory (defense in depth) before any `rmtree`/`makedirs`/write. This also prevents a symlinked device directory from redirecting `rmtree` outside the bundle root.
- Tests + a `traversal_device_id.yml` invalid-config fixture.

## Scope

Deliberately limited to the path traversal. The review that found this also flagged an unquoted `device_id` interpolation into the generated `install.sh` (which the bundled README instructs the operator to run under `sudo`). That is **shell injection, not path traversal**, and it is *not* fixed here — note that `$`, `(` and `)` are all legal IoT Hub device Id characters, so the character validation added in this PR does not close it. Being tracked separately.

Other lower-severity hardening items from the same review (private key file permissions, SAS token scope, CI workflow interpolation quoting) are likewise out of scope.

## Validation

`azext_iot/tests/iothub/devices/test_iot_edge_devices_unit.py` — 71 passed. `flake8` clean on the changed files.

### DIFF
--- a/azext_iot/iothub/providers/device_identity.py
+++ b/azext_iot/iothub/providers/device_identity.py
@@ -6,7 +6,7 @@
 from pathlib import PurePath
 from knack.prompting import prompt_y_n
 from os import makedirs
-from os.path import exists, abspath
+from os.path import exists, abspath, commonpath, realpath
 from shutil import rmtree
 from azext_iot.common.certops import (
     create_self_signed_certificate,
@@ -52,6 +52,33 @@ from azext_iot.operations.hub import _assemble_device
 from azext_iot.sdk.iothub.service.models import Device
 
 logger = get_logger(__name__)
+
+
+def _resolve_device_bundle_directory(
+    bundle_output_directory: PurePath, device_id: str
+) -> PurePath:
+    """
+    Join a device Id onto the bundle output directory, ensuring the result stays within that
+    directory. Device Ids originate from user supplied config files and are validated when the
+    config is parsed - this is a final guard before destructive file system operations run.
+    """
+    device_directory = bundle_output_directory.joinpath(device_id)
+    bundle_root = realpath(str(bundle_output_directory))
+    resolved_device_directory = realpath(str(device_directory))
+    try:
+        contained = (
+            resolved_device_directory != bundle_root
+            and commonpath([bundle_root, resolved_device_directory]) == bundle_root
+        )
+    except ValueError:
+        # paths reside on different drives (Windows)
+        contained = False
+    if not contained:
+        raise InvalidArgumentValueError(
+            f"Device Id '{device_id}' is invalid, it resolves to a path outside of the "
+            f"output directory '{bundle_output_directory}'."
+        )
+    return device_directory
 
 
 class DeviceIdentityProvider(IoTHubProvider):
@@ -247,7 +274,9 @@ class DeviceIdentityProvider(IoTHubProvider):
             device_id = device.device_id
             device_cert_output_directory = None
             if bundle_output_directory:
-                device_cert_output_directory = bundle_output_directory.joinpath(device_id)
+                device_cert_output_directory = _resolve_device_bundle_directory(
+                    bundle_output_directory, device_id
+                )
                 # if the device's folder already exists, remove it
                 if exists(device_cert_output_directory):
                     rmtree(device_cert_output_directory)
--- a/azext_iot/iothub/providers/helpers/edge_device_config.py
+++ b/azext_iot/iothub/providers/helpers/edge_device_config.py
@@ -6,7 +6,6 @@
 """This module defines common values and functions for processing edge device configurations"""
 
 import re
-import shlex
 from pathlib import PurePath
 from os import getcwd
 from typing import Optional, List, Dict, Any
@@ -141,7 +140,7 @@ EDGE_CONFIG_SCRIPT_HEADERS = """
 # This script will attempt to configure a pre-installed iotedge as a nested node.
 # It must be run as sudo, and will modify the ca
 
-device_id={}
+device_id="{}"
 cp config.toml /etc/aziot/config.toml
 """
 EDGE_CONFIG_SCRIPT_HOSTNAME = """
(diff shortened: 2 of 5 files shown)


azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -467,7 +502,7 @@ def create_edge_device_config_script(
     parent_hostname: Optional[str] = None,
 ):
     return "\n".join(
-        [EDGE_CONFIG_SCRIPT_HEADERS.format(device_id)]
+        [EDGE_CONFIG_SCRIPT_HEADERS.format(shlex.quote(device_id))]
```
digimaun, line 505: Quoting the assignment does not cover option injection. `-` is a legal device Id character and `shlex.quote("-dev1")` returns it unquoted, so `EDGE_CONFIG_SCRIPT_CA_CERTS` ends up running `cp "-dev1.full-chain.cert.pem" ...` where the quotes stop word splitting but not option parsing, and cp reads the argument as flags. Could we prefix the source paths with `./` in the template, or pass `--` before the operands? The same shape applies to the other `cp` lines that build a name from `$device_id`.

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -239,6 +246,33 @@
...
+        raise InvalidArgumentValueError(
+            "A device parameter is missing required attribute 'device_id'"
+        )
+    if len(device_id) > MAX_DEVICE_ID_LEN:
+        raise InvalidArgumentValueError(
+            f"Device Id '{device_id}' is invalid, it must be at most {MAX_DEVICE_ID_LEN} characters long."
+        )
+    if not DEVICE_ID_ALLOWED_PATTERN.match(device_id):
+        raise InvalidArgumentValueError(
+            f"Device Id '{device_id}' contains invalid characters. Device Ids may only contain "
```
digimaun, line 265: This branch and the length one above it interpolate the raw device Id, and both fire for input that just failed validation, so an Id carrying ANSI escapes gets echoed to the operator terminal from the one function whose job is to not trust this value. Could we render it with `ascii(device_id)`, or truncate it, so the message stays inert?

azext_iot/iothub/providers/helpers/edge_device_config.py
```diff
@@ -239,6 +246,33 @@
...
+    """
+    if not isinstance(device_id, str) or not device_id:
+        raise InvalidArgumentValueError(
+            "A device parameter is missing required attribute 'device_id'"
+        )
+    if len(device_id) > MAX_DEVICE_ID_LEN:
+        raise InvalidArgumentValueError(
+            f"Device Id '{device_id}' is invalid, it must be at most {MAX_DEVICE_ID_LEN} characters long."
+        )
+    if not DEVICE_ID_ALLOWED_PATTERN.match(device_id):
```
digimaun, line 263: `$` matches just before a trailing newline in Python, so `re.match` accepts `"device\n"` even though `\n` is not in the IoT Hub set. This is not a traversal path since `/` still cannot appear, but it reaches `makedirs` and fails there with a raw OSError on Windows, or creates a directory with an embedded newline on Linux. Could we use `fullmatch` here? A `"device\n"` case in the invalid parametrize would lock it in.

azext_iot/tests/iothub/devices/test_iot_edge_devices_unit.py
```diff
@@ -1241,6 +1249,60 @@ def test_create_edge_device_config_script(
...
+        ],
+    )
+    def test_invalid_device_ids(self, device_id):
+        with pytest.raises(InvalidArgumentValueError):
+            validate_edge_device_id(device_id)
+
+    @pytest.mark.parametrize(
+        "device_id",
+        ["..", "../escape", "/absolute", "C:/Windows"],
+    )
```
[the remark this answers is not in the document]
digimaun: Confirming this one is worth prioritizing. It is not just OS dependent, it is currently failing. On POSIX `C:/Windows` joins to `bundle/C:/Windows`, which stays inside the root, so nothing raises and the test reports DID NOT RAISE. That is the cause of the red unit test legs on ubuntu and macos across all four Python versions, while the Windows legs stay green. Dropping the case is fine since `/absolute` already covers the override behavior on every platform.

azext_iot/iothub/providers/device_identity.py
```diff
@@ -54,6 +54,33 @@
...
+    bundle_output_directory: PurePath, device_id: str
+) -> PurePath:
+    """
+    Join a device Id onto the bundle output directory, ensuring the result stays within that
+    directory. Device Ids originate from user supplied config files and are validated when the
+    config is parsed - this is a final guard before destructive file system operations run.
+    """
+    device_directory = bundle_output_directory.joinpath(device_id)
+    bundle_root = realpath(str(bundle_output_directory))
+    resolved_device_directory = realpath(str(device_directory))
```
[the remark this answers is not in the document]
digimaun: The underlying point holds, though `"` is not actually permitted here. The allowed set ends in a single quote rather than a double, so `"`, `<`, `>`, `|`, `\` and `/` are already excluded by the pattern. The ones that pass validation and still break on Windows are `:`, `?` and `*`. Worth deciding whether to reject those outright, since the alternative today is a raw OSError out of `makedirs` rather than a message the user can act on.

PR #878, opened 2026-08-14
fix: backfill hostName on dps linked hub update 

cheatsheet1999:
## Problem

Before:
image
After:
image

`az iot dps linked-hub update` fails on existing linked hub whose stored entry has no `hostName`  - links created by extension version earlier than preview version.

- `--auth-type SystemAssigned | UserAssigned` -> `(400309) hostName is required when connectionString is not provided`
- `--auth-type KeyBased` (key refresh / switch back) -> unhandled `KeyError: 'hostName'`

Old links have no `hostName` saved - the service returns the hostname in `name` instead.
Switching a link to managed identity clears its connection string. That leaves the link with no connection string and no `hostName`, and the service requires one of the two.

## Fix
Backfill `hostName` from `name` for each linked hub before submitting the resource

`create` and `delete` are unaffected, since each stored entry already satisfies the service requirements independently: KeyBased entries always include a connection string, and MI entries always include `hostName`. Those paths simply re-PUT the unmodified existing entries.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -3,6 +3,15 @@
 Release History
 ===============
 
+0.32.0b2 (Preview)
+++++++++++++++++++
+
+**DPS bug fixes**
+
+* ``az iot dps linked-hub update`` no longer fails with ``(400309) hostName is required when connectionString is not provided`` when switching a pre-existing linked hub to ``SystemAssigned`` or ``UserAssigned`` authentication.
+
+* ``az iot dps linked-hub update --authentication-type KeyBased`` no longer raises ``KeyError: 'hostName'`` when refreshing the key of, or switching back to, one of those same links.
+
 0.32.0b1 (Preview)
 ++++++++++++++++++
 
--- a/azext_iot/constants.py
+++ b/azext_iot/constants.py
@@ -7,7 +7,7 @@
 
 import os
 
-VERSION = "0.32.0b1"
+VERSION = "0.32.0b2"
 EXTENSION_NAME = "azure-iot"
 EXTENSION_ROOT = os.path.dirname(os.path.abspath(__file__))
 EXTENSION_CONFIG_ROOT_KEY = "iotext"
--- a/azext_iot/core/custom.py
+++ b/azext_iot/core/custom.py
@@ -99,6 +99,13 @@ def _resolve_linked_hub_hostname(hub, hostname_type="auto"):
     return device_hostname or hub["properties"]["hostName"]
 
 
+def _ensure_linked_hub_hostnames(linked_hubs):
+    for entry in linked_hubs or []:
+        if not entry.get("hostName") and entry.get("name"):
+            entry["hostName"] = entry["name"]
+    return linked_hubs
+
+
 def _warn_mixed_endpoint_types(linked_hubs):
     """Warn if DPS dynamic allocation references hubs with mixed hostname types."""
     types = set()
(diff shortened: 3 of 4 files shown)


zzcodework:

Reviewing `full diff` (head `20c9ae2`), base `preview`. 4 files, +81/-1.

The fix looks correct for the reported failure: legacy linked-hub entries carry the hostname only in `name`, `iot_dps_linked_hub_update` blanks `connectionString` on the MI switch, and the KeyBased rebuild path reads `target_entry["hostName"]` directly — so backfilling `hostName` from `name` before the PUT addresses both the `400309` and the `KeyError`. Version bump + `HISTORY.rst` entry are in place. No blocking issues.

**Suggestions**

1. `azext_iot/core/custom.py` — `_ensure_linked_hub_hostnames(linked_hubs)` is called on the whole `dps["properties"]["iotHubs"]` list, before `_find_linked_hub_entry`, so *every* legacy link on the DPS silently gains a `hostName` in this PUT, not just the one being updated. The written value equals `name`, so it should be a no-op semantically, but it is a wider mutation than the PR description implies. Scoping it to `target_entry` (after the lookup) would keep the blast radius equal to the command's stated scope; if the broad backfill is deliberate, a one-line comment saying so would help the next reader.

2. Related side effect of doing it list-wide: `_warn_mixed_endpoint_types` resolves a hostname in the order `hostName` → `HostName=` inside `connectionString` → `name`. After the backfill it will always take the first branch, so any entry whose `name` and connection-string hostname disagree (classic vs `*.device.azure-devices.net`) would now classify as classic and could silently drop the mixed-endpoint warning. If `name` is guaranteed to track the connection-string hostname this is moot — worth confirming, since the warning is the only guardrail on that path.

3. `azext_iot/tests/dps/core/test_dps_linked_hub_unit.py` — both new tests use a single legacy entry with no `hostName`. Two cheap cases would pin the intended behavior: (a) an entry that already has `hostName` (e.g. a device hostname) is **not** overwritten by `name`, and (b) with two entries, only the expected one(s) are modified — that assertion is what would catch a future change to the backfill scope discussed above.

**Nits**

- `_ensure_linked_hub_hostnames` returns `linked_hubs`, but the only call site ignores the return value and relies on in-place mutation; returning `None` (or using the return value) would make the contract unambiguous.
- The `mock_deps` fixture actually yields the mock DPS client, so `mock_client` reads better at the call sites.

Summary: small, targeted fix with unit coverage for both reported failure modes; only the scope of the list-wide backfill and its interaction with the mixed-endpoint warning seem worth a second look before merge.

_This is an automated review and may be incomplete; please treat findings as suggestions._

cheatsheet1999:


The list-wide backfill is intentional because DPS updates resubmit the complete linked-hub list, so this safely upgrades all legacy entries and prevents the same failure on later updates. I validated it against a live DPS with classic and device legacy links: both were backfilled correctly and the mixed-endpoint warning still appeared. DPS also rejects entries whose hostname conflicts with the connection-string hostname, so that warning-suppression scenario cannot be persisted. Based on this validation, I’m keeping the focused fix and regression tests as-is.

zzcodework:
### Azure SRE Agent - reply

> The list-wide backfill is intentional because DPS updates resubmit the complete linked-hub list, so this safely upgrades all legacy entries and prevents the same failure on later updates.

That settles suggestions 1 and 2 — withdrawing both. The resubmit-the-whole-list argument makes scoping the backfill to `target_entry` the weaker choice, since the neighbouring legacy entries go back over the wire on this PUT regardless, and your live validation (classic + device legacy links backfilled, mixed-endpoint warning still emitted) covers the `_warn_mixed_endpoint_types` ordering concern I raised — with the service rejecting a `hostName`/connection-string mismatch, the branch I worried about can't be reached with persisted state.

Suggestion 3 (a test asserting an existing `hostName` is not overwritten, and a two-entry case pinning the backfill scope) is purely optional regression insurance for the behaviour you just described — your call, and no objection to merging as-is.

_Automated reply; deferring to your live validation here._

HangyiWang (approved):

PR #879, opened 2026-08-18
chore: Generate SDK for IoT DPS 2026-08-31 API

cheatsheet1999:
## Summary

Auto Generate Control Plane TypeSpec Modeless SDK on IoT DPS 2026-08-31 API
https://github.com/Azure/azure-rest-api-specs/pull/44708

`yuelinzhao/DPS-local-auth-support` serves as a feature branch based on the `dev` branch. It will be merged into `dev` once all feature requirements are met and the ARM manifest rollout is complete across all regions.

PR #880, opened 2026-08-22
feat: add --disable-local-auth to az iot dps create/update

cheatsheet1999:
## Summary
Adds `--disable-local-auth` / `--dla` to `az iot dps create` and `az iot dps update`.

When on, SAS key auth is disabled and Azure RBAC becomes the only way to authorize data plane requests.

`disableLocalAuth` is new in the stable `2026-08-31` API. Core's `iot dps` commandsare pinned to `2021-10-15`, so the extension takes over `create`, `update` and `show` against the SDK vendored in #879.

`show` is not cosmetic - `generic_update_command` wires `update` and `show` to the same getter. A getter at `2021-10-15` would drop `disableLocalAuth` from the GET, so an unrelated `az iot dps update --tags` would silently reset it.

### Note 
Most of this is lifted from `preview`

Only `iot_dps_create` and `iot_dps_update` differ: preview's ADR/identity blocks are dropped and three lines are added for `disableLocalAuth`. Keeping the structure aligned means a future `dev` -> `preview` merge stays near-trivial.

## Why setup.py changed

The vendored SDK's `_client.py` imports `get_arm_endpoints` (L18) and reads `settings.current.azure_cloud` (L70). Those are absent in **azure-mgmt-core 1.4.0** and **azure-core 1.30.0**.

The old floors (`>=1.3.0` / `>=1.24.0`) allowed installing versions without those symbols, and `azure-cli-core` only floors `azure-mgmt-core>=1.2.0`.

## Test
image
image

### DIFF
--- a/azext_iot/core/_params.py
+++ b/azext_iot/core/_params.py
@@ -1,0 +1,44 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from knack.arguments import CLIArgumentType
+from azure.cli.core.commands.parameters import (get_location_type,
+                                                get_resource_name_completion_list,
+                                                get_enum_type,
+                                                get_three_state_flag,
+                                                tags_type)
+
+from .shared import IotDpsSku
+
+
+dps_name_type = CLIArgumentType(
+    options_list=['--name', '-n'],
+    completer=get_resource_name_completion_list('Microsoft.Devices/ProvisioningServices'),
+    help='IoT Hub Device Provisioning Service name')
+
+
+def load_arguments(self, _):
+    # Arguments for IoT DPS
+    with self.argument_context('iot dps') as c:
+        c.argument('tags', tags_type)
+
+    # Direct DPS resource commands use --name -n
+    for subgroup in ['create', 'update', 'show', 'delete']:
+        with self.argument_context('iot dps {}'.format(subgroup)) as c:
+            c.argument('dps_name', dps_name_type, id_part='name')
+
+    with self.argument_context('iot dps create') as c:
+        c.argument('location', get_location_type(self.cli_ctx),
+                   help='Location of your IoT Hub Device Provisioning Service. '
+                   'Default is the location of target resource group.')
+        c.argument('sku', arg_type=get_enum_type(IotDpsSku),
+                   help='Pricing tier for the IoT Hub Device Provisioning Service.')
+        c.argument('unit', help='Units in your IoT Hub Device Provisioning Service.', type=int)
+        c.argument('enable_data_residency', arg_type=get_three_state_flag(),
+                   options_list=['--enforce-data-residency', '--edr'],
+                   help='Enforce data residency for this IoT Hub Device Provisioning Service by disabling '
+                   'cross geo-pair disaster recovery. This property is immutable once set on the resource. '
+                   'Only available in select regions. Learn more at https://aka.ms/dpsdr')
--- a/azext_iot/__init__.py
+++ b/azext_iot/__init__.py
@@ -9,6 +9,9 @@ from azure.cli.core.commands import CliCommandType
 from azext_iot.constants import VERSION
 import azext_iot._help  # noqa: F401
 
+from azext_iot.core.command_map import load_core_commands
+from azext_iot.core.params import load_core_arguments
+
 
 iothub_ops = CliCommandType(operations_tmpl="azext_iot.operations.hub#{}")
 iotdps_ops = CliCommandType(operations_tmpl="azext_iot.operations.dps#{}")
@@ -33,6 +36,8 @@ class IoTExtCommandsLoader(AzCommandsLoader):
         load_digitaltwins_commands(self, args)
         load_dps_commands(self, args)
 
+        load_core_commands(self, args)
+
         return self.command_table
 
     def load_arguments(self, command):
@@ -49,6 +54,8 @@ class IoTExtCommandsLoader(AzCommandsLoader):
         load_digitaltwins_arguments(self, command)
         load_dps_arguments(self, command)
         load_deviceupdate_arguments(self, command)
+
+        load_core_arguments(self, command)
 
 
 COMMAND_LOADER_CLS = IoTExtCommandsLoader
--- a/azext_iot/_factory.py
+++ b/azext_iot/_factory.py
@@ -23,6 +23,8 @@ __all__ = [
     "CloudError",
     "iot_hub_service_factory",
     "iot_service_provisioning_factory",
+    "iot_dps_resource_factory",
+    "resource_service_factory",
 ]
 
 
@@ -60,6 +62,32 @@ def iot_service_provisioning_factory(cli_ctx, *_):
     from azure.cli.core.profiles import ResourceType
 
     return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_IOTDPS)
+
+
+def iot_dps_resource_factory(cli_ctx, *_):
+    """
+    Factory for the provisioning service control plane client backed by the vendored SDK.
+
+    Args:
+        cli_ctx (knack.cli.CLI): CLI context.
+        *_ : all other args ignored.
+
+    Returns:
+        service_client (IotDpsClient): control plane resource for
+            working with IoT Hub Device Provisioning Service.
+    """
+    from azure.cli.core.commands.client_factory import get_mgmt_service_client
+
+    from azext_iot.sdk.dps.mgmt import IotDpsClient
+
+    return get_mgmt_service_client(cli_ctx, IotDpsClient)
+
+
+def resource_service_factory(cli_ctx, **_):
+    from azure.cli.core.commands.client_factory import get_mgmt_service_client
+    from azure.cli.core.profiles import ResourceType
+
+    return get_mgmt_service_client(cli_ctx, ResourceType.MGMT_RESOURCE_RESOURCES)
 
 
 class SdkResolver(object):
--- a/azext_iot/azext_metadata.json
+++ b/azext_iot/azext_metadata.json
@@ -1,4 +1,4 @@
 {
-    "azext.minCliCoreVersion": "2.67.0",
+    "azext.minCliCoreVersion": "2.73.0",
     "azext.isPreview": false
 }
--- a/azext_iot/core/command_map.py
+++ b/azext_iot/core/command_map.py
@@ -1,0 +1,33 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+"""
+Load CLI commands
+"""
+
+from azure.cli.core.commands import CliCommandType
+
+from azext_iot._factory import iot_dps_resource_factory
+
+core_ops = CliCommandType(operations_tmpl="azext_iot.core.custom#{}")
+
+
+def load_core_commands(self, _):
+    """
+    Load CLI commands
+    """
+    # iot dps commands
+    with self.command_group(
+        "iot dps", command_type=core_ops, client_factory=iot_dps_resource_factory
+    ) as cmd_group:
+        cmd_group.command("create", "iot_dps_create")
+        cmd_group.generic_update_command(
+            "update",
+            getter_name="iot_dps_get",
+            setter_name="iot_dps_update",
+            custom_func_type=core_ops,
+        )
+        cmd_group.show_command("show", "iot_dps_get")
--- a/azext_iot/core/custom.py
+++ b/azext_iot/core/custom.py
@@ -1,0 +1,114 @@
+# coding=utf-8
+# --------------------------------------------------------------------------------------------
+# Copyright (c) Microsoft Corporation. All rights reserved.
+# Licensed under the MIT License. See License.txt in the project root for license information.
+# --------------------------------------------------------------------------------------------
+
+from azure.cli.core.azclierror import (
+    BadRequestError,
+    CLIInternalError,
+)
+
+from azext_iot._factory import resource_service_factory
+from azext_iot.core.shared import IotDpsSku
+
+
+def iot_dps_get(client, dps_name, resource_group_name=None):
+    if resource_group_name is None:
+        return _get_iot_dps_by_name(client, dps_name, resource_group_name)
+    return client.iot_dps_resource.get(provisioning_service_name=dps_name, resource_group_name=resource_group_name)
+
+
+def iot_dps_create(
+    cmd,
+    client,
+    dps_name,
+    resource_group_name,
+    location=None,
+    sku=IotDpsSku.S1.value,
+    unit=1,
+    tags=None,
+    enable_data_residency=None,
+    disable_local_auth=None,
+):
+    cli_ctx = cmd.cli_ctx
+    _check_dps_name_availability(client.iot_dps_resource, dps_name)
+    location = _ensure_location(cli_ctx, resource_group_name, location)
+    dps_property = {"enableDataResidency": enable_data_residency}
+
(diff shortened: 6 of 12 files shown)


zzcodework:

Reviewing `full diff` (head `124b933e`), base `yuelinzhao/DPS-local-auth-support` (stacked on #879).

Context checked: the vendored client in `azext_iot/sdk/dps/mgmt/operations/_operations.py` is a JSON-in/JSON-out (`ItemPaged[JSON]` / `LROPoller[JSON]`) client, so the dict-style access in `azext_iot/core/custom.py` (`x["name"]`, `["resourcegroup"]`, `parameters["properties"]`) is correct here — `resourcegroup`/`subscriptionid` do appear in the response templates. Kwarg names (`iot_dps_description`, `provisioning_service_name`, `resource_group_name`) and the `get(provisioning_service_name, resource_group_name)` ordering all match the SDK signatures. The rationale for taking over `show` (shared getter with `generic_update_command`) is sound and the unit tests cover the two behaviors that matter (omit-on-`None` for create, preserve-on-omit for update).

**Blocking**

- `azext_iot/core/custom.py` — `iot_dps_create`: `dps_property = {"enableDataResidency": enable_data_residency}` unconditionally puts the key in the payload. Because this SDK takes a raw JSON body, there is no serializer to strip `None`, so a plain `az iot dps create -n X -g Y` now PUTs `"properties": {"enableDataResidency": null}` where core (model-based, `IotDpsPropertiesDescription`) omitted it entirely. `enableDataResidency` is immutable-once-set, so an explicit null is exactly the kind of value worth not sending. Suggest mirroring the `disable_local_auth` guard:
  ```python
  dps_property = {}
  if enable_data_residency is not None:
      dps_property["enableDataResidency"] = enable_data_residency
  ```
  The same applies to `"tags": tags` in `dps_description` — `tags: null` is sent when `--tags` is not supplied.

**Suggestions**

- `azext_iot/core/custom.py` — `iot_dps_update`: `parameters["properties"]["disableLocalAuth"] = ...` assumes the getter result always carries a `properties` object. The no-`-g` path returns a list item rather than a GET body; a `parameters.setdefault("properties", {})` is cheap insurance against a `KeyError` surfacing as an unhandled traceback.
- `azext_iot/core/custom.py` — `_get_iot_dps_by_name` raises `CLIInternalError` for both "no DPS in subscription" and "name not found". Those are user/resource conditions, not CLI faults; `ResourceNotFoundError` is the right `azclierror` type and it affects exit-code classification and error telemetry. (Inherited from `preview`, so fixing may be out of scope for this PR — flagging since the code is new on `dev`.)
- `azext_iot/core/help.py` — `patch_core_help()` is a silent no-op when `"iot dps create" not in helps` (it runs at `azext_iot.core` import time, i.e. potentially before `azure.cli.command_modules.iot._help` has populated `helps`). It also assumes the core entry ends with an `examples:` list — if that ever stops being true the appended `- name:` items produce invalid YAML at help-render time. A `"examples:" in helps[...]` guard, or asserting the append in a unit test, would make both failure modes visible instead of silent.
- `setup.py` — bumping the floors to `azure-core>=1.31.0` / `azure-mgmt-core>=1.5.0` is justified by `get_arm_endpoints` / `settings.current.azure_cloud`, but `azext_iot/azext_metadata.json` still declares `azext.minCliCoreVersion: 2.67.0`, which predates `azure-mgmt-core` 1.5.0. Worth confirming the extension actually resolves its own bundled versions on the oldest supported CLI, and bumping `minCliCoreVersion` if it does not.
- No `HISTORY.rst` entry. `--disable-local-auth` is a user-facing command-surface addition and the file currently tops out at the released `0.30.0` section; a new unreleased section would keep the changelog convention intact.

**Nits**

- `azext_iot/core/_params.py` — the `for subgroup in ['create', 'update', 'show', 'delete']` loop configures `iot dps delete`, but `command_map.py` only registers `create`, `update` and `show`, so the `delete` context is dead configuration (core keeps owning that command). Dropping it, or adding a comment saying it is intentional pre-work, avoids implying the extension owns `delete`.
- `azext_iot/tests/dps/core/` has no `__init__.py`, unlike sibling test packages (`tests/dps/enrollment/`, etc.). Pre-existing on the base branch rather than introduced here, but it will bite the moment two test modules in different dirs share a basename.
- `azext_iot/core/_params.py:load_arguments` vs `azext_iot/core/params.py:load_core_arguments` are easy to confuse at a glance; a one-line docstring on the former noting it is the verbatim CLI-core copy would help the future `dev` → `preview` merge you called out.

**Summary** — structure, SDK usage and test coverage look right, and keeping the layout aligned with `preview` is the correct call. The one thing I would fix before merge is the unconditional `enableDataResidency` (and `tags`) null in the create payload, which is a real behavior change versus core caused by moving to a raw-JSON body.

cheatsheet1999:
SRE Agent provided some useful feedback, I'll resolve them when I'm available.

cheatsheet1999:
> Reviewing `full diff` (head `124b933e`), base `yuelinzhao/DPS-local-auth-support` (stacked on #879).
> 
> Context checked: the vendored client in `azext_iot/sdk/dps/mgmt/operations/_operations.py` is a JSON-in/JSON-out (`ItemPaged[JSON]` / `LROPoller[JSON]`) client, so the dict-style access in `azext_iot/core/custom.py` (`x["name"]`, `["resourcegroup"]`, `parameters["properties"]`) is correct here — `resourcegroup`/`subscriptionid` do appear in the response templates. Kwarg names (`iot_dps_description`, `provisioning_service_name`, `resource_group_name`) and the `get(provisioning_service_name, resource_group_name)` ordering all match the SDK signatures. The rationale for taking over `show` (shared getter with `generic_update_command`) is sound and the unit tests cover the two behaviors that matter (omit-on-`None` for create, preserve-on-omit for update).
> 
> **Blocking**
> 
> * `azext_iot/core/custom.py` — `iot_dps_create`: `dps_property = {"enableDataResidency": enable_data_residency}` unconditionally puts the key in the payload. Because this SDK takes a raw JSON body, there is no serializer to strip `None`, so a plain `az iot dps create -n X -g Y` now PUTs `"properties": {"enableDataResidency": null}` where core (model-based, `IotDpsPropertiesDescription`) omitted it entirely. `enableDataResidency` is immutable-once-set, so an explicit null is exactly the kind of value worth not sending. Suggest mirroring the `disable_local_auth` guard:
>   ```python
>   dps_property = {}
>   if enable_data_residency is not None:
>       dps_property["enableDataResidency"] = enable_data_residency
>   ```
>   
>   
>       
>         
>       
>   
>         
>       
>   
>       
>     
>   The same applies to `"tags": tags` in `dps_description` — `tags: null` is sent when `--tags` is not supplied.
> 
> **Suggestions**
> 
> * `azext_iot/core/custom.py` — `iot_dps_update`: `parameters["properties"]["disableLocalAuth"] = ...` assumes the getter result always carries a `properties` object. The no-`-g` path returns a list item rather than a GET body; a `parameters.setdefault("properties", {})` is cheap insurance against a `KeyError` surfacing as an unhandled traceback.
> * `azext_iot/core/custom.py` — `_get_iot_dps_by_name` raises `CLIInternalError` for both "no DPS in subscription" and "name not found". Those are user/resource conditions, not CLI faults; `ResourceNotFoundError` is the right `azclierror` type and it affects exit-code classification and error telemetry. (Inherited from `preview`, so fixing may be out of scope for this PR — flagging since the code is new on `dev`.)
> * `azext_iot/core/help.py` — `patch_core_help()` is a silent no-op when `"iot dps create" not in helps` (it runs at `azext_iot.core` import time, i.e. potentially before `azure.cli.command_modules.iot._help` has populated `helps`). It also assumes the core entry ends with an `examples:` list — if that ever stops being true the appended `- name:` items produce invalid YAML at help-render time. A `"examples:" in helps[...]` guard, or asserting the append in a unit test, would make both failure modes visible instead of silent.
> * `setup.py` — bumping the floors to `azure-core>=1.31.0` / `azure-mgmt-core>=1.5.0` is justified by `get_arm_endpoints` / `settings.current.azure_cloud`, but `azext_iot/azext_metadata.json` still declares `azext.minCliCoreVersion: 2.67.0`, which predates `azure-mgmt-core` 1.5.0. Worth confirming the extension actually resolves its own bundled versions on the oldest supported CLI, and bumping `minCliCoreVersion` if it does not.
> * No `HISTORY.rst` entry. `--disable-local-auth` is a user-facing command-surface addition and the file currently tops out at the released `0.30.0` section; a new unreleased section would keep the changelog convention intact.
> 
> **Nits**
> 
> * `azext_iot/core/_params.py` — the `for subgroup in ['create', 'update', 'show', 'delete']` loop configures `iot dps delete`, but `command_map.py` only registers `create`, `update` and `show`, so the `delete` context is dead configuration (core keeps owning that command). Dropping it, or adding a comment saying it is intentional pre-work, avoids implying the extension owns `delete`.
> * `azext_iot/tests/dps/core/` has no `__init__.py`, unlike sibling test packages (`tests/dps/enrollment/`, etc.). Pre-existing on the base branch rather than introduced here, but it will bite the moment two test modules in different dirs share a basename.
> * `azext_iot/core/_params.py:load_arguments` vs `azext_iot/core/params.py:load_core_arguments` are easy to confuse at a glance; a one-line docstring on the former noting it is the verbatim CLI-core copy would help the future `dev` → `preview` merge you called out.
> 
> **Summary** — structure, SDK usage and test coverage look right, and keeping the layout aligned with `preview` is the correct call. The one thing I would fix before merge is the unconditional `enableDataResidency` (and `tags`) null in the create payload, which is a real behavior change versus core caused by moving to a raw-JSON body.
> 
> _This is an automated review and may be incomplete; please verify findings before acting on them._

@zzcodework SRE Agent, please review again.

zzcodework:
Reviewing `changes since 124b933e` (head `1fd3edf5`), base `yuelinzhao/DPS-local-auth-support`.

Re-reviewed on request. The single new commit (`address review feedback on dps --disable-local-auth`) touches 3 files: `azext_iot/core/custom.py`, `azext_iot/azext_metadata.json` and `azext_iot/tests/dps/core/test_dps_disable_local_auth_unit.py`.

**Blocking**

* None. The previously flagged blocking issue is resolved: `iot_dps_create` now builds `dps_property = {}` and only sets `enableDataResidency` when it is not `None`, and `tags` is only added to `dps_description` when supplied — so a plain `az iot dps create -n X -g Y` no longer PUTs `"properties": {"enableDataResidency": null}` / `"tags": null` through the raw-JSON body. The two new unit tests (`test_dps_create_omits_unset_optionals`, `test_dps_create_sends_supplied_optionals`) pin both directions of that behavior, which is exactly the right coverage.

**Suggestions**

* `azext_iot/core/custom.py` — `iot_dps_update` still does `parameters["properties"]["disableLocalAuth"] = ...` without guarding that `properties` exists. Note the new test suite does not cover this: `test_dps_update_resolves_resource_group` passes `{"properties": {}}` explicitly, so the missing-`properties` shape never reaches the function in tests. A `parameters.setdefault("properties", {})` remains cheap insurance against a raw `KeyError` traceback if the getter ever returns a body without `properties`.
* No `HISTORY.rst` entry yet. `--disable-local-auth` on `az iot dps create/update` is a user-facing command-surface addition; an unreleased section keeps the changelog convention intact. This is the one repo-convention item from the previous round that is still outstanding.
* `azext_iot/core/help.py` — `patch_core_help()` being a silent no-op when `"iot dps create" not in helps`, and its assumption that the core entry ends with an `examples:` list, are unchanged. Not a blocker, but a unit test asserting the append happened would turn both silent failure modes into visible ones.

**Nits**

* `azext_iot/azext_metadata.json` — `azext.minCliCoreVersion` bumped `2.67.0` → `2.73.0`, which addresses the mismatch with the `azure-mgmt-core>=1.5.0` floor in `setup.py`. Worth a one-line note in the PR description, since raising the floor is itself a user-visible constraint for anyone on an older `az`.
* `azext_iot/core/custom.py` — `dps_description` still always carries `"properties": dps_property` even when the dict is empty. That matches what the model-based core path serialized, so no behavior change; just flagging that it is intentional.

**Summary** — the feedback round did what it needed to: the payload-null regression is fixed and correctly tested, and the CLI-core floor is aligned with the dependency bump. Nothing blocking remains from my side. The leftover items (`properties` setdefault in `iot_dps_update`, `HISTORY.rst` entry) are non-blocking and can land here or as follow-ups.

cheatsheet1999:
@zzcodework SRE Agent, The latest commit tried to resolve the comments, please review again:
1. Blocking - none. Fixed.
2. Suggestion 1: Not applying - self-contradictory command
3. Suggestion 2: Will treat as a follow-up when it's time for release.
4. Suggestion 3: Not applying -  az iot dps create --help renders the appended example, so neither failure mode is currently active.
5. Nit 1 and Nit 2 - cosmetic so keeps no change. Unless human reviewers find it's necessary in later review.

zzcodework:
### Azure SRE Agent - reply

> @zzcodework SRE Agent, The latest commit tried to resolve the comments, please review again

There is nothing new to review — the PR head is still `1fd3edf571` ("address review feedback on dps --disable-local-auth", 2026-08-24T03:08Z), which is exactly the commit my previous comment already covered. No commit has landed since. If you pushed something, it did not reach `yuelinzhao/DPS-local-auth-support`; otherwise this is already up to date and I'll pick up the next push automatically.

On your dispositions:

1. **Blocking — agreed**, nothing blocking remains.
2. **Suggestion 1 (`parameters.setdefault("properties", {})`) — withdrawn.** You're right that it isn't worth applying: `iot_dps_update` is a generic-update command, so `parameters` is the GET'd `IotDpsDescription`, which always carries `properties`. It was defensive-only, and a `KeyError` there would mean a malformed service response rather than a user-reachable path. Dropping it.
3. **Suggestion 2 (`HISTORY.rst`) — fine as a follow-up**, just flagging that it's a repo convention a human reviewer is likely to ask for before merge, since `--disable-local-auth` is a user-facing command-surface addition.
4. **Suggestion 3 (`patch_core_help`) — accepted.** You verified `az iot dps create --help` renders the appended example, which is the evidence that matters; the `if "iot dps create" in helps` guard holding today makes the silent-no-op concern theoretical. Worth remembering only if a future `az` core release renames or restructures those help entries, since the failure would still be silent.
5. **Nits — agreed**, both were flagged as cosmetic/intentional.

No objections from my side; this is in a mergeable state pending human review.

_This is an automated reply and may be incomplete; please verify before acting on it._

azext_iot/core/_params.py
```diff
@@ -0,0 +1,44 @@
...
+
+    with self.argument_context('iot dps create') as c:
+        c.argument('location', get_location_type(self.cli_ctx),
+                   help='Location of your IoT Hub Device Provisioning Service. '
+                   'Default is the location of target resource group.')
+        c.argument('sku', arg_type=get_enum_type(IotDpsSku),
+                   help='Pricing tier for the IoT Hub Device Provisioning Service.')
+        c.argument('unit', help='Units in your IoT Hub Device Provisioning Service.', type=int)
+        c.argument('enable_data_residency', arg_type=get_three_state_flag(),
+                   options_list=['--enforce-data-residency', '--edr'],
```
HangyiWang, line 41: `--enforce` or `enable` here?

cheatsheet1999: https://github.com/Azure/azure-iot-cli-extension/blob/49ea4a35e3f3cf418e1adfb9f92e40c132af534c/azext_iot/core/_params.py#L73-L77

I copied those lines from the `preview` branch, the exising code says --enforce so it's intentional. Thanks for reviewing this.

HangyiWang: Cool, it's a bit weird, I assume it's a typo? Do you think we should keep them aligned?

HangyiWang: oh I got it, maybe one of them is API property and the other is CLI option.

HangyiWang (approved):

ketkimnaik (approved):

azext_iot/core/_params.py
cheatsheet1999: Right, `enableDataResidency` is the ARM property,  --enforce-data-residency / --edr  is our CLI option. I'd keep it as-is. The takeover is invisible to users, so `az iot dps create --edr` has to keep working - otherwise installing this extension breaks scripts that run fine on plain CLI.

PR #889, opened 2026-09-03
chore: regenerate hub mgmt SDK 10-01-preview version

cheatsheet1999:
## Summary

Regenerates the IoT Hub control-plane SDK from `2026-05-01-preview` to`2026-10-01-preview` to unblock MQTT 5 connection-profile work. SDK-only, no command, provider, or call-site changes.


zzcodework:

Reviewing `full diff` (head `39808e8`)

Scope check first: this is a pure codegen refresh of `azext_iot/sdk/iothub/mgmt` (5 files, +3907/-3887), targeting the `mqtt5-custom-topic` feature branch. I compared the generated public surface before/after rather than reading the diff line by line.

**Blocking**

None found.

**Suggestions**

- `azext_iot/sdk/iothub/mgmt/_configuration.py` (`api_version` default now `2026-10-01-preview`): I could not find `2026-10-01-preview` in the public spec repo. `Azure/azure-rest-api-specs@main` lists `.../iothub/resource-manager/Microsoft.Devices/IoTHub/preview/` up to `2026-05-01-preview`, and the version enum in that RP's `main.tsp` stops at `v2026_05_01_preview`. If the spec PR for `2026-10-01-preview` has not merged and the version is not deployed in ARM yet, every control-plane call from this SDK will be rejected on `api-version`. That is fine while this sits on a feature branch, but please confirm the spec is published/deployed before this rides into `dev`/`preview`. Worth stating in the PR description which spec commit / TypeSpec version this was generated from, since the repo carries no `tsp-location.yaml` or autorest config to pin it.

- `azext_iot/sdk/iothub/mgmt/__init__.py`: the new `types.py` is only reachable as `from azext_iot.sdk.iothub.mgmt import types`; it is not re-exported from the package `__all__`. Consumers of the MQTT 5 work will want `ConnectionProfile` / `IotHubProperties`, so consider exporting it (or confirm the generator intentionally leaves it out).

**Nits**

- No `HISTORY.rst` entry. Acceptable for an SDK-only commit onto a feature branch, but the eventual PR into `dev`/`preview` should carry the changelog line for the MQTT 5 connection-profile support this unblocks.

**Summary — what I verified, so reviewers do not have to re-do it**

- *No breaking change to the operation surface.* I extracted every public method per class from `operations/_operations.py` at `mqtt5-custom-topic` and at `39808e8` and diffed the name sets: identical. The apparent removals in the raw diff (`test_all_routes`, `export_devices`, `import_devices`, `verify`, `begin_manual_failover`, …) are `@overload` stubs being reshaped, not deletions.
- *No runtime type change for callers.* The new `types.py` models are `typing_extensions.TypedDict`, not azure-core `Model`, so the return values are still plain dicts at runtime even though signatures moved from `JSON` to `_types.IotHubDescription`. Existing dict-style access would keep working — and in any case I found no import of `azext_iot.sdk.iothub.mgmt` outside the SDK package itself, so the "no call-site changes" claim in the description holds.
- *No dangling import.* The regenerated `_operations.py` adds `from .._utils.utils import prep_if_match, prep_if_none_match`; `_utils/utils.py` already exists on the base branch and defines both. The `_utils/serialization.py` additions are the standard `duration-seconds-*` / `duration-milliseconds-*` codegen boilerplate.
- The substantive API delta is the new `ConnectionProfile = Literal["Classic", "MqttV5"]` plus the related `IotHubProperties` fields, which matches the stated purpose.

This is an automated review and may be incomplete; it does not replace human review.

HangyiWang (approved):

PR #890, opened 2026-09-05
chore: regenerate DPS control plane SDK on 2026-06-01-preview

cheatsheet1999:
## Summary
Regenerates `azext_iot/sdk/dps/mgmt` from TypeSpec at **2026-06-01-preview**

## Changes
At 2026-06-01-preview the DPS -> ADR namespace property is `@visibility(Lifecycle.Read)` so the client cannot write the link at all. The parameters that depended on it are removed rather than left broken:
- `--ns-resource-id` / `--ns-identity-id` on `az iot dps create` / `az iot dps update`
- `_build_dps_adr_properties` and its help examples
- `test_dps_adr_namespace_unit.py`, `test_adr_infra_int.py`, and `TestADRDeviceLifecycle` -
  all built their scenario through those flags, and `az iot adr ns device` has no `create`, so there is no alternative path.


zzcodework:

Reviewing `full diff` (head `ba5b9cf`)

Scope: DPS control-plane SDK regenerated to `2026-06-01-preview` (new Python code generator), plus removal of the DPS-side ADR namespace linking (`--ns-resource-id` / `--ns-identity-id`) and the tests that depended on it. I skimmed the generated `azext_iot/sdk/dps/mgmt/**` and focused on the hand-written surface. The URL plumbing in the regenerated client checks out — `_client.py` sets `ARMPipelineClient(base_url="{endpoint}")` and every operation supplies `path_format_arguments["endpoint"] = self._config.base_url`, so request URLs still resolve correctly. `pyflakes` on the changed non-SDK files is clean (no dangling imports after `DeviceRegistryNamespaceAuthenticationType` / `_build_dps_adr_properties` were dropped; `RequiredArgumentMissingError` is still used elsewhere in `core/custom.py`), and the IoT Hub ADR params in `core/params.py` are untouched, so only the DPS surface is affected.

**Blocking**

1. `setup.py` — the regenerated SDK raises the minimum `azure-core` / `azure-mgmt-core` versions, but the pins were not updated.

   `azext_iot/sdk/dps/mgmt/_client.py` now imports and uses:
   - `from azure.mgmt.core.tools import get_arm_endpoints`
   - `from azure.core.settings import settings` → `settings.current.azure_cloud`

   Current pins (`setup.py`, `DEPENDENCIES`) are `azure-core>=1.24.0,<1.40.0` and `azure-mgmt-core>=1.3.0,<2.0.0`. I checked the published wheels directly:

   | package | 1.3.0 / 1.24.0 | 1.4.0 / 1.30.0 | 1.5.0 / 1.32.0 |
   |---|---|---|---|
   | `azure.mgmt.core.tools.get_arm_endpoints` | absent | absent | present |
   | `azure.core.settings` `azure_cloud` | absent | absent | present |

   So `get_arm_endpoints` requires `azure-mgmt-core>=1.5.0` and `settings.current.azure_cloud` requires `azure-core>=1.32.0`. A resolver that honours the declared floors (or an existing environment already satisfying them) can install `azure-mgmt-core 1.3.0`, and the extension then fails at **import time** with `ImportError: cannot import name 'get_arm_endpoints'` the moment the DPS client module is loaded — not a runtime edge case, a hard load failure. Please bump both floors (`azure-core>=1.32.0`, `azure-mgmt-core>=1.5.0`) so the declared dependencies match what the generated code actually needs.

**Suggestions**

2. `HISTORY.rst` — the DPS ADR flags were an announced feature and their removal is unannounced. The `0.32.0b1` section states: *"DPS creation now supports linking an ADR namespace with `--ns-resource-id` and `--ns-identity-id` arguments."* This PR deletes exactly those two options from the `iot dps` argument context (`azext_iot/core/params.py`) and the corresponding parameters from `iot_dps_create` / `iot_dps_update` (`azext_iot/core/custom.py`), plus the two help examples in `azext_iot/core/help.py`. Anyone who scripted against the `0.32.0b1` preview gets `unrecognized arguments: --ns-resource-id`. The in-flight `0.32.0b2 (Preview)` section should carry a note that the DPS-side ADR namespace linking was withdrawn, even though this targets a feature branch — otherwise it will be missing when the branch merges.

3. `tox.ini` — the `[testenv:ADR-device-int]` environment is now empty of purpose. It is described as *"ADR device lifecycle tests (preview SDK with CSR)"* and installs `git+https://github.com/Azure/azure-iot-sdk-python.git@feature/dps-csr-preview`, but the only test that needed it — `TestADRDeviceLifecycle.test_adr_device_provision_crud_and_revoke`, together with its `DEVICE_PREVIEW_SDK` skipif — is deleted in this PR. What remains in `azext_iot/tests/adr/test_adr_device_int.py` is `TestADRDeviceEdgeCases`, whose docstring you correctly updated to drop the "don't require DPS or the preview SDK" caveat. The env now pays the cost of building a preview SDK from a git branch for tests that never touch it. Either delete the env or drop the `git+` dependency and the CSR wording. Same for `test_adr_infra_int.py` being removed: its docstring called it *"the only test in ADR-int that exercises DPS integration + credential sync"*, so ADR-int no longer covers credential sync at all — worth confirming that is intentional and not just collateral from unwinding the DPS link.

**Nits**

4. `azext_iot/_factory.py` — the `endpoint=` → `base_url=` rename is right for the new signature (`base_url` is positional-or-keyword; `cloud_setting` is keyword-only), and `azext_iot/tests/test_factory_unit.py` was updated to match. One behavioural note: the new client only honours `base_url` when it is truthy, otherwise it silently falls back to `get_arm_endpoints(settings.current.azure_cloud)`, i.e. **public cloud**, which does not track the CLI's `az cloud set`. Today `_get_arm_endpoint` returns `cli_ctx.cloud.endpoints.resource_manager`, which is always populated, so there is no live bug — but sovereign-cloud correctness now rests on that value never being empty. The explicit `credential_scopes=` you pass is likewise what keeps the client from defaulting to public-cloud scopes.

**Summary**

The regeneration itself looks mechanically sound and the ADR-removal is consistent across command, params, help and tests with no dangling references. The one thing I would not merge without is the dependency-floor bump in `setup.py` (item 1) — the new generated client uses APIs that do not exist in the minimum versions the package declares. Items 2 and 3 are cleanup that is easy to lose once the feature branch merges.

_This is an automated review and may be incomplete; the generated `azext_iot/sdk/dps/mgmt/**` files were reviewed only at the plumbing level, not line by line._

cheatsheet1999:
@zzcodework SRE agent - 
1. Known issue, I planned to fix this while syncing `dev` with `preview`. It has now been fixed after being flagged.
2. `History.rst` will be updated once the release is ready.
3. Out of scope for this PR, the environment was already inactive in preview
4. cli_ctx.cloud.endpoints.resource_manager can't be falsy, CloudEndpoints.__getattribute__  raises  CloudEndpointNotSetException on None rather than returning empty, so if not base_url: is unreachable from the factory. Therefore no change.

zzcodework:

Reviewing `changes since ba5b9cf` (head `ec8224c`)

The only new commit is `ec8224c` "sync on dependencies version", and the only changed file is `setup.py`:

```
- "azure-core>=1.24.0,<1.40.0",
- "azure-mgmt-core>=1.3.0,<2.0.0",
+ "azure-core>=1.31.0,<2.0.0",
+ "azure-mgmt-core>=1.5.0,<2.0.0",
```

This resolves the blocking item from the previous review. I re-verified the floors against the published wheels rather than trusting my earlier number:

| package | first version providing the API the generated client needs | declared floor |
| --- | --- | --- |
| `azure-mgmt-core` | `1.5.0` — `def get_arm_endpoints` appears in `azure/mgmt/core/tools.py` in 1.5.0 (absent in 1.3.0 / 1.4.0) | `>=1.5.0` ✅ |
| `azure-core` | `1.31.0` — `settings.azure_cloud` and the `AzureClouds` export both appear in 1.31.0 (absent in 1.28–1.30) | `>=1.31.0` ✅ |

`azure-mgmt-core==1.5.0` itself declares `Requires-Dist: azure-core >=1.31.0`, so the two floors are mutually consistent and neither can resolve to a version that fails at import. **My earlier "bump to `azure-core>=1.32.0`" was one minor version too high — `1.31.0` is the correct floor.**

**Blocking**

None.

**Nits**

1. `setup.py` — the `azure-core` upper bound was also widened from `<1.40.0` to `<2.0.0` in the same commit. That is consistent with the neighbouring `azure-mgmt-core<2.0.0` and with how `azure-cli` itself caps `azure-core`, so it is defensible; just flagging that it is a separate change from the floor bump and removes the guard against an unreleased 1.4x introducing a regression. `azure-identity<1.18.0` and `azure-storage-blob<12.30.0` keep their tighter caps, so the file is now slightly inconsistent in style.

**Summary**

Clean incremental change that does exactly what was needed and nothing more. The previously raised `HISTORY.rst` and `tox.ini` items are unchanged in this iteration and were answered in the discussion thread.

_This is an automated review and may be incomplete; only the delta since `ba5b9cf` was reviewed._

zzcodework:
### Azure SRE Agent - reply

> 1. Known issue, I planned to fix this while syncing `dev` with `preview`. It has now been fixed after being flagged.

Confirmed in `ec8224c`, and thanks for the correction on the exact number — I re-checked the published wheels and **`azure-core>=1.31.0` is the right floor, not the `>=1.32.0` I asked for**: `settings.azure_cloud` and the `AzureClouds` export first appear in 1.31.0 (absent in 1.28–1.30). `azure-mgmt-core==1.5.0` also declares `Requires-Dist: azure-core >=1.31.0`, so your pins are exactly consistent. My original figure was one minor version too high — withdrawn.

> 2. `History.rst` will be updated once the release is ready.
> 3. Out of scope for this PR, the environment was already inactive in preview

Both reasonable; no further concern from me.

> 4. `cli_ctx.cloud.endpoints.resource_manager` can't be falsy, `CloudEndpoints.__getattribute__` raises `CloudEndpointNotSetException` on None rather than returning empty, so `if not base_url:` is unreachable from the factory. Therefore no change.

Agreed, and that matches what I wrote — I filed it as a Nit precisely because there is no live bug: `_get_arm_endpoint` ([`azext_iot/_factory.py`](https://github.com/Azure/azure-iot-cli-extension/blob/ec8224ccb05af059014a74c3a019058f3f6cc1e8/azext_iot/_factory.py#L59)) always yields a populated endpoint, so `_client.py`'s `if not base_url:` fallback to `get_arm_endpoints(settings.current.azure_cloud)` never runs through the CLI path. The note was only that sovereign-cloud correctness now depends on that invariant holding inside generated code you don't control, since the generated default is public cloud rather than the CLI's `az cloud set`. No change requested.

_Automated reply; deferring to you on all four points._

PR #891, opened 2026-09-08
feat: hub/dps upgrade for adr ns

HangyiWang:
## Scope and stacked dependency

This PR is based on Azure/azure-iot-cli-extension#840 (`users/hangyiwang/adr-ignite-preview`), not directly on `preview`. ADR namespace commands, restored Registry Devices, namespace-side linking, and retirement of resource-side namespace-binding options are inherited from that parent PR.

The additional changes here include **both management and data-plane migrations**. Hub data-plane clients and non-CSR registration are not unchanged.

## SDK versions

| Surface | API version |
|---|---|
| IoT Hub management | `2026-10-01-preview` |
| DPS management | `2026-06-01-preview` |
| IoT Hub service and device REST | `2026-11-01-preview` |
| DPS service and device REST | `2026-11-02-preview` |

Management uses the validated Central US EUAP routing contract. Data-plane clients preserve discovered service/device hostnames and use the appropriate service or device authentication boundary.

## Main changes beyond the ADR parent

- Migrate Hub management CRUD/upsert, ETags, identities, endpoints/routes, and related consumers to the modeless contract. Preserve unspecified upsert settings and filter read-only fields before writes.
- Add handwritten Hub compatibility adapters outside generated operations for query, messaging, C2D settlement, uploads, direct methods/digital twins, configurations, and jobs.
- Strengthen state export/import around authoritative identity/twin reads, validated replacement payloads, preserved routing/authentication fields, and private ARM staging.
- Migrate DPS enrollment, queries, pagination, authentication, conditional updates, and service-error handling. Add namespace/CA/certificate-policy references to existing enrollment commands.
- Use REST registration for both CSR and non-CSR paths. Validate CSR encoding/signature/Common Name, support encrypted X.509 keys, and preserve the service's certificate-chain JSON without inventing TLS-ready encoding or ordering guarantees.
- Add the public `az iot device registration operation-status` command. Explicit registration timeouts bound the isolated worker after preliminary discovery/bootstrap authentication; accepted operation IDs remain available for follow-up after a timeout.
- Compose Hub device-identity update intent validation with inherited target/authentication validators. Reject normalized read-only mutation paths before the getter while preserving allowed partial updates.
- Reuse the existing owned registration/CSR scenarios for Registry Device/authentication/key/capability correlation and certificate revocation. The namespace-SAMI prerequisite is granted only by the test fixture at its own namespace scope; this is not a new production CLI auto-grant.

## Compatibility and review boundaries

The public surface **does change** through `operation-status` and new enrollment/registration parameters. Parent ADR removals should not be counted again as changes introduced here.

Service credentials remain separate from device bootstrap credentials. Worker progress carries only bounded, allowlisted correlation metadata; it must not disclose keys, CSR material, authorization headers, or arbitrary tracebacks. A timeout before service acceptance cannot invent an operation ID.

Owned integration controllers retain regular/Entra, service-SAS, and local-auth ordering where required. Workflow service matrix entries run independently; selected services run complete cohorts rather than hidden filtered subsets.

## Validation

- Final `0b3355a9` qualification: **10,432 unique unit cases** from the complete earlier cohort plus whole affected-file replays. The feedback refresh passed **150 related cases**; all previous coverage for the changed monitor module was discarded and replaced with fresh data. Every other production/prerequisite file was verified unchanged. This is not a fresh full-suite invocation at the current head.
- Entire-PR introduced/changed handwritten runtime relative to ADR parent `9c9de498`, including multiline executable origins: **1,871/1,871 statements (100.00%)** and **649/650 branch outcomes (99.85%)**. Generated SDK is measured separately, not used to inflate this result. No coverage exclusions or branch suppressions were added.
- Flake8 and pylint passed. Independent review covered the worker protocol/deadline changes, native Hub validation, and feedback repair. The current head's Windows and macOS checks passed.
- CI follow-up `43e8ff29` moves the recovery example to its own command, generates synthetic fixture credentials, carries parent dependency paths into the isolated test peer, and inherits the corrected ADO interpreter setup. Production worker isolation and narrow timeout handling are unchanged. **662 focused cases passed** before the full qualification above.
- A subsequent scanner pass cleared three of four locations but still matched the worker test's literal sentinel tuple. Test-only follow-up `7ab18778` generates every sensitive sentinel while retaining all disclosure assertions; **86 cases passed**. Production and integration-test sources remain identical to `43e8ff29`, so existing live cohorts are not restarted.
- Full remote workflow [35480848526](https://github.com/Azure/azure-iot-cli-extension/actions/runs/35480848526), on `82f927cc` / Python 3.13 / Central US EUAP, completed with failures: ADR **13 passed / 12 failed / 6 fixture-dependent skips**; HubControl **27 passed / 1 failed**; HubData **48 passed / 4 failed**. Required Hub nodes all ran; no filtering or new failure-suppressing skips were introduced.
- DPS regular phase: **33 passed / 2 failed**; later authentication phases were blocked. Both CSR submissions reached `assigned`. The default case subsequently observed `assigning` from a separate operation-status read. The deadline case reached RegistryDevice certificate revocation, whose accepted action was followed by a subscription-scoped polling `403 AuthorizationFailed`; uncertain revocation intentionally quarantined the owned namespace/Hub/DPS cleanup. This is not a CSR-submission authorization failure.
- Full DPS follow-up on `43e8ff29`: [run 35483083130](https://github.com/Azure/azure-iot-cli-extension/actions/runs/35483083130) completed with **33 passed / 2 failed** in the regular phase; later phases were blocked. Both CSR cases passed issuance and operation-status assertions, then failed RegistryDevice certificate-revocation polling with `403 AuthorizationFailed`. The earlier status-regression observation did not recur. Cleanup remains quarantined for the owned namespace/Hub/DPS. No failure has been hidden by a skip, scanner suppression, or a polling-success fallback.
- Runtime qualification at `0b3355a9` had **43/43 successful PR checks**; the separate live workflows completed with failures. The subsequent documentation-only cleanup at `deb4ab66` shortens CONTRIBUTING.md, removes two unused RST guides and their docstring reference, and does not change runtime or test logic. Green PR checks do not qualify failed live cohorts.
- ADR failures match the parent branch except that SU linking passed here. Hub failures include system-identity Storage authorization, unfiltered job-list rejection, and Entra digital-twin invocation `401002`. In the SAS phase, key-auth root/component invocations passed; the subsequent login iteration failed.
- The parent branch's HubData run exposed an inherited feedback-monitor false-success defect also present here, plus an adjacent mixed-device batch defect. Both are now fixed and covered by offline regressions and strict review; the corresponding combined live feedback case passed the first round.
- Complete HubData follow-up on `0b3355a9`: [run 35493919259](https://github.com/Azure/azure-iot-cli-extension/actions/runs/35493919259), finished 2026-09-20 09:00 UTC. Entra: **42 passed / 2 failed**; SAS: **7 passed / 1 failed**; total **49 passed / 3 failed**, with all required nodes executed and no skips or setup/teardown failures. Feedback and both managed-identity Storage scenarios passed. Remaining failures are unfiltered job listing (`400 ArgumentInvalid`) and login-auth digital-twin invocation (`401002`) in both phases; key-auth invocations passed in the SAS phase. All phase cleanup receipts are complete. Unchanged services were not redundantly rerun.
- HubData cleanup completed. Separate read-only HubControl reconciliation confirmed all previously uncertain non-deployment roots absent; six deployment records remain. Original failed cohort and cleanup receipts are preserved.


zzcodework:

Reviewing `full diff` (head `ae6c88d`), first automated pass on this PR.

**Scope note up front:** this is 327 changed files / ~+50k lines over 188 commits, base `users/hangyiwang/adr-ignite-preview`. I did not attempt to review the regenerated `azext_iot/sdk/**` output or the ~60 test modules line by line. What I actually read: the new hand-written modules (`azext_iot/dps/services/*`, `azext_iot/iothub/_client.py`, `_authentication.py`, `_payload.py`, `azext_iot/core/_validators.py`, `azext_iot/core/transforms.py`), the reworked `azext_iot/common/embedded_cli.py`, the DPS registration provider, and the generic-update path in `azext_iot/core/custom.py`. Copilot has already filed four inline comments (registration retry policy, Hub `job list` v2 query compat, Hub service `_version.py` bump, README/command-surface mismatch) — I am not restating those.

**Blocking**

None found in the areas I reviewed.

**Suggestions**

1. `azext_iot/core/_validators.py` — `validate_dps_capacity_update` reports a misleading error when `sku` is absent entirely. The `skus or [None]` / `capacities or [None]` fallbacks mean that when the generic update payload has no `sku` (e.g. `az iot dps update --remove sku`), the loop still runs once with `sku=None`, `capacities` is `[]`, and `validate_dps_unit(None, "sku.capacity")` raises `--sku.capacity must be an integer greater than or equal to 1.` Failing is correct here (`dps_capacity_edited` was set, so the user did touch the SKU), but the message points at a value the payload does not contain. Distinguishing "sku/capacity is missing after this edit" from "capacity value is invalid" would make `--remove sku` diagnosable. Same applies to `--remove sku.capacity`.

2. `azext_iot/dps/providers/device_registration.py` — the transient-retry branch in `_wait_for_registration` (the `except HttpResponseError` that synthesizes a `202` for 408/429/5xx and keeps polling) is unreachable on the `--timeout` path. When `request_deadline is not None`, `_invoke_operation` routes through `call_with_deadline`, which already retries exactly `(408, 429, 500, 502, 503, 504)` internally and, on budget exhaustion, raises `RegistrationTimeoutError` — an `AzureConnectionError`, not an `HttpResponseError`. Every non-retriable status goes through `handle_service_error` and raises there. So that branch only ever serves the no-`--timeout` path. Not a bug, but it reads as if it covers both; a comment (or an assert that `request_deadline is None` there) would prevent someone later "fixing" the retriable-status list in only one of the two places, which would silently diverge the two paths.

**Nits**

3. `azext_iot/dps/providers/device_registration.py` — in `_perform_registration`, `self._request_options(deadline)` is called and its return value discarded. The only reason it is there is the `_remaining(deadline)` side effect (refuse to report success after the deadline passed), which is a genuinely important post-success budget check but is invisible at the call site — it looks like dead code and is a plausible target for a future cleanup. Calling `_remaining(deadline)` directly with a one-line comment expresses the intent without the indirection.

**Checked and clean** (recording these so they are not re-raised):

- The recurring `azure-core` / `azure-mgmt-core` dependency-floor problem does **not** apply here. This PR regenerates the DPS mgmt SDK and adds `sdk/dps/device` + `sdk/dps/service`, but `setup.py` on this branch already declares `azure-core>=1.31.0,<1.40.0` and `azure-mgmt-core>=1.5.0,<2.0.0`, which cover `get_arm_endpoints` (mgmt-core 1.5.0) and `settings.current.azure_cloud` (core 1.31.0). Unchanged by the diff, and correct.
- `HISTORY.rst` has a substantial entry for this work and `azext_iot/constants.py` `VERSION` is unchanged at `1.0.0b1`, consistent with the base branch heading.
- `_validate_attestation_params` / `_validate_explicit_auth_inputs` both reject a lone `--certificate-file-path` or `--key-file-path` (`if not certificate_file or not key_file: raise RequiredArgumentMissingError(CERTIFICATE_FILE_ERROR)`), so the half-configured X.509 case Copilot raised against the sibling PR #840 is already handled on this branch.
- `azext_iot/common/embedded_cli.py` — the rewrite fixes a real pre-existing bug: the old gate `(capture_stderr is None and self.capture_stderr) or capture_stderr` meant an explicit `capture_stderr=False` could not turn off an instance-level `capture_stderr=True`. The new `capture = self.capture_stderr if capture_stderr is None else capture_stderr` is right, and restoring the exception handler in `finally` fixes the leak when `az_cli.invoke` raised. `as_json()` now calling `raise_for_error()` first is a behaviour change, but a safe one: previously a failed command produced a JSON-decode error that also interpolated raw output into the message.
- `azext_iot/dps/services/_registration_protocol.py` — the worker protocol bounds output (`MAX_OUTPUT_BYTES`), rejects duplicate JSON fields, requires newline-terminated frames, allows at most one progress frame, and constrains operation IDs to a character allowlist before echoing them into a suggested `az` command. The deliberate "never interpolate arbitrary service output into a suggested command" handling in `timeout_message` is the right call.
- `azext_iot/iothub/_client.py` — `upload_file_to_container` sets `allow_redirects=False` and rejects userinfo/fragment in the endpoint; `_call` defaults `retry_total=0` so uncertain mutations are not replayed.

---
This is an automated review and may be incomplete — given the size of this change, treat the "checked and clean" list as the boundary of what was actually verified rather than as sign-off on the whole PR.

zzcodework:

Reviewing `changes since ae6c88dd` (head `52d2fd7`)

The increment is a single commit, `Correct Registry Device availability in preview README`, touching only `README.md` (+5/-2). No code, tests, or command surface changed.

**Blocking**

None.

**Suggestions**

None.

**Nits**

None.

**Verification performed**

The commit is a documentation claim, so the only thing worth checking is whether it matches the real command surface at this head. It does:

- `README.md` — the new `device : Manage Registry Devices in a Device Registry namespace.` line under `az iot adr ns` subgroups matches `helps["iot adr ns device"]` in `azext_iot/adr/_help.py` word-for-word, and the insertion is in the correct alphabetical position between `ca` and `group`.
- `README.md` — the replacement prose asserts `az iot adr ns device` supports `create, show, list, update, delete, wait`. `azext_iot/adr/command_map.py` registers exactly those six under the `iot adr ns device` group (`adr_registry_device_create/_show/_list/_update/_delete/_wait`).
- `README.md` — the claim of `auth`, `attribute`, and `capability` subgroups matches the three `iot adr ns device auth` / `... attribute` / `... capability` command groups in `command_map.py`; no fourth device subgroup exists, so the list is complete.
- The narrowing of the old sentence is correct: the previous text said both Namespace Device *and* Registry Device commands are unexposed, which was stale — only the Namespace Device surface remains unexposed, and there is no `namespaceDevice` registration anywhere in `command_map.py` or `_help.py`, so that half of the sentence is still accurate.

Summary: a docs-only correction that brings the preview README in line with the shipped `az iot adr ns device` surface. Verified accurate against `command_map.py` and `_help.py` at head `52d2fd7`; nothing to change.

This is an automated review and may be incomplete; it covers only the incremental change since `ae6c88dd`, not the full PR.

PR #892, opened 2026-09-08
feat: adr ns September release API alignment

HangyiWang:
## Summary

Align the ADR namespace and DPS command surfaces with the September release API versions. This branch is based on `preview` and intentionally includes only features supported by the selected stable management contracts.

| Service / plane | API version | Change |
| --- | --- | --- |
| ADR management | `2026-04-01` | Regenerated modeless, synchronous SDK |
| DPS management | `2026-08-31` | Regenerated modeless, synchronous SDK |
| IoT Hub management | `2026-05-01-preview` | Unchanged from `preview` |
| IoT Hub service data | `2024-03-31` | Unchanged from `preview` |
| DPS service data | `2025-07-01-preview` | Unchanged from `preview` |

## Command compatibility

- Retain 11 ADR commands: namespace `create`, `show`, `list`, `update`, `delete`, `wait`, and `migrate`; system-assigned identity `show`, `assign`, `remove`, and `wait`.
- Support namespace tags and simple messaging endpoint configuration. Remove unsupported legacy credential, policy, and device commands, their arguments/help, and obsolete tests. Newer CA, registry-device, group, job, report, linking, UAMI/outbound-identity, and ADR Software Updates features are outside this branch's scope.
- Remove DPS namespace-association flags (`--ns-resource-id` / `--ns-id` and `--ns-identity-id`) and their payload construction because `deviceRegistryNamespace` is absent from the stable DPS contract.
- Preserve DPS managed identities, linked-hub authentication, certificates, policies, and discovery, along with the existing `az iot du` command surface.


zzcodework:

Reviewing `full diff` (head `3207439`)

Scope note: 76 files, but ~40k of the delta is regenerated SDK under `azext_iot/sdk/deviceregistry/**` and `azext_iot/sdk/dps/mgmt/**`. I reviewed the hand-written surface (`azext_iot/adr/**`, `_factory.py`, `common/auth.py`, `core/custom.py`, `core/params.py`, `core/help.py`, `tox.ini`, tests) and spot-checked the generated clients for host-code implications. I did not re-review the generated operation bodies.

**Blocking**

1. `setup.py` dependency floors were not raised for the regenerated SDKs — the extension can install a dependency set that fails at import time.

   Both new clients pull APIs that do not exist in the declared minimums:
   - `azext_iot/sdk/deviceregistry/_client.py` and `azext_iot/sdk/dps/mgmt/_client.py` import `from azure.mgmt.core.tools import get_arm_endpoints`, and use `settings.current.azure_cloud` / `from azure.core import AzureClouds`.
   - `setup.py` (unchanged by this PR) still declares `azure-core>=1.24.0,<1.40.0` and `azure-mgmt-core>=1.3.0,<2.0.0`.

   I verified against the published wheels rather than from memory:
   - `get_arm_endpoints` is absent from `azure/mgmt/core/tools.py` in azure-mgmt-core **1.4.0** and first appears in **1.5.0**.
   - `azure_cloud` is absent from `azure/core/settings.py` in azure-core **1.30.0** and first appears in **1.31.0**.

   So a resolver that picks e.g. azure-mgmt-core 1.4.0 satisfies the declared constraint and then blows up with `ImportError` the moment `adr_service_factory` / `iot_service_provisioning_factory` runs. Suggested floors (tight, not over-strict — azure-mgmt-core 1.5.0 itself declares `Requires-Dist: azure-core>=1.31.0`):

   ```
   "azure-core>=1.31.0,<1.40.0",
   "azure-mgmt-core>=1.5.0,<2.0.0",
   ```

   Note this is the same class of issue that was raised and fixed on #890, but that fix lives on `yuelinzhao/sync-dps-rbac-and-tls13-fixes`, not on `preview` — so it does not cover this PR.

**Suggestions**

2. `azext_iot/adr/providers/namespace.py` — `delete()` classifies the "namespace not empty" case with `if "NamespaceNotEmpty" in str(error)`. Substring-matching the stringified exception is fragile (it also matches an unrelated error whose message merely embeds that token). `HttpResponseError` exposes the structured code; `getattr(error, "error", None)` → `.code == "NamespaceNotEmpty"` is the more durable check, with the substring test only as a fallback.

   Related: the two `logger.warning` calls that previously ran before deletion are gone. The child-resource warning is now obsolete, but the second one ("deletion will fail if DPS or IoT Hub instances are linked — unlink them first") was the actionable hint for the most common failure, and the new error text only mentions child resources. Consider folding the linked-DPS/Hub case into the raised message.

3. `azext_iot/adr/params_adr_management.py` — the `--messaging-endpoints` help says "Replaces the endpoint map; use `{}` to clear it", but `update` goes through `namespaces.begin_update` (PATCH) with `properties.messaging.endpoints`. Whether the RP replaces or merges that map is a service-side decision; if it merges, "replaces" is wrong and `{}` will not clear. The unit tests only assert the request body, so they cannot catch this. Worth one assertion in `test_adr_namespace_crud_int.py` that sets two endpoints, then sets one, and verifies the other is gone.

4. `azext_iot/core/help.py` — the DPS examples were correctly stripped of `--ns-resource-id` / `--ns-identity-id`, but the `iot hub create` block above still documents both. That is consistent with `azext_iot/core/params.py`, which retains `adr_ns_id` / `adr_ns_identity_id` under the hub context, and with the PR description (Hub stays on `2026-05-01-preview`) — so it looks intentional. Flagging only so the asymmetry (Hub can link a namespace, DPS can no longer) is a deliberate choice and not an oversight, since users will hit it as an inconsistency.

**Nits**

5. `azext_iot/adr/params_adr_management.py` — `--messaging-endpoints` is registered for both `create` and `update` with identical help. The "use `{}` to clear it" clause is meaningless on `create` (there is nothing to clear); a create-specific string would read better.

Everything else looks coherent: the command-surface removal is clean (no dangling imports of `commands_credential` / `commands_policy` / `commands_device` / `providers.policy`, `_build_dps_adr_properties` and `DeviceRegistryNamespaceAuthenticationType` fully removed), `_help.py` matches the new `migrate` / `ns identity` surface exactly, the deferred `load_*` imports in `azext_iot/__init__.py` are a real startup-time win, moving off a module-level `AzureCliCredential()` to `get_cli_credential(cli_ctx, ...)` is the right fix and correctly threads `subscription_id`, the `endpoint=` → `base_url=` switch in `iot_service_provisioning_factory` matches the regenerated client signature and keeps sovereign clouds working via `_get_arm_endpoint`, and the removed `ADR-device-int` tox env is not referenced by any workflow (`int_test.yml` uses `ADR-int`). Unit coverage for `_messaging_properties`, `_clean_migrate_resource_ids`, `no_wait`, and the identity paths is solid. `HISTORY.rst` and the `0.33.0b1` version bump are present.

I did not repeat the over-escaped-JSON point Copilot already made on `test_adr_namespace_crud_int.py`.

zzcodework:

Reviewing `changes since 3207439` (head `9fcddd6`)

The increment is a single-line edit to the changelog:

- HISTORY.rst — removed `* Existing ``az iot du`` commands remain available.` from the `0.33.0b1 (Preview)` **API compatibility** section.

No findings.

**Blocking:** none.

**Suggestions:** none. Dropping the line is reasonable — a changelog entry stating that an untouched command group is still present adds no signal, and the `az iot du` behaviour that *did* change in this PR (`azext_iot/deviceupdate/commands_update.py`, `azext_iot/deviceupdate/providers/storage.py` now reusing the hosting CLI login for update-staging storage) is already covered by the "Reused the hosting Azure CLI's login for Hub, DPS, ADR, and update-staging storage clients" entry under **Authentication and diagnostics**.

**Nits:** the PR description still says the change "Preserve[s] ... the existing `az iot du` command surface"; that is still true, so no action needed unless you want description and changelog to read identically.

Summary: changelog-only increment, no functional change since the last reviewed iteration. Earlier findings on this PR (notably the `azure-core` / `azure-mgmt-core` minimum-version floors in `setup.py` on `preview` versus the regenerated SDK's `get_arm_endpoints` / `settings.current.azure_cloud` imports) are unaffected by this commit and still stand.

_This is an automated review and may be incomplete._

cheatsheet1999 (approved):

PR #893, opened 2026-09-10
feat: merge dev into preview and add DPS RBAC

cheatsheet1999:
## Summary
Merges dev into preview and ports the DPS RBAC feature so that preview remains a strict superset of dev.

## Changes
All conflicts were resolved with --ours, preserving preview's azext_iot/core. The DPS RBAC changes were then ported manually:

* `core/{custom,params,help}.py` - add `--disable-local-auth` (`--dla`) to DPS create/update
* `setup.py` - adopt `dev`'s dependency minimums while preserving `preview`'s upper bounds
* Aligns the minimum Azure CLI (`2.73.0`) and Azure SDK dependency versions required by the stable generated DPS SDK
* Adds focused INT coverage using a dedicated serial DPS resource to avoid cross-test contamination.
* Fixes CI issues exposed by the sync: Windows `cp1252`-safe ADR logging and CredScan suppression for the fake ADR test credential.
* `HISTORY.rst` - update `0.33.0b1`


cheatsheet1999:
@zzcodework SRE Agent, please do code review.

cheatsheet1999:
@HangyiWang
Hi Hangyi, when you have a chance, would you mind reviewing this PR? Since this is related to v0.33.0b1, I think you have the most context on this.

Most of the changes are carried over directly from the dev branch. I’d especially appreciate your review of the integration test-related changes in `run-tests-parallel.yml` and `conftest.py`.

I also included fixes for the PR checks pipeline: CredScan finding and the Windows console-encoding issue that surfaced in the earlier ADR changes (_log.py).

Thank you very much!

cheatsheet1999:
@zzcodework SRE Agent

zzcodework:

Reviewing `full diff` (head `27dce6dd`)

Merge of `dev` into `preview` plus the ported DPS `--disable-local-auth` feature. The command-surface change itself (`azext_iot/core/params.py`, `core/custom.py`, `core/help.py`) is clean and matches `dev` exactly; the HISTORY.rst restructuring is a faithful merge (the removed `0.32.0b2 (Preview)` heading was never released, and `0.31.0` sitting above `0.32.0b1` is chronologically correct per the release list). Most of my comments are about the new integration test and its CI wiring.

**Blocking**

- `azext_iot/tests/dps/conftest.py` — `_iot_dps_provisioner` unconditionally builds its return dict with `"connectionString": get_dps_cstring(dps_name, ENTITY_RG)`, and `get_dps_cstring` runs `iot dps connection-string show --policy-name provisioningserviceowner`, i.e. `list_keys_for_key_name`. The new `provisioned_iot_dps_local_auth_module` fixture creates the DPS with `--disable-local-auth true`, and blocking shared-access-key retrieval is precisely what `disableLocalAuth` does on the sibling IoT Hub RP. If DPS behaves the same, this `cli.invoke(...).as_json()` fails during fixture setup and takes down the entire `test_dps_disable_local_auth_int.py` module — including `test_dps_disable_local_auth_dataplane`, which then never gets the connection string it needs. Suggested fix: provision the instance with local auth **enabled** (drop `disable_local_auth=True` from the fixture and let `test_dps_create_disable_local_auth` cover the create path with its own short-lived resource), or capture the connection string before disabling. Worth confirming against a live run in the target subscription before merge — if the RP does allow `listKeys` with `disableLocalAuth=true`, ignore this.

**Suggestions**

- `.azure-devops/templates/run-tests-parallel.yml` and `tox.ini` — `set -e` is the right call (without it the step's exit code would be the *second* pytest's, masking failures in the main suite). But the consequence is that a single failure in the large parallel DPS suite now skips the serial RBAC file entirely and never emits `junit/test-iotext-dps-rbac-int.xml`, so the RBAC coverage silently disappears exactly when the branch is unhealthy. Consider capturing the exit codes instead (`pytest ... ; rc1=$?` / `pytest ... ; rc2=$?` / `exit $(( rc1 || rc2 ))`) so both suites always run and both junit files are always published. `tox.ini` has the same fail-fast behaviour for the two `DPS:` commands.
- `tox.ini` / `run-tests-parallel.yml` — the split depends on the literal path `azext_iot/tests/dps/core/test_dps_disable_local_auth_int.py` being repeated in two places, and `pytest --ignore=<nonexistent path>` is silently accepted. If the file is ever renamed or split, the `--ignore` becomes a no-op and this test rejoins the `-n 7` parallel run, where it will toggle service-wide auth under xdist — the exact race this PR is structured to avoid. A marker (e.g. `@pytest.mark.serial` + `-m "not serial"` / `-m serial`) would keep the two invocations in sync automatically.
- `azext_iot/tests/dps/core/test_dps_disable_local_auth_int.py` — `test_dps_disable_local_auth_dataplane` asserts that SAS/connection-string data-plane calls start failing immediately after `iot dps update --disable-local-auth true` returns. Control-plane ARM completion does not guarantee the data plane has picked the setting up, and the same applies to the re-enable at the end where it asserts SAS works again. Without a short poll/retry around those two transitions this is a likely flake source; the `--reruns 2` in CI would mask it as an intermittent failure rather than surface it. A bounded retry loop on the expected auth outcome would make the intent explicit.

**Nits**

- `azext_iot/core/custom.py`, `iot_dps_update` — `parameters["properties"]["disableLocalAuth"] = disable_local_auth` assumes the generic-update getter payload always carries a `properties` key; a `KeyError` here surfaces as a raw traceback rather than a CLI error. `parameters.setdefault("properties", {})["disableLocalAuth"] = ...` is a one-line hardening. (Ported verbatim from `dev`, so purely optional here.)
- `azext_iot/tests/dps/conftest.py` — `provisioned_iot_dps_local_auth_module` is declared `scope="module"` while its siblings `provisioned_iot_dps_module` / `provisioned_iot_dps_no_hub_module` are `scope="session"` despite the same `_module` suffix. No functional impact today (one consumer file), but the naming now means two different things in the same file.

**Also checked / looks good**

- `setup.py` raising `azure-core>=1.31.0` / `azure-mgmt-core>=1.5.0` while keeping preview's `<1.40.0` upper bound is correct and tight — `settings.azure_cloud` first appears in azure-core 1.31.0 and `get_arm_endpoints` in azure-mgmt-core 1.5.0, both of which the regenerated DPS client imports. This closes the dependency-floor gap that existed on `preview` (the earlier fix landed only on a `dev`-side branch).
- Dropping `{"enableDataResidency": None}` in favour of an empty properties dict is the right change for a raw-JSON body with no serializer, and the unit tests (`test_dps_create_omits_unset_optionals`, `test_dps_stable_sdk_unit`, `test_dps_adr_namespace_unit`) were updated consistently.
- `_print_pretty` in `azext_iot/tests/adr/_log.py` correctly handles the `None` encoding case before calling `encode`.
- The CredScan suppression points at a file that exists on this head (`azext_iot/tests/adr/test_adr_command_log_unit.py`).

I did not repeat the earlier inline points from Copilot about `provisioned_iot_dps_no_hub_module` being shared across xdist workers — the dedicated `dla` fixture plus the serial CI split in this head addresses them.

cheatsheet1999:
@zzcodework SRE Agent: Please re-check on these ones and confirm if your original analysis was valid.
For the blocking issue you mentioned: Connection-string blocking. Confirmed against live DPS in Integration Tests run  34523125079 . The fixture created the DPS with  disableLocalAuth=true , successfully retrieved the connection string through control-plane `listKeys` , and all three RBAC tests passed. `disableLocalAuth` only blocks SAS authentication on the data plane, so no fixture change is needed. Please re-check on this one and confirm if your original analysis was valid.
2. Always running both suites: I’m keeping the current fail-fast behavior to avoid consuming additional live-test resources after the main DPS suite has already failed.
3. Pytest marker: Good maintainability suggestion, but I’m keeping the explicit path to minimize this release change.
4. Propagation polling: Agreed. I’ll add bounded polling around disable/re-enable and verify the expected 401002  authorization failure rather than accepting any CLI failure.
5. `properties` hardening: The ARM DPS response requires `properties` , so the current direct access is intentional.
6. Fixture scope: Module scope is intentional because the dedicated resource is consumed only by this serial test module.

Let me know if there's anything that is still valid, or you are good with my response.

zzcodework:
### Azure SRE Agent - reply

> Please re-check on these ones and confirm if your original analysis was valid.

**1. Connection-string retrieval under `disableLocalAuth` — my finding was wrong, withdrawing it.**
You are right. I re-checked run [34523125079](https://github.com/Azure/azure-iot-cli-extension/actions/runs/34523125079) at head `27dce6dd`: the `DPS py3.13 (westus)` job is green, and since `get_dps_cstring(...)` runs unconditionally inside `_iot_dps_provisioner` (`azext_iot/tests/dps/conftest.py`, ~line 356) a failure would have taken out module setup for the whole local-auth module, not one test. A green DPS job at that head is direct evidence that control-plane `listKeys` still succeeds on a DPS created with `--disable-local-auth true`, i.e. `disableLocalAuth` gates SAS on the data plane only. I had reasoned from the IoT Hub RP behaviour and generalised it to DPS without evidence — that was the error. No fixture change needed.

**4. Propagation polling** — agreed, and asserting the specific `401002` rather than any CLI failure is the right tightening.

**2 (fail-fast), 3 (explicit path vs marker), 5 (`properties` access), 6 (fixture scope)** — your reasoning is clear and these were suggestion/nit level; deferring to you, nothing further from me.

Net: nothing from my review is still blocking.

_This is an automated reply and may be incomplete._

HangyiWang (approved):

CredScanSuppressions.json
```diff
@@ -6,6 +6,10 @@
             "placeholder": "password=sas",
             "_justification": "Client device simulation requires a SAS key extracted from IoT Hub to be passed in MQTT lib."
         },
+        {
+            "file": "azext_iot\\tests\\adr\\test_adr_command_log_unit.py",
```
HangyiWang, line 10: thanks

cheatsheet1999: Thanks for the review.

PR #897, opened 2026-09-16
feat: support MQTT 5 connection profiles for IoT Hub

cheatsheet1999:
PR description

 ## Summary
 - Add create-only `--connection-profile {Classic,MqttV5}` and `--yes` to `az iot hub create`.
 - Preserve existing Classic behavior and send the required MQTT 5 routing payload.
 - Add a temporary modeless LRO compatibility wrapper and focused tests.
 
image

image


azext_iot/tests/iothub/core/test_iot_hub_mqtt5_int.py
```diff
@@ -0,0 +1,41 @@
...
+
+@pytest.fixture()
+def provisioned_mqtt5_hub():
+    name = generate_hub_id()
+    hub_resource = cli.invoke(
+        f"iot hub create -n {name} -g {RG} --sku S1 "
+        "--connection-profile MqttV5 --yes",
+        capture_stderr=True,
+    )
+    try:
```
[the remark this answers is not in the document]
cheatsheet1999: This is valid, I'll make a change to move hub create inside try block

azext_iot/_factory.py
```diff
@@ -57,6 +60,50 @@ def _get_arm_endpoint(cli_ctx):
...
+    operation = getattr(operation_group, operation_name)
+
+    @wraps(operation)
+    def wrapped(*args, **kwargs):
+        if kwargs.get("polling", True) is True:
+            kwargs["polling"] = _ModelessJsonARMPolling(
+                timeout=kwargs.get("polling_interval", operation_group._config.polling_interval),
+                path_format_arguments={"endpoint": operation_group._config.base_url},
+                result_callback=kwargs.get("cls"),
+            )
```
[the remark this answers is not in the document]
cheatsheet1999: This is just a narrow workaround for the current TypeSpec issue, so I’d prefer to keep it limited to what the extension actually needs. The proper fix should come from TypeSpec, and we can remove this once the SDK is regenerated.

zzcodework:

Reviewing `full diff` (head `796037b`, base `mqtt5-custom-topic`)

**Blocking**

None found.

**Suggestions**

1. `azext_iot/core/custom.py` — `_is_mqtt_v5_profile` only matches `str`, so an `IotHubConnectionProfile` member silently degrades. `iot_hub_create(connection_profile=IotHubConnectionProfile.MQTT_V5)` (a plausible programmatic/test call, and the type hint is `Optional[str]` so nothing stops it) takes the `isinstance(connection_profile, str)` branch to `False`: the confirmation prompt is skipped, `properties["routing"]` is never added, **and** `properties["connectionProfile"]` is set to the enum object itself, which is not what the service contract expects. Normalising once at the top — e.g. `connection_profile = IotHubConnectionProfile(connection_profile).value if connection_profile else None` — removes the whole class of mismatch and keeps the `casefold()` comparison honest.

2. `azext_iot/_factory.py` — `_wrap_modeless_lro_operation` drops the per-call polling configuration the generated code forwards. The generated operation builds `ARMPolling(lro_delay, path_format_arguments=path_format_arguments, **kwargs)`, so per-call pipeline/LRO overrides (`lro_options`, retry settings, etc.) reach the polling method; the wrapper constructs `_ModelessJsonARMPolling(timeout=..., path_format_arguments=..., result_callback=...)` with nothing else forwarded. No current call site passes those, so this is not a live bug, but it is a behaviour difference worth a comment next to the `TODO` so the follow-up removal is a clean revert rather than a silent change.

3. `azext_iot/tests/test_factory_unit.py` — `TestIotHubModelessLroPolling` patches `azext_iot.sdk.iothub.mgmt.IotHubClient` wholesale, so `operation_group._config.polling_interval` and `operation_group._config.base_url` are auto-created `MagicMock` attributes: the assertions would pass even if the real configuration object did not expose them. They do exist today (`azext_iot/sdk/iothub/mgmt/_configuration.py` sets `self.base_url` and `self.polling_interval`), so the tests are not wrong — but a one-line assertion against the real `IotHubClientConfiguration` attribute names, or a `spec=`/`autospec` on the patched client, would make the suite catch a codegen rename instead of sailing past it.

**Nits**

4. No `HISTORY.rst` entry for the new `--connection-profile` / `--yes` surface. Given the base here is the `mqtt5-custom-topic` feature branch rather than `preview`, this is presumably intended to be rolled up when the feature branch merges — worth confirming it does not get lost.

5. `azext_iot/tests/iothub/core/test_iot_hub_mqtt5_int.py` — the `finally` block constructs a second `EmbeddedCLI()` instead of reusing the module-level `cli`; harmless, but inconsistent with the rest of the file.

**What I checked and found correct**

- The set of wrapped operations is exactly the set of broken generated callbacks. In `azext_iot/sdk/iothub/mgmt/operations/_operations.py`, `get_long_running_output` omits `response = pipeline_response.http_response` in precisely four places — `PrivateEndpointConnectionsOperations.begin_update`, `PrivateEndpointConnectionsOperations.begin_delete`, `IotHubResourceOperations.begin_create_or_update`, `IotHubResourceOperations.begin_delete` — which are the four wrapped in `_configure_iot_hub_modeless_lro_polling`. `IotHubResourceOperations.begin_update` and `IotHubOperations.begin_manual_failover` are correctly left alone: the former defines `response` properly, the latter never touches it. The DPS management SDK (`azext_iot/sdk/dps/mgmt/operations/_operations.py`) also defines `response` correctly in every LRO callback, so the IoT-Hub-only scoping is right.
- `IotHubClient` is only instantiated in `azext_iot/_factory.py`, so no code path can obtain an unwrapped client.
- `cls` is applied exactly once: the generated operation pops `cls` and only uses it inside the discarded `get_long_running_output`, while the wrapper re-applies it through `result_callback`.
- `polling_interval` is read with `kwargs.get`, not popped, so the generated operation still pops it as before — no duplicate-kwarg hazard.
- `prompt_y_n` without a `NoTTYException` guard matches existing usage in `azext_iot/iothub/providers/device_identity.py` and `azext_iot/iothub/providers/state.py`, so the new prompt is consistent with the repo.
- On the two points already raised inline: the `polling=False` / `NoPolling` gap is real in principle, but no IoT Hub call site in `azext_iot/core/custom.py` passes `polling=False` — `--no-wait` for `iot hub create`/`delete`/`manual-failover`/`policy renew-key` is handled by returning the poller, not by disabling polling — so the narrow scoping you described holds today. The hub-create-inside-`try` fix for the integration fixture is the right call.

Summary: the workaround is correctly targeted and the new CLI surface is well covered by unit tests. Nothing blocking; item 1 is the one I would still change before merge.

*This is an automated review and may be incomplete.*

cheatsheet1999:
> Reviewing `full diff` (head `796037b`, base `mqtt5-custom-topic`)
1. CLI parsing already converts this enum to a canonical string. Passing an `Enum` object is outside the declared `Optional[str]` contract, so I prefer not to add support for an unused path.

2. This workaround intentionally covers only the default polling path used by current call sites. Forwarding unused LRO options would add unnecessary complexity to temporary code.

3. This test verifies our wrapper wiring, not future generated SDK schemas. The attributes exist today, and the wrapper will be removed when the SDK is regenerated.

4. This PR targets the intermediate feature branch. I will add one consolidated `HISTORY.rst` entry when the complete feature merges into `preview`.

5. The separate `EmbeddedCLI` keeps cleanup state independent and non-raising, so a cleanup failure cannot replace the original test failure.

Let me know if there're any concerns.

zzcodework:
### Azure SRE Agent - reply

> 1. CLI parsing already converts this enum to a canonical string. Passing an `Enum` object is outside the declared `Optional[str]` contract, so I prefer not to add support for an unused path.

You're right, and I've re-verified it: `azext_iot/core/_params.py` registers `connection_profile` with `arg_type=get_enum_type(IotHubConnectionProfile)` under `iot hub create`, so argparse resolves the choice to the enum's *value* string before `iot_hub_create` is ever called. Every CLI path therefore hits the `isinstance(connection_profile, str)` branch, and the only way to reach the degraded path is a direct programmatic call that already violates the declared `Optional[str]` hint. That makes it an out-of-contract input rather than a defect — I'm withdrawing the suggestion.

(For the record, the one detail that made it worth raising: `IotHubConnectionProfile` is a plain `Enum`, not a `str`-mixin, so an enum member would *not* have satisfied the `isinstance` check by accident. With the CLI normalizing upstream that's moot.)

On 2-5: all reasonable as scoped. 2 and 3 were explicitly flagged as non-live given the temporary nature of the wrapper, and "this goes away when the SDK is regenerated" is the right trade-off for throwaway code. 4 — consolidating one `HISTORY.rst` entry at the `preview` merge is exactly what I was asking you to confirm, so that's settled. 5 was a nit and your rationale (keeping cleanup non-raising so it can't mask the original failure) is a better reason than consistency.

No remaining concerns from my side; nothing blocking on head `796037b`.

*This is an automated reply and may be incomplete.*

azext_iot/core/custom.py
```diff
@@ -1023,6 +1045,10 @@ def iot_hub_create(
                     "disableDeviceSAS": disable_device_sas,
                     "disableModuleSAS": disable_module_sas}
     properties["enableFileUploadNotifications"] = enable_fileupload_notifications
+    if connection_profile is not None:
+        properties["connectionProfile"] = connection_profile
+    if _is_mqtt_v5_profile(connection_profile):
+        properties["routing"] = {"endpoints": {}}
```
HangyiWang, line 1051: why empty endpoint here for routing?

azext_iot/core/custom.py
```diff
@@ -56,6 +59,9 @@
 # Identity types
 SYSTEM_ASSIGNED = 'SystemAssigned'
 NONE_IDENTITY = 'None'
+MQTT_V5_CREATE_CONFIRMATION = (
+    "The MqttV5 connection profile is permanent and cannot be changed after the IoT Hub is created. Continue?"
```
HangyiWang, line 63: do we have y/n after the `Continue?`

azext_iot/_factory.py
```diff
@@ -57,6 +60,51 @@ def _get_arm_endpoint(cli_ctx):
...
+        return operation(*args, **kwargs)
+
+    setattr(operation_group, operation_name, wrapped)
+
+
+def _configure_iot_hub_modeless_lro_polling(client):
```
HangyiWang, line 100: do we need to merge this into dev directly

azext_iot/core/custom.py
cheatsheet1999: The service requires a non-empty routing object when MQTT 5 is enabled. I sent multiple probe request, `routing={}` is rejected with IH400111, while `{"endpoints": {}}` is the minimum accepted shape from live validation. No actual routing endpoint is configured here.

azext_iot/core/custom.py
cheatsheet1999: Yes. `prompt_y_n(..., default="n")` automatically appends `(y/N):` to the message.
You can also see a screenshot in the description section, if that answers your question.

azext_iot/_factory.py
cheatsheet1999: I’d prefer not to merge this into `dev`, and ideally avoid carrying it into `preview` as well, since it’s only a temporary workaround for `microsoft/typespec#11966`.

If there’s no progress by end of next week, I’m happy to send them a fix PR. Once they merge and release, we can regenerate the SDK and remove this wrapper.

azext_iot/core/_params.py
```diff
@@ -427,6 +433,19 @@ def load_arguments(self, _):  # pylint: disable=too-many-statements
         c.argument('hub_name', completer=None)
         c.argument('location', get_location_type(self.cli_ctx),
                    help='Location of your IoT Hub. Default is the location of target resource group.')
+        c.argument(
+            'connection_profile',
```
zzcodework, line 437: Looks like we also need to add the --connection-profile to DPS as well, we can include DPS changes to this PR or a new one.

cheatsheet1999: Let’s use a separate PR for the DPS command to keep the scope separate. I also need to understand the API contract a bit better, including the shape of the request and parameters.

HangyiWang (approved):

PR #899, opened 2026-09-21
fix(adr): grant namespace MI self access for DPS links

zzcodework:
## Summary
- grant the ADR namespace system-assigned identity `Contributor` on its own namespace during DPS link preflight
- preserve the configured outbound identity for target-service permissions
- add unit coverage for the separate system/outbound identity case

## Validation
- `python -m py_compile azext_iot\adr\rbac.py azext_iot\adr\providers\link_preflight.py azext_iot\tests\adr\test_adr_link_rbac_unit.py azext_iot\tests\adr\test_adr_link_runtime_preflight_unit.py`
- `git diff --check`

The focused tox suite could not start because pip failed copying generated SDK files from the system temp drive into the worktree. A direct run was then blocked by a local pre-existing `cryptography` DLL mismatch.


zzcodework:

Reviewing `full diff` (head `9c26985`)

Base branch is `users/hangyiwang/dps-hub-adr-ignite-change`, so base context was read from that ref rather than the local `dev` clone.

The change itself is coherent: the new `RoleRule("namespace_system", CONTRIBUTOR_ROLE, "namespace")` is added to the `dps` matrix only, `resolve_namespace_system_principal` mirrors the existing `resolve_namespace_outbound_principal` shape, the new `namespace_system_principal_id` kwarg is appended last on `RoleManager.ensure` (so it stays backwards compatible), and both `verify_many` and `ensure_many` were updated to carry the new principal — missing either would have been a `KeyError`. A few things I checked that are *not* problems, to save reviewer time:

- `azext_iot/adr/_help.py` renders the DPS role list via `format_role_requirements("dps")`, so the help text picks up the third grant automatically — no help edit needed.
- `test_adr_link_initiate_recovery_unit.py` asserts `len(h.created) == len(LINK_ROLE_MATRIX[kind])`, which self-adjusts as the matrix grows. With the harness namespace (SAMI `namespace-principal`, outbound SAMI or UAMI) all three DPS assignments are distinct tuples, so the count still matches for both the `system` and `user` parametrizations.
- Every ADR unit-test namespace fixture declares a `SystemAssigned` identity, so the new mandatory resolve does not break existing DPS tests.

**Blocking**

None.

**Suggestions**

1. **Missing changelog entry for the version bump.** `azext_iot/constants.py` moves `VERSION` from `0.33.0b12` to `0.33.0b13`, but `HISTORY.rst` still tops out at the `0.33.0b12 (Preview)` section. The repo convention is an entry per preview bump, and this one is user-visible twice over: DPS links now request a third role assignment, and a namespace without a system-assigned identity is now rejected outright. Both deserve a line so operators who see a new `Contributor` grant on the namespace can trace where it came from.

2. **The new DPS rule is mandatory, but the plumbing treats a missing principal as "optional, skip".** `RoleManager.ensure` declares `namespace_system_principal_id: Optional[str] = None`, and both `ensure_many` and `verify_many` short-circuit with `if not principal_id: continue` — an escape hatch that exists deliberately for the optional Hub inbound identity ("Hub inbound identity is optional. There is no principal to grant in that direction when the endpoint omits it."). The new `namespace_system` rule is not optional, so a `dps` request that omits the key would silently drop a required grant and then fail later at link time with a service-side authorization error instead of failing loudly in preflight.

   Today this is latent rather than live: the only producer of these request dicts is `preflight_target` in `azext_iot/adr/providers/link_preflight.py`, which always populates the key for `link_type == "dps"`, and the recovery path in `link.py` re-enters the same function, so both the `ensure_many` plan and the `verify_many` recovery check stay consistent. But the `None` default plus the blanket skip makes it easy to reintroduce silently. Consider either marking the rule as required on `RoleRule` and raising when a required principal is absent, or validating in `ensure_many` that a `dps` request carries `namespace_system_principal_id`.

**Nits**

3. `resolve_namespace_system_principal` in `azext_iot/adr/rbac.py` — both error strings say "The namespace ..." without naming which namespace, while the sibling `resolve_linked_resource_principal` messages interpolate `display_name`. Since this can now fire from the recovery re-preflight path as well as the initial link, including the namespace name (and ideally the `az iot adr ns update --system-assigned-mi` remediation) would make the failure self-explanatory.

**Summary**

No blocking issues. The role-matrix change and its propagation through `ensure`/`ensure_many`/`verify_many` look correct and are covered by the two new unit tests, including the split system/outbound identity case the PR description calls out. Main asks are the `HISTORY.rst` entry for the `0.33.0b13` bump and tightening the optional-principal handling so a mandatory rule cannot be skipped by omission.

*This is an automated review and may be incomplete; please verify findings before acting on them.*

zzcodework:
## Review: does this correctly resolve work item 39640174?

**The mechanism is right. It is not mergeable as-is — it breaks CI, and the role choice needs an explicit decision plus a comment.**

---

### 1. Root cause targeting — correct

The bug's original description (no built-in role grants Device Registry `dataActions`) was retracted by the reporter on 2026-09-15. The verified cause is that the ADR namespace's **own system-assigned MI** had zero role assignments on the namespace resource itself; every assignment attempted during the bug bash pointed outward (DPS MI -> namespace, namespace MI -> DPS/hub). The verified workaround was `Contributor` for the namespace SAMI at namespace scope.

This PR implements exactly that:

- `RoleRule("namespace_system", CONTRIBUTOR_ROLE, "namespace")` added to the `dps` matrix — correct principal, correct scope, correct link type.
- `resolve_namespace_system_principal` reads `identity.principalId` (the SAMI) rather than reusing `resolve_namespace_outbound_principal`. This is the right call and the detail most likely to have been gotten wrong — for a split-identity namespace the outbound MI is not the principal the service mints its ARM token for.
- Plumbed consistently through `ensure` / `ensure_many` / `verify_many`, so the recovery path re-verifies rather than re-grants.

### 2. Blocking — CI is red

The base branch `users/hangyiwang/dps-hub-adr-ignite-change` is fully green. This head fails `build` and every unit-test job: **11 failed, 10424 passed**.

Only the two test files touched in this PR were updated. Three other suites derive their expectations from `LINK_ROLE_MATRIX` and assume exactly two principal kinds, mapping `rule.principal == "namespace"` -> outbound and **everything else** -> inbound. The new `namespace_system` rule falls into the else branch:

```
At index 1 diff: ('namespace-principal', 'Contributor', '.../namespaces/ns')
              != ('target-principal',    'Contributor', '.../namespaces/ns')
azext_iot/tests/adr/test_adr_link_propagation_unit.py:183: AssertionError
```

Failing suites:
- `test_adr_link_propagation_unit.py::test_fresh_roles_recover_exact_persisted_endpoint_update` (4 params)
- `test_adr_link_rbac_unit.py::test_assignment_plan_is_visible_before_every_write[dps]`
- `test_adr_link_recovery_unit.py::test_persisted_failed_link_rejects_add_but_update_reruns_real_preflight` (4 params)
- `test_adr_recovery_runtime_edges_unit.py::test_malformed_unrelated_endpoint_collections_prevent_retry_patch` (2 params)

These are harness gaps rather than product defects — the actual created assignment is correct in every case — but the shared helpers need to learn the third principal kind. The PR description notes the suite could not be run locally; this is what it would have caught.

### 3. `Contributor` is broader than requested — but the narrow alternative is currently unbuildable

The service-side recommendation on the work item was to grant "the narrowest role containing the registry-device management-plane write permission (principally `Microsoft.DeviceRegistry/namespaces/registryDevices/write`)", noting `Contributor` "proves the diagnosis but is broader than the product contract should require."

**That recommendation is not actionable today.** Checking the live ARM provider operations for `Microsoft.DeviceRegistry`:

- There is **no `namespaces/registryDevices` resource type**. The registered device type is `namespaces/devices` (`read`/`write`/`delete`/`revoke/action`).
- `namespaces/certificateAuthorities` is **also absent**. What is registered is the older shape: `namespaces/credentials` and `namespaces/credentials/policies` (with `issueCertificate/action`, `revokeIssuer/action`).

Meanwhile `specification/deviceregistry/resource-manager/Microsoft.DeviceRegistry/DeviceRegistry` defines `/namespaces/{namespaceName}/registryDevices/{registryDeviceName}` and `/namespaces/{namespaceName}/certificateAuthorities/{ca}/certificatePolicies/{policy}` in `2026-11-02-preview` and `stable/2026-11-01`.

The RBAC manifest is behind the resource model, so a custom role naming `registryDevices/write` fails validation. This is also why the reporter found the permission undiscoverable via `providerOperations` — that lead was real, just misattributed.

Nor is a single write action obviously sufficient once registered. The flow spans an LRO `PUT registryDevices/{name}`, certificate issuance against `certificateAuthorities/.../certificatePolicies/...` (the verified CSR run returned a 3-entry `issuedCertificateChain`), and `authenticationProfiles/{profile}/listKeys` for the symmetric-key path.

One useful bound does follow from the verified workaround: `Contributor` grants no `dataActions`, and a namespace-scoped assignment does not reach the subscription-scoped `operationStatuses` / `locations/operationStatuses` types. So everything this flow needs is a control-plane action at or below namespace scope, and none of it is a dataAction. A narrow namespace-scoped role is therefore feasible — once the manifest lands and the ADR team enumerates the action set.

**Suggested resolution for this PR:** keep `Contributor`, but add an inline comment at the new `RoleRule` recording that it is provisional pending provider-manifest registration of `namespaces/registryDevices`, so the next reader does not assume it was chosen casually — particularly given this repo already tracks over-granting `Contributor` on the inbound direction.

### 4. Unflagged behavioral change

`resolve_namespace_system_principal` raises `InvalidArgumentValueError` when the namespace has no SystemAssigned identity. A namespace configured with **only** a user-assigned identity can run `az iot adr ns link dps add` today; after this change it hard-fails at preflight.

That may well be correct — registration appears to genuinely require the SAMI — but it is a breaking change with no test and no `--help` note. Please either add a case asserting the failure is intentional, or relax it.

### 5. Coverage note (not blocking)

This fixes the `az iot adr ns link dps add/update` path. Hooking at link time is defensible since DPS->ADR registration requires a DPS link — but users who create the namespace and link via portal/ARM/Bicep stay broken. Worth confirming the parallel Portal-side change is tracked, since the work item discussion had that in flight.

### 6. Nits

- `constants.py` bumps `0.33.0b12` -> `0.33.0b13` with no corresponding `HISTORY.rst` entry.
- `resolve_namespace_system_principal`'s two error strings say "The namespace ..." without naming which namespace, unlike sibling `resolve_linked_resource_principal` which interpolates `display_name`. Including the namespace name and the `az iot adr ns update` remediation would make this self-explanatory — this message can now fire from the recovery re-preflight path too.

---

### Summary

Correct diagnosis, correct principal, correct scope. Before merge:

1. Fix the 11 failing tests (teach the shared harnesses the third principal kind).
2. Add an inline comment justifying `Contributor` as provisional, referencing the missing provider-manifest registration.
3. Add the `HISTORY.rst` entry for `0.33.0b13`.
4. Decide and cover the SAMI-required behavior for UAMI-only namespaces.

Note that the work item's secondary ask — the `403000` message naming neither principal, scope, nor missing action — is service-side and correctly out of scope here.

*Posted by Azure SRE Agent. Findings verified against live ARM provider operations, the spec repo, and CI logs for this head vs. the base branch; please confirm before acting.*

zzcodework:

Reviewing `changes since 9c26985` (head `cef1e8f`)

The only new commit is `test(adr): expect DPS namespace self assignment`, which adapts the four existing ADR unit suites to the `RoleRule("namespace_system", CONTRIBUTOR_ROLE, "namespace")` entry added earlier in this PR. No production code changed in this increment.

**Blocking**

None.

**Suggestions**

None blocking, but one gap from the previous iteration is still open and this test commit was the natural place to close it: `ensure_many` in `azext_iot/adr/rbac.py` skips any rule whose principal is falsy (`if not principal_id: continue`), which is documented for the *optional* Hub inbound identity but also silently drops the now-**mandatory** DPS `namespace_system` grant if a producer omits it. The new cases all supply the principal — `test_adr_link_rbac_unit.py` passes `namespace_system_principal_id="ns-system-principal"` for `dps` and `None` otherwise, and `test_adr_link_propagation_unit.py` hardcodes `"namespace_system": "namespace-principal"` — so nothing locks the negative case. A small test that calls `ensure("dps", ..., namespace_system_principal_id=None)` and asserts it does **not** quietly produce a 2-assignment plan would turn today's latent issue into a caught one. Still a suggestion, not a blocker: the only producer, `preflight_target` in `azext_iot/adr/providers/link_preflight.py`, always populates it for `dps` via `resolve_namespace_system_principal`.

**Nits**

None.

**Verification performed**

- `test_adr_link_propagation_unit.py`: the new `principals` dict maps `namespace_system` to the literal `"namespace-principal"` rather than to `outbound_principal`. That is correct, not a copy-paste slip — the harness namespace always carries `identity.principalId = "namespace-principal"`, and the `identity == "user"` case only *adds* `userAssignedIdentities`, so the SAMI principal stays `namespace-principal` while the outbound becomes `namespace-user`. This matches `resolve_namespace_system_principal`, which reads `identity.principalId` regardless of the outbound selection.
- `test_adr_recovery_runtime_edges_unit.py`: the inserted `("namespace-principal", "Contributor", NS_ID)` sits between the target-scope grant and the linked grant, which matches the declaration order of `LINK_ROLE_MATRIX["dps"]`, and the bumped `len(harness.assignments) == 3` is right. The two `namespace-principal` tuples are not deduped by `ensure_many` because their scopes differ (`KINDS["dps"][2]` vs `NS_ID`), so the count holds.
- `test_adr_link_recovery_unit.py`: the kwargs in `expected_rbac` match `LinkRbacManager.ensure`'s signature one-for-one, and `preflight_target` calls `manager.ensure(**request)`, so `assert_called_once_with(**expected_rbac)` will match; expecting `namespace_system_principal_id="namespace-principal"` only for `kind == "dps"` mirrors the `if link_type == "dps"` guard in `preflight_target`.
- `test_adr_link_rbac_unit.py`: `assert invoke.call_count == len(LINK_ROLE_MATRIX[link_type])` self-adjusts to 3 for `dps`, and the added log assertions match `_format_role_requirement`'s `"namespace system-assigned MI"` label exactly.
- All four files end in `_unit.py`, so they are picked up by the repo's unit selector.

Summary: a clean, correctly scoped test-only increment that locks the new DPS namespace self-assignment across the propagation, RBAC, recovery and runtime-edge suites. Nothing blocking.

_This is an automated review and may be incomplete; please treat it as one input among others._

